PR #61, opened 2016-10-05
change hostname in container to mender-client instead of IP

GregorioDiStefano:


GregorioDiStefano:
@pasinskim please merge if okay

PR #71, opened 2016-12-06
Remove misleading suggestion that integration setup is for production

maciejmrowiec:


maciejmrowiec:
@eysteins in relation to mailing list question

PR #74, opened 2016-12-08
split compose into backend and client parts

bboozzoo:
Move client service to `docker-compose.client.yml`, update README.

@mendersoftware/rndity @GregorioDiStefano @maciejmrowiec

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,21 @@ Again, recompiling your local binary and restarting `integration` will make
 your changes take effect. Note that the correct API Gateway config is probably already
 set up for an existing service; if not, refer the previous section on how to modify it.
 
+## Demo client
+
+The setup comes with a predefined client service (mender-client) that runs a
+qemu VM in a container. The client will connect to the backend by accessing
+`docker.mender.io` host (an alias assigned to `mender-api-gateway` service). The
+client container will not be started by default and needs to be included
+explicitly when running docker compose by listing multiple compose files as
+described in [compose manual](https://docs.docker.com/compose/extends/#/multiple-compose-files).
+
+To start the backend and a demo client run the following command:
+
+```
+docker-compose -f docker-compose.yml -f docker-compose.client.yml up
+```
+
 ## Contributing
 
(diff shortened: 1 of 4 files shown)


bboozzoo:
anyone else wants to take a look?

GregorioDiStefano:
@bboozzoo lets wait until @maciejmrowiec takes a look, so this is a pretty fundamental change to our environment.

pasinskim:
I like the change as well, and it might be useful to continue with having separate yaml files for S3 and minio (as @bboozzoo suggested). 

This should come with some corresponding PR in documentation: https://github.com/mendersoftware/mender-docs/blob/master/01.Getting-started/02.Create-a-test-environment/docs.md

maciejmrowiec:
Looks good, definitely step in the right direction!

What do we do with the `up` script? it used to spin up env with client and now it wont if I understand this. How about having option --with-client or something that will load both compose files.

bboozzoo:
Maybe `up` should include a client container by default? @maciejmrowiec @pasinskim 

I don't use this script often (or at all) so I'm ok with client being included in `up` but skipping it in `docker-compose.yml`

maciejmrowiec:
had a brief chat with @eysteins, would be nice for `./up` script start demo environment by default to make it as simple as possible to get going, and later we can have --production flag to get it away

bboozzoo:
Ok, so we should be fine. With latest commits `up` will also start the client.

I guess nobody will be using `up` for production anyway as it's already mangling /etc/hosts.

bboozzoo:
ping?

bboozzoo:
anyone else wants to comment or can we just merge it now?

bboozzoo:
@pasinskim @GregorioDiStefano do you need to update integration tests once this is merged?

bboozzoo:
ping

GregorioDiStefano:
I think you can just merge this, and ill fix whatever breaks with integration

bboozzoo:
Very well, merging

PR #86, opened 2016-12-20
storage-proxy: storage proxy with rate and connection limiting

bboozzoo:
Add a storage proxy that applies download rate and connection limiting to
incoming GET requests.

The service is implemented using openresty and Lua extensions. When connection
limit is hit, a 503 Service Temporarily Unavailable is returned.

Minio is hidden behing storage proxy for all requests, i.e. internal requests
from Deployments Service will also go through storage proxy. This is required
because of signed request in S3, which the target host name is a part of.

The default limits are:
- download rate: 1MB/s
- max connections: 100

### DIFF
--- a/storage-proxy/nginx.conf
+++ b/storage-proxy/nginx.conf
@@ -1,0 +1,96 @@
+worker_processes  auto;
+
+error_log  /dev/stdout warn;
+pid        /var/run/nginx.pid;
+
+events {
+    worker_connections  1024;
+}
+
+env DOWNLOAD_SPEED;
+env MAX_CONNECTIONS;
+
+http {
+
+    init_by_lua_block {
+        ngx.log(ngx.WARN, "download speed limit: " .. (os.getenv("DOWNLOAD_SPEED") or "not set"))
+        ngx.log(ngx.WARN, "max connections: " .. (os.getenv("MAX_CONNECTIONS") or "not set"))
+    }
+
+    upstream minio_backend {
+        server minio:9000 max_fails=0;
+    }
+
+    lua_shared_dict my_limit_conn_store 100m;
+
+    server {
+
+        proxy_max_temp_file_size 0;
+
+        listen 9000;
+
+        location / {
+            access_by_lua_block {
+                -- rate and connection limiting is applied only to GET requests
+                if ngx.req.get_method() == "GET" then
+
+                   local max_connections = tonumber(os.getenv("MAX_CONNECTIONS"))
+                   if max_connections ~= nil then
+                      local limit_conn = require "resty.limit.conn"
+
+                      local lim, err = limit_conn.new("my_limit_conn_store", max_connections, 0, 1)
+                      if not lim then
+                         ngx.log(ngx.ERR,
+                                 "failed to instantiate a resty.limit.conn object: ", err)
+                         return ngx.exit(500)
+                      end
+
+                      local key = ngx.var.binary_remote_addr
+                      local delay, err = lim:incoming(key, true)
+                      if not delay and err ~= "rejected" then
+                         ngx.log(ngx.ERR, "failed to limit req: ", err)
+                         return ngx.exit(500)
+                      elseif not delay or delay > 0 then
+                         ngx.log(ngx.WARN, "connection rejected")
+                         return ngx.exit(503)
+                      end
+
+                      if lim:is_committed() then
+                         local ctx = ngx.ctx
+                         ctx.limit_conn = lim
+                         ctx.limit_conn_key = key
+                         ctx.limit_conn_delay = delay
+                      end
+                   end
+
+                   local download_speed = os.getenv("DOWNLOAD_SPEED")
+                   if download_speed ~= nil then
+                      ngx.var.limit_rate = download_speed
+                   end
+                end
+            }
+
+            log_by_lua_block {
+                local ctx = ngx.ctx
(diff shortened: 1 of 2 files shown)


storage-proxy/nginx.conf
```diff
@@ -0,0 +1,100 @@
...
+
+    lua_shared_dict my_limit_conn_store 100m;
+
+    server {
+
+        proxy_max_temp_file_size 0;
+
+        listen 9000;
+
+        limit_conn perserver 1;
```
kjaskiewiczz: ?

bboozzoo: fixed

maciejmrowiec:
as minio+nginx is exchangeable with S3 do you guys think it would be good idea to give it separate compose file so if you want to run with S3 you can easily take it away? - just a thought

storage-proxy/nginx.conf
```diff
@@ -0,0 +1,96 @@
...
+
+    upstream minio_backend {
+        server minio:9000 max_fails=0;
+    }
+
+    lua_shared_dict my_limit_conn_store 100m;
+
+    server {
+
+        proxy_max_temp_file_size 0;
```
maciejmrowiec: +1

storage-proxy/nginx.conf
```diff
@@ -0,0 +1,96 @@
...
+
+        listen 9000;
+
+        location / {
+            access_by_lua_block {
```
bboozzoo: `limit_rate` does not work inside `if` block (or at least I could not get past nginx complaining that limit_rate cannot be used in `if` configuration block). It wouldn't be possible to use `if ($request_method == GET) { ... }` and we obviously do not want to limit deployments -> minio upload. Given this, we'd have to use a separate prefix, what breaks request signing for all incoming requests (URL is part of data being signed).

Another alternative is allowing deployments to access minio directly and applying `limit_rate` on `location /` but again signing will be broken because host is part of signature as well.

Lua also makes environment access easy. An alternative is to generate a part of (or a whole) `nginx.conf` with `envsubst`, meaning we'd have to use a custom entrypoint to preform substitution, hence possibly an Dockerfile and another repository. But this does not resolve either of the first 2 problems.

bboozzoo:
I already have a variant where storage is split into separate files right here: https://github.com/bboozzoo/mender-integration/tree/bboozzoo/storage-split 

If there's any interest in this change I can open a PR

storage-proxy/nginx.conf
bboozzoo: Yeah, so I missed that and assume that rate limiting is in both directions.

maciejmrowiec:
not sure how others but I like it ;D settings overwrite looks pretty neat

maciejmrowiec:
@GregorioDiStefano ^^

bboozzoo:
@GregorioDiStefano @pasinskim since client download speed is capped @1MB/s your integration tests may take longer, I suggest you override the setting in an external compose file, something like:

```
storage-proxy:
    environment:
        - DOWNLOAD_SPEED: 100m
```

bboozzoo:
rebased

storage-proxy/nginx.conf
maciejmrowiec: what should we do with this one? should we simplify? or there are other reasons why we can't just limit_rate

bboozzoo: The whole block does rate and connection limiting. Rate limiting is this piece only:
```
+                   local download_speed = os.getenv("DOWNLOAD_SPEED")
+                   if download_speed ~= nil then
+                      ngx.var.limit_rate = download_speed
+                   end
```

maciejmrowiec: could something simple like this work?

```server {
    limit_rate = 1m;
    limit_conn perserver 100;
}```

Comment (bboozzoo) [inline storage-proxy/nginx.conf]: Yes. You'd need a custom entrypoint, a config template, accept that GET requests affect other methods. 

This is actually the version I started with, but since there was no reasonable way of just taking values from environment and shoving them into nginx variables I moved to a lua variant.

File: storage-proxy/nginx.conf [hunk 4]
```diff
@@ -0,0 +1,96 @@
...
+                      local limit_conn = require "resty.limit.conn"
+
+                      local lim, err = limit_conn.new("my_limit_conn_store", max_connections, 0, 1)
+                      if not lim then
+                         ngx.log(ngx.ERR,
+                                 "failed to instantiate a resty.limit.conn object: ", err)
+                         return ngx.exit(500)
+                      end
+
+                      local key = ngx.var.binary_remote_addr
```
maciejmrowiec: will this still fly when clients are NAT'ed? they would have same address from server perspective

storage-proxy/nginx.conf
maciejmrowiec: got it

storage-proxy/nginx.conf
bboozzoo: Yes, I would expect the address to be the same in such scenario. The again, opening multiple downloads will have a similar effect, and so far I've seen connection limiting to work properly (ex. setting MAX connections to 2 and the third download attempt gets 503).

maciejmrowiec: thanks for clarifying

GregorioDiStefano:
thanks for pointing out the config change @bboozzoo

PR #88, opened 2016-12-21
minio image updated

kjaskiewiczz:
minio image tag updated to existing one
minio section of docker-compose file updated


pasinskim:
Verified that this works.

PR #94, opened 2016-12-29
use authenticated session when making requests

GregorioDiStefano:
@kacf


tests/MenderAPI/__init__.py
```diff
@@ -15,7 +15,16 @@
 import admission
 import deployments
 import artifacts
+import requests
 
-adm = admission.Admission()
-deploy = deployments.Deployments()
+r = requests.post("https://127.0.0.1:8080/api/management/0.1/useradm/auth/login", verify=False)
```
kacf: Should be:

```
r = requests.post("https://%s/api/management/0.1/useradm/auth/login" % gateway, verify=False)
```

tests/MenderAPI/__init__.py
```diff
@@ -15,7 +15,16 @@
 import admission
 import deployments
 import artifacts
+import requests
```
kacf: Did you forget to submit this file? There is no such file in my directory.

kacf: Oh, wait, is this a 3rd party module? Maybe set it apart from the rest then.

tests/MenderAPI/__init__.py
```diff
@@ -16,6 +16,16 @@
 import deployments
 import artifacts
 
-adm = admission.Admission()
-deploy = deployments.Deployments()
+import requests
+
+r = requests.post("https://%s:8080/api/management/%s/useradm/auth/login" % (gateway, api_version), verify=False)
```
kacf: Lose the `:8080` as well. It's already part of `gateway`.

PR #96, opened 2016-12-30
Add docker-compose file for client development.

kacf:
With it you can use your build dir as input instead of the image.

### DIFF
--- a/docker-compose.client-dev.yml
+++ b/docker-compose.client-dev.yml
@@ -1,0 +1,9 @@
+version: '2'
+services:
+
+    #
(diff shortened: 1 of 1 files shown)


kacf:
@GregorioDiStefano

GregorioDiStefano:
the point of this is to reference this file when bringing up docker-compose in order to override the existing mender-client container, right?

kacf:
Exactly. How it will work if we launch the containers from within the tests is something I have to think about, but I'll deal with that at a later time.

GregorioDiStefano:
looks good!

PR #98, opened 2017-01-02
HTTPS for minio

bboozzoo:
Depends on https://github.com/mendersoftware/deployments/pull/190

Provide HTTPS certificates for storage-proxy service. Default (demo) certificates are mounted in `docker-compose.demo.yml`.

Default `AWS_URI` is switched to use `https://..`.

The certificate of storage-proxy is mounted into deployments service container and environment is setup so that the entrypoint will add certificate to system trust.


bboozzoo:
Hm, will update PR, seems like I've included some extra scripts.

mchalski:
+1

maciejmrowiec:
looks great!

bboozzoo:
Pushed some additional info about token signing keys and a note about docker-compose.demo.yml.

README.md
```diff
@@ -107,6 +107,38 @@ can be achieved using a compose file with the following entry:
...
+by [Device Authentication](https://github.com/mendersoftware/deviceauth)
+and [User Administration](https://github.com/mendersoftware/useradm) services.
+
+Before starting mender in production, one needs to provide RSA private keys in
+PEM format for both services. This can be accomplished by generating key using
+openssl:
+
+```
+openssl genpkey -algorithm RSA -out key.pem -pkeyopt rsa_keygen_bits:2048
+```
```
kacf: The original content in mender-docs lists this as the command to use:

```
openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -sha256
```

What is the difference between the two? Are they interchangable? It look like yours is only generating the private key, and not a certificate.

bboozzoo: This one generates public key (rsa, 2048 bits). The one you have generates key (rsa 4096 bits), CSR & signs the certificate. 2 different things.

bboozzoo:
I'll be merging when tests complete. Objections?

bboozzoo:
And the tests fail because the client cannot download the image the certificate cannot be verified.

ssl/storage-proxy/s3.docker.mender.io.crt
```diff
@@ -0,0 +1,18 @@
...
+0cWVpKsKsjVV9eaoze3nhZzp5O57AVzEhFTMzMopvyj1iLs9RhADO6ARIjAZmAcA
+3tdsC9hEChgMczUKykhVqfD7Efx2DJEbj1WqOv+wxPcF/urWGz6ltlUBM2lRbLHc
+0uIzJ+Q1LKyIHKZ+mj4+2kZxmwn27VY8VlPaI2HNPVBsho+uowIDAQABMA0GCSqG
+SIb3DQEBCwUAA4IBAQAWAiLBAdiTa4vB0J/OZDLg4ioCkACivIO93DVq5T8u0K9/
+u1GTfz0eaOiwspsPofjrlYq8StD6FvXzy+MMinp3JFBEH7FKksm+P+oXTpMcRGPH
+OLLdUYqwjFHViQqO30UnQvUjW4y7wjxgqyjuZTb3t4m5h0ovXLd2wVufkLH1pavk
+VPxZJc2KeZ0shHCeuxgPTbz+36Ein/H1dI4cKZ9ia7pcxIGIIgxmEmus09QiUrA3
+BhMrGSWvnXP/E62GeHd1/IfbnyECQ+vqb7T0U9HFCeBSnDKuP/lerbejZ4zPhpPr
+SizKxVb2vBhVGvfahnVMkB+CaCSwC7g2+4lY16f3
+-----END CERTIFICATE-----
```
kacf: Is this the key that should be added to the client as per https://tracker.mender.io/browse/MEN-920?

kacf: Never mind, I'm blind, it says so in the ticket.

estenberg:
Nice!
There is quite a bit of duplication with what's in the new Administration section of the docs here, though. I would rather just link to the docs section in order to not get conflicting/outdated information.

estenberg:
Also note that we use crt/key extensions here, while we used only pem (also for certs) up until now.
It doesn't seem to be a standard convention either way, but we should stick with one of them. Pick your favourite. :)

bboozzoo:
Maybe we should rename the one in API gateway, from what I've seen cert files usually have `*.crt` suffix. On the other hand, it's not like a written rule or anything. I'll open a jira ticket for renaming API gateway file to `cert.crt`

bboozzoo:
PR updated. most of the pieces that duplicated [Administration](https://docs.mender.io/Administration) guide were removed. I left the services diagram so that people at least see the different pieces without the need to jump between sites for no particular reason.

bboozzoo:
Tests are green now. Ready for merge?

PR #99, opened 2017-01-02
Properly authenticate, and no longer use session

GregorioDiStefano:
@kacf

### DIFF
--- a/tests/MenderAPI/admission.py
+++ b/tests/MenderAPI/admission.py
@@ -24,8 +24,8 @@ from MenderAPI import gateway, api_version, logger
 
 class Admission():
     s = None
-    def __init__(self, s):
-        self.s = s
+    def __init__(self, auth_header):
+        self.auth_header = auth_header
         self.admission_base_path = "https://%s/api/management/%s/admission/" % (gateway, api_version)
 
     def get_devices(self):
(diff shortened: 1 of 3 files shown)


tests/MenderAPI/admission.py
```diff
@@ -60,8 +60,9 @@ def get_devices_status(self, status=None):
 
 
     def set_device_status(self, device_id, status):
-        r = self.s.put(self.admission_base_path + "devices/%s/status" % device_id,
-                         headers={'Content-Type': 'application/json'},
+        r = requests.put(self.admission_base_path + "devices/%s/status" % device_id,
+                         verify=False,
+                         headers={'Content-Type': 'application/json', self.auth_header.keys()[0]: self.auth_header.values()[0]},
                          data=json.dumps({"status": status}))
```
kacf: I would instead do:

```
headers = dict(self.auth_header)
headers['Content-Type'] = 'application/json'
r = requests.put(self.admission_base_path + "devices/%s/status" % device_id,
                         verify=False,
                         headers=headers,
                         data=json.dumps({"status": status}))
```

This will save us a lot of headache later if we ever add any fields to the `auth_header`.

GregorioDiStefano: sorry, that is indeed ugly, will fix.

PR #111, opened 2017-01-10
Small integration testing tweaks

GregorioDiStefano:
@kacf @maciejmrowiec

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,25 @@
+Mender Integration testing
+===================
+
+
+####Requirements
+Python 2.7, requests, pytest, fabric, docker, docker-compose
+
+Install all Python requirements by running:  `pip2.7 install -r requirements.txt`
+
+Running tests is straight forward, just execute `bash run.sh`
+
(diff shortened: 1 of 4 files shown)


tests/README.md
```diff
@@ -0,0 +1,19 @@
...
+Running tests is straight forward, just execute `bash run.sh`
+
+`run.sh` is simple wrapper that  grabs the latest test images, artifacts tools, and other test requirements. It also calls `py.test` with `--runfast` and `--runslow`, which runs all fast and slow tests.
+
+####OS X
+
+Currently, running integration tests on OS X is not straight forward due to: 
+https://github.com/docker/docker/issues/22753
+
+This can be fixed by running the tests inside a docker container attached to the integration network. This is currently not implemented.
```
kacf: This won't work unfortunately, because it needs to call docker from within the test framework. The only way it'll work currently, is using a VM.

GregorioDiStefano: I explicitly state this wont work, as this isnt implemented

GregorioDiStefano: this makes sense now, we won't be able to do this.

kacf: Indeed, it will never work, so no use in mentioning it. :-)

PR #113, opened 2017-01-10
script to generate all required server keys

estenberg:
To make it easier to generate keys and self-signed certs for use in production.
@bboozzoo can you take a look, please?

### DIFF
--- a/keygen
+++ b/keygen
@@ -1,0 +1,41 @@
+#!/bin/bash
+set -e
+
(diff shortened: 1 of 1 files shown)


bboozzoo:
The script looks ok, but it generates certificates for demo hostnames. Is it something we want to do?

estenberg:
Thanks, I changed to take the CNs as environment input now, aligned with how we do it in the services as well.

PR #115, opened 2017-01-11
docker-compose.yml: fix mongo version (3.4)

mchalski:
there are a couple tags we could fix on -
'3.4.1', '3.4', '3' and 'latest' all point to the same image.

decided to fix on the minor version - this way we'll
automagically get patches for 3.4 as they come.

Issues: MEN-905


maciejmrowiec:
@kacf what do you think about this silent patch version?

PR #116, opened 2017-01-11
Added Amazon S3 integration test

GregorioDiStefano:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,6 @@ before_install:
 
 script:
     - if [ "$JOB_TYPE" = compile_and_integration_tests_fast ]; then ( cd tests && bash run.sh --runfast ); fi
-    - if [ "$JOB_TYPE" = compile_and_integration_tests_slow ]; then ( cd tests && bash run.sh --runslow ); fi
-
-    # CRON JOB SECTION
-    - if [ "$TRAVIS_EVENT_TYPE" = "cron" ]; then ( cd tests && bash run.sh --runnightly ); fi
-
 
     - if [ "$JOB_TYPE" = compile_and_integration_tests_slow ]; then
         ( cd tests && bash run.sh --runslow );
@@ -32,6 +27,9 @@ script:
     - if [[ "$JOB_TYPE" = compile_and_integration_tests_amz_s3 ]] && [[ -n $AWS_ACCESS_KEY_ID ]] && [[ -n $AWS_SECRET_ACCESS_KEY ]]; then
         ( cd tests && bash run.sh --runs3 );
       fi
+
+    # CRON JOB SECTION
+    - if [ "$TRAVIS_EVENT_TYPE" = "cron" ]; then ( cd tests && bash run.sh --runnightly ); fi
 
 notifications:
     slack:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
     matrix:
         - JOB_TYPE=compile_and_integration_tests_fast
         - JOB_TYPE=compile_and_integration_tests_slow
+        - JOB_TYPE=compile_and_integration_tests_amz_s3
 
     global:
         TEST_BRANCH: $TRAVIS_BRANCH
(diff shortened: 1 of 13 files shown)


GregorioDiStefano:
Bah, AWS credentials are available during pull requests, so we will only catch errors during the nightly.

GregorioDiStefano:
@kacf @bboozzoo

docker-compose.s3.yml
```diff
@@ -0,0 +1,18 @@
...
+    #
+    mender-deployments:
+        image: mendersoftware/deployments:latest
+        extends:
+            file: common.yml
+            service: mender-base
+        networks:
+            - mender
+        environment:
+            AWS_ACCESS_KEY_ID: %AWS_KEY%
```
bboozzoo: ${AWS_KEY} ${AWS_SECRET} will work too right? Those %..% remind me of windows, bleh.

tests/run.sh
```diff
@@ -33,8 +35,16 @@ if [[ ! -f broken_update.ext4 ]]; then
     dd if=/dev/urandom of=broken_update.ext4 bs=10M count=5
 fi
 
+if [[ -z $AWS_ACCESS_KEY_ID ]] || [[ -z $AWS_SECRET_ACCESS_KEY ]]; then
+    echo "AWS credentials (AWS_SECRET_ACCESS_KEY / AWS_SECRET_ACCESS_KEY) are not set, not running S3 tests"
+else
+    sed -i "s/%AWS_KEY%/${AWS_ACCESS_KEY_ID}/" ../docker-compose.s3.yml
```
bboozzoo: so if you replace %..% with ${AWS_KEY} etc you could just export AWS_KEY AWS_... here

docker-compose.s3.yml
GregorioDiStefano: i purposely didnt use ${VARIABLE} here because I didn't want to use bash syntax

.travis.yml
```diff
@@ -21,3 +22,8 @@ script:
     - if [ "$JOB_TYPE" = compile_and_integration_tests_slow ]; then
         ( cd tests && bash run.sh --runslow );
       fi
+
+    - if [ "$JOB_TYPE" = compile_and_integration_tests_amz_s3 ]; then
+        ( cd tests && bash run.sh --runs3 );
+      fi
```
kacf: If what you're saying is true about the credentials (and if it weren't it would be very insecure), then perhaps we should add some logic here to detect whether we're running a PR build or a master build.

tests/common_docker.py
```diff
@@ -15,7 +15,7 @@
 from fabric.api import *
 import pytest
 import subprocess
-
+import time
```
kacf: Nm, I see it's used further down.

.travis.yml
GregorioDiStefano: good point

tests/tests/amazon_s3/test_s3.py
```diff
@@ -0,0 +1,60 @@
...
+
+        Helpers.verify_reboot_performed()
+        assert Helpers.get_active_partition() == previous_inactive_part
+        deploy.check_expected_status(deployment_id, "success", len(get_mender_clients()))
+
+        for d in adm.get_devices():
+            deploy.get_logs(d["id"], deployment_id, expected_status=404)
+
+        Helpers.verify_reboot_not_performed()
+        assert Helpers.yocto_id_installed_on_machine() == expected_image_id
```
GregorioDiStefano: I've duplicated the `test_update_image_successful` test  and created this new test with the s3 flag enabled. 

I'm not sure I get your suggestion, but if I understood correctly, is change the test running logic to just check if the AWS_URI variable is set to s3 instead of using the decorator to explicitly run the test?

kacf: No, what I meant was: Use the s3 decorator on the original `test_update_image_successful` test instead, and remove this one. AFAICS, what happens in the test isn't influenced by what's in the test, but by what's in the `AWS_URI` variable. Hence you can simply run the original test = less duplication.

GregorioDiStefano: Okay, but this would mean we only run the most basic test with S3, and all other tests use minio

kacf: Not necessarily, we could expand it to run other tests as well. This is an even stronger argument to reuse existing tests instead of adding new ones.

GregorioDiStefano: Okay, I think this makes sense.

bboozzoo: IMO it's ok if we just S3 in just the basic test. We just want to verify that it still works. Using minio locally should be more efficient (especially with bandwidth limits removed) and does not count towards S3 traffic quotas.

GregorioDiStefano: We have to reset the docker-compose environment to use S3. All the basic integration tests right now use the standard setup with minio, but if the s3 flag is set, we also need to use the correct fixture to setup the integration environment with AWS.

What if I just disable `test_update_image_successful` if s3 testing is enabled and keep the separate s3 test?

kacf: I see what you mean. Final proposal then: Make a separate test, but call this test from within it. That will avoid fixtures for the called function, and will use the fixture from the calling function instead.

As you understand I'm completely allergic to code duplication. :-)

GregorioDiStefano: Makes sense! and yes, the code duplication would suck.

GregorioDiStefano:
@kacf updated

docker-compose.s3.yml
```diff
@@ -0,0 +1,19 @@
+version: '2'
```
bboozzoo: You don't need this file anymore.

tests/common_setup.py
```diff
@@ -59,3 +59,20 @@ def standard_setup_two_clients_bootstrapped():
...
     set_setup_type(ST_TwoClientsBootstrapped)
+
+@pytest.fixture(scope="function")
+def standard_setup_one_client_bootstrapped_with_s3():
+    if setup_type() == ST_OneClientsBootstrapped_AWS_S3:
+        return
+
+    stop_docker_compose()
+
+    docker_compose_cmd("-f ../docker-compose.s3.yml up -d")
```
bboozzoo: use `docker-compose.storage.s3.yml` instead

GregorioDiStefano:
@bboozzoo

PR #117, opened 2017-01-11
docker-compose: storage overlays for minio/s3

mchalski:
Issues: MEN-915

@mendersoftware/rndity @maciejmrowiec @GregorioDiStefano

PR #119, opened 2017-01-12
compose: demo: move demo setup settings to docker-compose.demo.yml

bboozzoo:
Move demo setup related settings to demo overrides file.

There's also a couple of PRs #117 #116 that extract storage to separate files. I'll rebase when any of these is merged.

### DIFF
--- a/demo
+++ b/demo
@@ -1,0 +1,12 @@
+#!/bin/bash -x
+
+
+CLIENT="-f docker-compose.client.yml"
+[ -n "${NO_CLIENT}" ] && unset CLIENT
+
+exec docker-compose \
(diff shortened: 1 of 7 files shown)


estenberg:
As mentioned I would remove the gateway port setting, even from the demo layer so it defaults to 443. The ports cannot simply be changed here in the extends file as it depends on client and UI as well. If people really need to change from the default https port we can always document that down the line.

bboozzoo:
@estenberg you have to explicitly list the ports that are published. I'll export 443 next to 8080.

bboozzoo:
merge?

PR #130, opened 2017-01-18
storage-proxy: propagate the Host header to minio

mchalski:
proxy fills out the Host header with vanilla nginx $http_host.
the address mapping of proxy and minio is the same, so this will
always yield correct results.

(note that a different approach was considered first - use env vars
and lua, but it's unnecessary)

Issues: MEN-948

### DIFF
--- a/storage-proxy/nginx.conf
+++ b/storage-proxy/nginx.conf
@@ -93,7 +93,7 @@ http {
             proxy_request_buffering off;
             proxy_http_version 1.1;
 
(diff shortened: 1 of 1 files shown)


mchalski:
approve/merge anyone?

bboozzoo:
BTW. can you verify that the setup still works if you change service alias name from `s3.docker.mender.io` to something else?

mchalski:
@bboozzoo yes, changed the domain to `storage.acme.com`, regenerated/mounted certs, uploaded and deleted an image - all successful.

PR #132, opened 2017-01-19
when verifying reboot, try to connect multiple times to the client

GregorioDiStefano:
@kacf I want to check that connecting multiple times after rebooting works, before considering the machine up and running. 

I think this might help with some of issues we are seeing that commands executed after reboot are failing due to connectivity issues.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -27,11 +27,13 @@ from fabric.contrib.files import exists
 
 logger = logging.getLogger()
 
+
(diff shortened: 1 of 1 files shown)


GregorioDiStefano:
made some changes suggested by @bboozzoo and retesting

bboozzoo:
let's merge and see if it helps

GregorioDiStefano:
ran twice, no errors found, ill merge this and see what happens

PR #136, opened 2017-01-20
Cleanup key locations

estenberg:
Must be merged with https://github.com/mendersoftware/mender-api-gateway-docker/pull/52.


bboozzoo:
Personally, not sure if `/keys` is a place where I would look for certificates. If others are ok with this change then we can merge it

estenberg:
ok, the reason is that I felt that all these key/cert files were related; you probably want to regenerate all of them in one go and you would need both a private key and a cert file for api gateway and storage proxy.

I couldn't find a good name except keys as a parent dir, and would be open to proposals!

bboozzoo:
IMO certificates should remain in some certificate related directory, `cert`, `certs`, `certificates`, `ssl` etc. `keys` would not be my first choice to look in if I were trying to locate certificates.

estenberg:
Do you have some reference from other / standard locations for similar tools?
Perhaps you could see if others in the team has an opinion or create an alternate proposal?

bboozzoo:
As I wrote earlier, if others are ok then we can merge this PR.

@mendersoftware/rndity @maciejmrowiec @pasinskim @kacf

kacf:
I have no strong preference, but I'm leaning towards using the parent dir `certs` instead of `keys` as well.

estenberg:
Alright.. The logical reason was that a cert embeds a key, but a key is not a cert.
So if we want a single dir, then it should probably be called keys (or something generic), not certs.
The main purpose of a single dir is that people would probably regenerate all these (not just the certs, for example).

But what about this structure:

certs/
├── api-gateway
│   ├── cert.crt
│   └── private.key
└── storage-proxy
    ├── cert.crt
    └── private.key

keys/
├── deviceauth
│   └── private.key
└── useradm
    └── private.key

estenberg:
@bboozzoo I updated the PR with new locations.
If this is ok and merged I will update keygen script & docs accordingly.

docker-compose.demo.yml
```diff
@@ -23,8 +23,8 @@ services:
         environment:
             MAPPED_PORT: 8080
         volumes:
-            - ./ssl/mender-api-gateway/cert.pem:/var/www/mendersoftware/cert/cert.pem
-            - ./ssl/mender-api-gateway/key.pem:/var/www/mendersoftware/cert/key.pem
+            - ./certs/api-gateway/cert.crt:/var/www/mendersoftware/cert/cert.crt
+            - ./certs/api-gateway/private.key:/var/www/mendersoftware/cert/private.key
```
estenberg: @bboozzoo I think this should be ready to merge with https://github.com/mendersoftware/mender-api-gateway-docker/pull/52/files already; they should be mounted to the right place, no?

bboozzoo: yup, we shoould merge both then

PR #139, opened 2017-01-24
deployment template

bboozzoo:
A template for use in production. There's a couple of assumptions that come with using this. First, project documentation will tell the end users how to consume the template for their own needs.

Next, I expect that the user will copy the template (`cp -r template mysetup`) and applly/modify relevant files there. This will be easier wrt. git pulls as there will be no conflicts raised when applying upstream changes.

The user is also expected to use the `keygen` utility that @eysteins added. This will generate required keys and certificates that later get mounted into specific containers.

Lastly, the user *needs* to do some editing (come up with key and secret for minio), find out the domain name he/she is using to publish mender backend.

I'm also considering bumping required compose version to 1.10. This allows us to use docker stacks.

### DIFF
--- a/template/prod.yml
+++ b/template/prod.yml
@@ -1,0 +1,101 @@
+# this is a template file for production setup, consult
+# https://docs.docker.com/compose/compose-file/ for details on syntax and usage
+#
+# Notes:
+# - integration/docker-compose.yml file is assumed to be included
+# - integration/docker-compose.storage.minio.yml is assumed to be included
+# - all services are part of `mender` network (service names are unchanged)
+# - keys and certificates are generated using keygen utility from integration
+#   repository, keys and certificates are stored in ./keys-generated directory
+# - certificates and key are mounted into containers using volumes
+# - minio artifacts are stored in a named volume `mender-artifacts`; volume
+#   needs to be created manually using `docker volume create mender-artifacts`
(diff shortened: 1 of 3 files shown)


template/prod.yml
```diff
@@ -0,0 +1,101 @@
+# this is a template file for production setup, consult
+# https://docs.docker.com/compose/compose-file/ for details on syntax and usage
+#
+# Notes:
+# - integration/docker-compose.yml file is assumed to be included
+# - integration/docker-compose.storage.minio.yml is assumed to be included
```
maciejmrowiec: what if I want to use S3 ?

bboozzoo: Then you're supposed to read the manual. We're not going to provide a wizard.

maciejmrowiec:
yea, I'll try to add to this when I'll get around that one, for sure volumes for persistence, db will be on local network so skipping certificates

estenberg:
Cool.

@bboozzoo  do you plan to add docs for this as well as part of prod setup (https://docs.mender.io/Administration/Production-installation)?
If not then we need some followups.

bboozzoo:
Yes, there's a task for this. I've already taken it in this sprint.

bboozzoo:
merge?

maciejmrowiec:
yep

PR #143, opened 2017-01-25
Add docker-compose scale option to demo script

GregorioDiStefano:
Scale mender clients with SCALE env. variable, and updated documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,9 @@ Requirements:
 To start using Mender, we recommend that you begin with the Getting started
 section in [the Mender documentation](https://docs.mender.io/).
 
+To quickly get started, and demo mender, run `./up`
+
+If you want to test with multiple devices (default=1) set the SCALE variable as such: `SCALE=3 ./up`
 
 ## Services
 
--- a/demo
+++ b/demo
@@ -4,6 +4,13 @@
 CLIENT="-f docker-compose.client.yml"
 [ -n "${NO_CLIENT}" ] && unset CLIENT
 
+if [ ! -z "$SCALE" ]; then
+   docker-compose \
+    -f docker-compose.yml \
+    -f docker-compose.client.yml \
+    scale mender-client="$SCALE"
+fi
+
 exec docker-compose \
      -f docker-compose.yml \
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -25,6 +25,9 @@ Requirements:
 To start using Mender, we recommend that you begin with the Getting started
 section in [the Mender documentation](https://docs.mender.io/).
 
+To quickly get started, and demo mender, run the `demo` script included in this repository.
```
maciejmrowiec: how  `up` and `demo` differ?

GregorioDiStefano: Good point, `up` needs to be initially called, since it setups up entries in the hosts file, etc
It eventually calls the demo script. 

Ill fix this

maciejmrowiec: noticed this script today and got confused a bit what should be used

maciejmrowiec: so if I run only `./demo` demo won't work because of hosts file setup missing ?

GregorioDiStefano: correct

demo
```diff
@@ -4,6 +4,13 @@
 CLIENT="-f docker-compose.client.yml"
 [ -n "${NO_CLIENT}" ] && unset CLIENT
 
+if [ ! -z "$SCALE" ]; then
```
bboozzoo: You can pass scale directly in arguments for `demo` like this: `./demo scale mender-client=3`

up
```diff
@@ -46,7 +46,6 @@ setuphosts() {
 }
 
 rundocker() {
-    # sudo -E docker-compose scale mender-client=3
     sudo -E ./demo up
```
bboozzoo: calling `./demo scale mender-client=3` before `./demo up` should be enough

GregorioDiStefano: I think it's easier to set an environment variable stating how many clients you want instead of knowing the scale argument of docker-compose.

bboozzoo: I disagree, your readme update states that SCALE is an env variable for `up`. `demo` is just a wrapper around docker-compose and it's only purpose is to include respective compose files. We also don't mention `demo` script in the readme because there are `up/down/update/reset`

PR #145, opened 2017-01-27
production template: more updates

bboozzoo:
Explicitly specify project name instead of letting docker-compose name it after the directory the files were in.

Note that `mender-production` becomes `menderproduction` when `docker-compose` creates networks and containers.

@maciejmrowiec @mendersoftware/rndity


bboozzoo:
I might be adding a bit more commits so let's keep this PR open for now

PR #148, opened 2017-02-01
compose/demo: add mongo data volume mounts

mchalski:
we already have them in prod.yml, but it's useful to also have them in demo -
the user might be trying out the system and recreating containers along the
way. it would be unexpected to lose whatever data was already there.

Issues: MEN-879

@maciejmrowiec @mendersoftware/rndity

### DIFF
--- a/docker-compose.demo.yml
+++ b/docker-compose.demo.yml
@@ -68,3 +68,19 @@ services:
         environment:
             # enable demo mode for UI ["true"/"false"]
             DEMO: "true"
+
+    mender-mongo-deployments:
+        volumes:
+            - /var/mender/data:/data:rw
+    mender-mongo-device-adm:
+        volumes:
+            - /var/mender/data:/data:rw
+    mender-mongo-device-auth:
+        volumes:
+            - /var/mender/data:/data:rw
+    mender-mongo-inventory:
+        volumes:
+            - /var/mender/data:/data:rw
+    mender-mongo-useradm:
+        volumes:
+            - /var/mender/data:/data:rw


maciejmrowiec:
@mchalski if we go DB persistence in demo, we need to make sure minio is also persistent otherwise you will be remembering artifacts in db which files were purged with setup reset 
@eysteins with this you do get presistence across demos (not clean setup on every start), something you'd like to have or better to keep away?

bboozzoo:
There is some misunderstanding. Data is persistent in the demo setup. It's just we're not using named volumes or anything similar, so tracking down where the data ends up being stored is a bit more complicated (`docker inspect` the container, locate volumes, find your mount path, identify the volume, inspect the volume).

maciejmrowiec:
not sure if I get you, does compose setup volumes implicitly? - I don't see any volumes configured  for minio in compose files

bboozzoo:
Yes, anything that is listed as `VOLUME ...` in dockerfile will make docker create a volume. If you don't use a named volume there (explicitly by passing `-v <namedvolume>:..` ), then docker will create anonymous volume. The volume will be referenced by the container. Once you remove the container (as in `docker rm..` or `docker-compose down`), the anonymous volume will not be referenced by anything. You can list them with `docker volume ls --filter dangling=true`. Volumes are not removed, unless the users asks for it

When bringing demo setup up with `./demo up`, services which have volumes in their Dockerfiles will get anonymous volumes that are automatically created by docker. Now, when you do `./demo down`, containers are removed and volumes are left behind (i.e. dangling, listed by `docker volume ls --filter dangling=true`). When you `./demo up` again, a new set of anonymous volumes will be created (all of them empty). Thus from user's point of view, it may look like `./demo down` has removed the volumes, but in reality, those volumes are still there, just the new instances of containers created by `up` got a new set of anonymous volumes.

maciejmrowiec:
In this case when we restart mongo we have fixed mounts specified in volume se during restart it will mount same database files, while minio restart will detach that volume and it will get clean one next time (db&minion go out of sync)

bboozzoo:
Then I suppose we should fix our minio fork and add VOLUME /export in Dockerfile, perhaps fix entrypoint/CMD as well.

bboozzoo:
Just to be clear, in my view there is no need for named volumes in demo setup for any of the containers.

mchalski:
hmm, ok. one last for-dummies question to wrap it all up:

* if smn plays with the demo, and only does `./demo up` + `./demo stop` they'll never lose data (plus - they won't have any dangling volumes)

* if they do `./demo up` + `./demo down` they  lose data (+get dangling volumes)

* this behavior can be utilized during conferences etc. to start with a clean slate or keep some stuff between demos.

ack me on this one and I'll just close this PR, because no named volumes are necessary indeed.

estenberg:
I don't follow all of this, but I think the current way seems to work well.
We have an explicit script to remove state: https://github.com/mendersoftware/integration/blob/master/reset, it will also warn that the state will be removed (docker compose will).

Then we document to use ./up and Ctrl-C or ./stop (https://github.com/mendersoftware/integration/blob/master/stop) if they want to save state.

bboozzoo:
So, close?

maciejmrowiec:
closing, cleaning up PR queue

PR #152, opened 2017-02-01
generate new keys using keygen utility

estenberg:
After we got the [keygen utility](https://github.com/mendersoftware/integration/blob/master/keygen), I generated new certs and keys and think we should use them in demo environment so that the keys generated by keygen are well-tested.

Generated new certs for API Gateway and Storage Proxy as output from the [keygen utility](https://github.com/mendersoftware/integration/blob/master/keygen).

Must be merged with https://github.com/mendersoftware/meta-mender/pull/244


estenberg:
@GregorioDiStefano ping, in relation to https://github.com/mendersoftware/meta-mender/pull/244

PR #154, opened 2017-02-03
Revert "generate new keys using keygen utility"

maciejmrowiec:
Reverts mendersoftware/integration#152

Breaks builds, reverting.


maciejmrowiec:
Reverting only integration setup part, keeping new keys in client repo, once we get new version on qemu client using new keys available, revert the revert to get it synced.

maciejmrowiec:
@eysteinsm @GregorioDiStefano  ^^

estenberg:
Sorry about that, I was not aware that it would take that long to "sync" these changes.
To that end I can not comment much on this.

kacf:
Normally it doesn't take that long, but because of the recent Yocto breakages, the client has been lagging behind for a long time. However, today it looks like we might finally be getting back on track.

maciejmrowiec:
not applying anymore, client did catch up now

PR #159, opened 2017-02-07
Make integration master point to master service images.

maciejmrowiec:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     # mender-deployments
     #
     mender-deployments:
-        image: mendersoftware/deployments:latest
+        image: mendersoftware/deployments:master
         extends:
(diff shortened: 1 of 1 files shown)


maciejmrowiec:
FYI: intending to point lastest into latest production releases (moved manually), pointing master integration explicitly towards master images

maciejmrowiec:
@GregorioDiStefano

maciejmrowiec:
@GregorioDiStefano any ideas why this would be failing?
BTW this not getting in may disrupt other builds as `latest` docker tag is not updated anymore only master is

GregorioDiStefano:
we are matching only "mendersoftware/api-gateway" and by default, this will only match "latest" 
we need to add an argument here to specify exactly which tag. 

i can fix tomorrow morning

PR #160, opened 2017-02-07
Make it an error to provide build dir that doesn't contain what we need

kacf:
It's important not to fall back to downloading in this case, because
this can lead to subtle and difficult-to-reproduce errors because the
version differs from what is expected.

With this patch we grab everything from the build if the BUILDDIR is
specified, and we fail if this does not work. Only if the variable is
not set do we resort to downloading. In both cases we expect to find
executables in the path, rather than in the current directory.

Also do some of the preparation steps regardless of whether the files
already exist. It's not an expensive operation, and it's better to be
sure they are up to date.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,11 @@
 *.pyc
 
 tags
+
 tests/.cache/
 tests/broken_update.ext4
 tests/core-image-full-cmdline-vexpress-qemu-broken-network.ext4
 tests/core-image-full-cmdline-vexpress-qemu.ext4
 tests/large_image.dat
 tests/results.xml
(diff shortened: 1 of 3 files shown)


kacf:
@GregorioDiStefano: This passed in Travis. It failed in Jenkins, but the failure looks [completely unrelated](https://ci.cfengine.com/job/yoctobuild-pr/331/consoleFull), and is long after this script would have failed, had it been wrong. I think it's a spurious error.

kacf:
And in fact, it is the same as the error in https://github.com/mendersoftware/mender/pull/186. Even more evidence it's unrelated.

PR #163, opened 2017-02-08
no-ssl: export minio on non-SSL port

bboozzoo:
Change minio's command line to start listening on port 9001. Avoids conflict
with storage-proxy which is already listening on 9000.

Override mender-deployments env and point it to minio.s3.docker.mender.io:9001.


bboozzoo:
rebased

PR #164, opened 2017-02-08
match any container tag when getting IP of docker container

GregorioDiStefano:


GregorioDiStefano:
Previously this would only match "latest", now it will match any container name

PR #169, opened 2017-02-09
compose: demo: template: use a single mongo instance for all services

bboozzoo:
Change the baseline docker-compose file to not include any mongo service
definitions.

Use a single mongo instance in demo setup.

Production template changed to use separate mongo for each service. Ultimately
in single host deployments we could also go with only one mongo instance.

Let's discuss options.

### DIFF
--- a/demo
+++ b/demo
@@ -9,4 +9,4 @@ exec docker-compose \
      -f docker-compose.storage.minio.yml \
      -f docker-compose.demo.yml \
      $CLIENT \
-     $*
+     "$@"
--- a/docker-compose.demo.yml
+++ b/docker-compose.demo.yml
@@ -4,10 +4,22 @@ services:
     mender-useradm:
         volumes:
             - ./keys/useradm/private.key:/etc/useradm/rsa/private.pem
+        depends_on:
+            - mongo-common
 
     mender-device-auth:
         volumes:
             - ./keys/deviceauth/private.key:/etc/deviceauth/rsa/private.pem
+        depends_on:
+            - mongo-common
+
+    mender-inventory:
+        depends_on:
+            - mongo-common
+
+    mender-device-adm:
+        depends_on:
+            - mongo-common
 
     mender-api-gateway:
         ports:
@@ -48,6 +60,8 @@ services:
             DEPLOYMENTS_AWS_AUTH_KEY: minio
             DEPLOYMENTS_AWS_AUTH_SECRET: minio123
             DEPLOYMENTS_AWS_URI: https://s3.docker.mender.io:9000
+        depends_on:
+            - mongo-common
 
     minio:
         networks:
@@ -62,3 +76,23 @@ services:
         environment:
             # enable demo mode for UI ["true"/"false"]
             DEMO: "true"
+
+    #
+    # single mongo instance for all services, extend as required
+    # during deployment
+    #
+    mongo-common:
+        image: mongo:3.4
+        networks:
+            mender:
+                aliases:
+                    - mongo-deployments
+                    - mongo-device-adm
+                    - mongo-device-auth
+                    - mongo-inventory
+                    - mongo-useradm
+                    - mender-mongo-deployments
+                    - mender-mongo-device-adm
+                    - mender-mongo-device-auth
+                    - mender-mongo-inventory
+                    - mender-mongo-useradm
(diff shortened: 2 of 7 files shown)


mchalski:
imo it's ok, for saas it's 1 cluster = N DBs = N tenants, so it only makes sense to have 1 cluster onprem (nevermind it's jsut a single node cluster, but still; at some point at least a replica set should be introduced).

maciejmrowiec:
This crossed my mind lately as well, makes sense for single server setups - not too much benefit there outside from slight more isolation - demo & prod. Make it less complicated in managing.

SaaS is another story, but as long as those stick to complete separation on application level we are good.

maciejmrowiec:
@bboozzoo how about including this in prod template as well, now as this is dedicated single server install there is no benefit of having those separated

bboozzoo:
ok, will update

bboozzoo:
it's going to be a post 1.0 change, right?

maciejmrowiec:
if we do this it needs to be 1.0.0, if not then effort for building migration for 1.1.0 may not be worth of continuing with the change (for production template)

maciejmrowiec:
@pasinskim 1.0.0 candidate

kacf:
IMHO this is too late and too big a change for 1.0.0. We are supposed to build the final tomorrow.

maciejmrowiec:
is it? it's just configuration thing aka point everyone to same db server, but I don't think this is worth the effort if we will need to write migrations, so it's merge before 1.0.0 or give up on this whole thing for now

bboozzoo:
Updated production setup to use a single mongo instance with `mender-mongo-data` volume.

If we go with this change, an update to documentation will be needed as well (mostly output of docker commands).

kacf:
> is it? it's just configuration thing aka point everyone to same db server

So we used to have many different db servers, now we will move to having only one? This sounds like a big change to me, with lots of breakage potential. I'm not against the change in general, but we are supposed to build a release package *today*. Clearly too late if you ask me.

maciejmrowiec:
In case you are not comfortable with this one so late then lets postpone it then.
Change is not as big as it can seem, Mongo does not have any db creation scripts or anything like this that could be impacted it creates things implicitly on inserts.

kacf:
If it was important for functionality, it might have different opinion, but yeah, I think we should drop this for 1.0.0. We have enough struggles. :-)

bboozzoo:
How about we merge this for 1.1.0?

As for migration, I think this will need to be covered in release notes for 1.1.0. I'm not sure where these notes should be put though. I'm guessing a changelog entry in commit message will be not visible enough.

maciejmrowiec:
Maybe we can wait with merging until we have upgrade documentation that includes this migration. I would prefer to avoid not being able to release because we forgot about preparing docs / scripts covering this.

Doc similar to this: https://tracker.mender.io/browse/MEN-1112

There is a ticket for this: https://tracker.mender.io/browse/MEN-1114
We can prioritise this a bit higher to get it out of the way.

bboozzoo:
I've pushed some helpers to ease DB migration. The process is as follows:
1. run `migration/dump-1.0-db` in a pre single-mongo setup 
2. pull changes
3. start mongo-common (optionally create named volumes as needed)
4. run `migration/restore-db`
5. start the rest of services

https://github.com/mendersoftware/integration/pull/169/commits/413c6f72e4bd928ed30e9802292c364cde9bcc27 https://github.com/mendersoftware/integration/pull/169/commits/7efcde282b2e1e537dfd086a1619fb9d1a4d5229 https://github.com/mendersoftware/integration/pull/169/commits/8b1ac608ecf4a1579eb870f2a0d6acd1ecb34d6d need to be cherry picked to 1.0.x branch.

We really should have done it before 1.0. Actually I'm not even sure I really want to have this change merged at all right now. I'm using it locally, and will keep doing so regardless of whether this hits master.

maciejmrowiec:
how will this look for 1.0.0 upgrade that does not have those cherry-picked ?

bboozzoo:
You'll have to dump the data yourself and prepare it for import with mongorestore.

maciejmrowiec:
we need to make sure that upgrade from 1.0.0 and 1.0.1 to 1.1.x have identical procedure, otherwise this is just going to be a nightmare

bboozzoo:
Ok, let's close this PR for now and discuss it offline.

PR #171, opened 2017-02-10
Correct ext4 and remove useless update parameter

GregorioDiStefano:
@kacf

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -x -e
 
+# make sure we are using the right ext4 image when testing
+RELEASE_VERSION="master"
+
 if [[ ! -f large_image.dat ]]; then
     dd if=/dev/zero of=large_image.dat bs=200M count=0 seek=1
 fi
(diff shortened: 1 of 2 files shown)


tests/run.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -x -e
 
+# make sure we are using the right ext4 image when testing
+RELEASE_VERSION="master"
```
kacf: Don't you want to inherit this from the environment?

GregorioDiStefano: initially i did, but what happens when someone makes a PR?

GregorioDiStefano:
@kacf

tests/run.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/bash
 set -x -e
 
-# make sure we are using the right ext4 image when testing
-RELEASE_VERSION="master"
+# we need to make sure we use the correct ext4 image for testing
+if [[ -z "$TEST_BRANCH" ]]; then
```
kacf: Actually this should be:

```
if [[ -z "$BUILDDIR" ]] && [[ -z "$TEST_BRANCH" ]]; then
```

Because if `BUILDDIR` is set, you are getting your artifacts from the build, not from any branch.

GregorioDiStefano: fixed, nice find

PR #179, opened 2017-02-16
Fix incorrect writing of file into ext4 using debugfs.

kacf:
Debugfs really deserves its name, "debug": Writing into a file name,
"/etc/hosts", will create a literal file with that name, including
slashes and everything, in the root! The slash is not a valid
filesystem character, so this corrupts the filesystem, and certainly
the file won't be found in the correct place.

Work around this by using the debugfs "cd" feature to get to the
correct directory first. Unfortunately you can't pass more than one -R
parameter, so we need to pipe the commands into it instead.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -73,7 +73,9 @@ class Helpers:
             output = subprocess.check_output(cmd, shell=True).strip()
             logging.info("Running: " + cmd + " returned: " + output)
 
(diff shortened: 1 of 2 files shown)


kacf:
Testing in Jenkins.

kacf:
Merging immediately, since this did fix the issue that's currently breaking all builds. I don't yet know if the whole build will pass, but I'll keep an eye on it and post additional commits if that happens.

PR #180, opened 2017-02-16
Merging master to 1.0.x

pasinskim:


maciejmrowiec:
wow, is this safe? this contains already cherry-picked commits

kacf:
Normally no, but we checked specifically.

PR #182, opened 2017-02-16
small change in download retry verification logic

GregorioDiStefano:
@pasinskim @kacf

### DIFF
--- a/tests/tests/test_fault_tolerance.py
+++ b/tests/tests/test_fault_tolerance.py
@@ -33,9 +33,9 @@ class TestFaultTolerance(MenderTesting):
 
         while int(time.time()) < timeout_time:
             with quiet():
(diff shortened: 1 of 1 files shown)


tests/tests/test_fault_tolerance.py
```diff
@@ -33,9 +33,9 @@ def wait_for_download_retry_attempts(self):
 
         while int(time.time()) < timeout_time:
             with quiet():
-                output = run("journalctl -u mender -l --no-pager | grep 'update fetch request failed' | wc -l")
+                output = run("journalctl -u mender -l --no-pager | grep 'update fetch failed' | wc -l")
```
kacf: I suggest to match this exactly: `msg="update fetch failed:`. It guarantees that the error can not bubble up and cause errors later.

PR #184, opened 2017-03-01
remove redundant success/fail tests

GregorioDiStefano:


GregorioDiStefano:
Instead of running  test_update_image_successful and  test_update_image_failed as individual tests, they will be test in conjunction in test_double_update and  test_failed_updated_and_valid_update

PR #186, opened 2017-03-01
make sure deployment statuses are correctly reported

GregorioDiStefano:


GregorioDiStefano:
@maciejmrowiec this it to make sure we don't get that surprise bug again (status not properly reported)

PR #188, opened 2017-03-02
use patched version of fabric

GregorioDiStefano:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,8 @@ env:
 
 before_install:
     - sudo apt-get -qq update && sudo apt-get -qq install -y e2tools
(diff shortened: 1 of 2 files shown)


GregorioDiStefano:
There is an issue with fabric/paramiko which results in us having to catch fabric exiting, and seeing: 

"Fatal error: Needed to prompt for a connection or sudo password (host: 172.21.0.5:8822), but abort-on-prompts was set to True"

everywhere.

I patched fabric to retry before asking for password, and opened a PR with the project: 

https://github.com/fabric/fabric/pull/1572

kacf:
Seems to be working fine now.

PR #192, opened 2017-03-07
Initial revision of license-overview-generator.

kacf:


kacf:
This was used to generate the 1.0.0 license usage text.

PR #193, opened 2017-03-10
tests: integration: adapt to admission/device authentication API changes

bboozzoo:
Device admission does not list devices, but device authentication data
sets (each with a distinct ID). Auth data set ID is not the same as device ID.
However device ID (the one that owns given auth set) is available as device_id
field in objects returned by the API.

@GregorioDiStefano @mendersoftware/rndity


GregorioDiStefano:
looks good to me

bboozzoo:
'fast' tests are passing with locally built images

PR #201, opened 2017-03-22
add conductor

mchalski:
Issues: MEN-1097

this was also tested with a real workflow uploaded to conductor, pinging the inventory service for devices. it seems that conductor can talk to our services just fine.

see comment in https://tracker.mender.io/browse/MEN-1097 to replicate the test.

### DIFF
--- a/conductor/server/config/config.properties
+++ b/conductor/server/config/config.properties
@@ -1,0 +1,39 @@
+# Database persistence model.  Possible values are memory, redis, and dynomite.
+# If ommitted, the persistence used is memory
+#
+# memory : The data is stored in memory and lost when the server dies.  Useful for testing or demo
+# redis : non-Dynomite based redis instance
+# dynomite : Dynomite cluster.  Use this for HA configuration.
+
+db=dynomite
+
+# Dynomite Cluster details.
+# format is host:port:rack separated by semicolon
+workflow.dynomite.cluster.hosts=mender-dynomite:8102:dummyregion
+
+# Dynomite cluster name
+workflow.dynomite.cluster.name=mender-dynomite
+
+# Namespace for the keys stored in Dynomite/Redis
+workflow.namespace.prefix=conductor
+
+# Namespace prefix for the dyno queues
+workflow.namespace.queue.prefix=conductor_queues
+
+# No. of threads allocated to dyno-queues (optional)
+queues.dynomite.threads=10
+
+# Non-quorum port used to connect to local redis.  Used by dyno-queues.
+# When using redis directly, set this to the same port as redis server
+# For Dynomite, this is 22122 by default or the local redis-server port used by Dynomite.
+queues.dynomite.nonQuorum.port=22122
+
+
+# Transport address to elasticsearch
+workflow.elasticsearch.url=mender-elasticsearch:9300
+
+# Name of the elasticsearch cluster
+workflow.elasticsearch.index.name=conductor
+
+# For single node dynomite, must be the same as 'rack'
+EC2_AVAILABILITY_ZONE=dummyregion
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -129,5 +129,30 @@ services:
                 aliases:
                     - mongo-useradm
 
+    mender-conductor:
```
maciejmrowiec: what do you think about having conductor in separate compose file similarly to minio+nginx ?

mchalski: imo not a good idea. conductor is both mandatory and not swappable, so it has to be there always as is.

docker-compose.yml
```diff
@@ -129,5 +129,30 @@ services:
...
+        image: mendersoftware/conductor:1.7.0
+        depends_on:
+            - mender-elasticsearch
+            - mender-dynomite
+        networks:
+            - mender
+
+    mender-conductor-ui:
```
bboozzoo: Do we need this UI for everything to work? Maybe we could put this into a separate compose file in {{extra}}

template/prod.yml
```diff
@@ -118,6 +118,18 @@ services:
             # mounts a docker volume named `mender-artifacts` as /export directory
             - mender-artifacts:/export:rw
 
+    mender-conductor:
+        volumes:
+            - ./conductor/server/config:/app/config
+        environment:
+            - CONFIG_PROP=config.properties
+
+    mender-conductor-ui:
```
bboozzoo: Again, not sure we should add even add conductor-ui by default.

conductor/server/config/config.properties
```diff
@@ -0,0 +1,36 @@
...
+#
+# memory : The data is stored in memory and lost when the server dies.  Useful for testing or demo
+# redis : non-Dynomite based redis instance
+# dynomite : Dynomite cluster.  Use this for HA configuration.
+
+db=dynomite
+
+# Dynomite Cluster details.
+# format is host:port:rack separated by semicolon
+workflow.dynomite.cluster.hosts=mender-dynomite:8102:us-east-1c
```
bboozzoo: us-east-1c?

docker-compose.yml
maciejmrowiec: ack

conductor/server/config/config.properties
mchalski: yeah, this is just to make the parser happy...you have to pass a region (a 'rack' in coductor-speak).

I could set it to `dummyregion` since `us-east` may be confusing

bboozzoo: `dummyregion` sounds ok

mchalski: uhh, just changing to `dummyregion` doesn't work, as if the `us-east-1c` is the single blessed value.

I found this bit of comment however:
https://github.com/Netflix/conductor/blob/2ae02585087f8808d2da52188d854b503339031a/server/src/main/resources/server.properties#L39

after setting `EC2_AVAILABILITY_ZONE=dummyregion` it works. but I'm not sure it's not even more confusing, at some point we'll have to figure out what it means exactly.

mchalski:
@bboozzoo all fixes done - ui removed to a different layer, region fixed, volumes added.

one note however regarding dynomite persistence - something seems wrong with redis itself, it doesn't persist any data. most of conductor's operational data is kept in redis, so the net effect is 'nothing is persisted across compose restarts'. 

`/etc/redis/redis.conf` says that snapshots are enabled, but even after waiting for a long time, no `.rdb` file was dumped to the fs. I can only trigger one with the `redis-cli` (and then it goes to `/dynomite`, even though it should go to `/var/lib/redis`).

I'll investigate it some more, in the meantime please check if the fixes at least *look* ok ;)

mchalski:
ping - added redis persistence on top of other fixes

PR #202, opened 2017-03-24
Load workflow definitions on conductor startup

kjaskiewiczz:
Custom entrypoint added.
Entrypoint statrts conductor, then loads workflow definitions
from workflow directory posting them to conductor API using curl.
All *.json file from workflows directory will be uploaded to conductor.

@bboozzoo @maciejmrowiec @mchalski

### DIFF
--- a/conductor/server/entrypoint/start_conductor.sh
+++ b/conductor/server/entrypoint/start_conductor.sh
@@ -1,0 +1,21 @@
+#!/bin/bash -e
+
+# start conductor
+/app/startup.sh &
+
+# wait for conductor
+sleep 60
+
+shopt -s nullglob
+
+# load task definitions
+for task in /srv/tasks/*.json; do
+    curl -v -X POST -H "Content-Type: application/json" http://localhost:8080/api/metadata/taskdefs -d @$task
+done
+
+# load workflow definitions
+for workflow in /srv/workflows/*.json; do
+    curl -v -X POST -H "Content-Type: application/json" http://localhost:8080/api/metadata/workflow -d @$workflow
+done
+
+wait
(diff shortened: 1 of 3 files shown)


conductor/server/entrypoint/start_conductor.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
bboozzoo: `#!/bin/bash` -e or `set -e`

conductor/server/entrypoint/start_conductor.sh
```diff
@@ -0,0 +1,11 @@
...
+/app/startup.sh &
+sleep 60
+
+# load workflow definitions
+shopt -s nullglob
+for workflow in /srv/workflows/*.json; do
+    curl -X POST -H "Content-Type: application/json" http://localhost:8080/api/metadata/workflow -d @$workflow
```
mchalski: please note 2 things:
* if a given workflow already exists, 409 conflict will be returned - do we want to handle this?
* workflows a re uploaded in 2 step, first task definitions, then workflow definitions, see e.g.:
https://tracker.mender.io/browse/MEN-1097?focusedCommentId=78777&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-78777

conductor/server/entrypoint/start_conductor.sh
kjaskiewiczz: done

conductor/server/entrypoint/start_conductor.sh
kjaskiewiczz: task definitions loading added;
I'm not sure if we want to handle 409

PR #203, opened 2017-03-27
conductor: device decommissioning workflow

mchalski:
Issues: MEN-1099

Changelog: None

### DIFF
--- a/conductor/server/tasks/delete_device_deployments.json
+++ b/conductor/server/tasks/delete_device_deployments.json
@@ -1,0 +1,12 @@
+[
+    {
+        "name": "delete_device_deployments",
+        "retryCount": 5,
+        "timeoutSeconds": 10,
+        "inputKeys": [],
+        "outputKeys": [],
+        "timeoutPolicy": "RETRY",
+        "retryLogic": "EXPONENTIAL_BACKOFF",
+        "retryDelaySeconds": 60
+    }
+]
--- a/conductor/server/workflows/decommission_device.json
+++ b/conductor/server/workflows/decommission_device.json
@@ -1,0 +1,54 @@
+{
+    "name": "decommission_device",
+    "description": "Removes device info from all services.",
+    "version": 1,
+    "tasks": [
+        {
+            "name": "delete_device_inventory",
+            "taskReferenceName": "delete_device_inventory",
+            "inputParameters": {
+                "http_request": {
+                    "uri": "http://mender-inventory:8080/api/0.1.0/devices/${workflow.input.device_id}",
+                    "method": "DELETE",
+                    "headers": {
+                        "X-MEN-RequestID": "${workflow.input.request_id}"
+                    }
+                }
+            },
+            "type": "HTTP"
+        },
+        {
+            "name": "delete_device_deployments",
+            "taskReferenceName": "delete_device_deployments",
+            "inputParameters": {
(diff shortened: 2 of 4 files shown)


conductor/server/tasks/delete_device_deployments.json
```diff
@@ -0,0 +1,13 @@
...
+    {
+        "name": "delete_device_deployments",
+        "retryCount": 5,
+        "timeoutSeconds": 10,
+        "inputKeys": [],
+        "outputKeys": [],
+        "timeoutPolicy": "RETRY",
+        "retryLogic": "EXPONENTIAL_BACKOFF",
+        "retryDelaySeconds": 60,
+        "responseTimeoutSeconds": 3600
```
kjaskiewiczz: What's the meaning of "responseTimoeutSeconds" parameter?
Why this timeout is set to one hour?

mchalski: `if greater than 0, the task is rescheduled if not updated with a status after this time. Useful when the worker polls for the task but fails to complete due to errors/network failure.`

I understand it as - the task was scheduled, the worker didn't pick it up (or failed to report state, e.g. `IN_PROGRESS`), so we time it out after X seconds.

so I don't think it even applies here - we don't have our own workers, `HTTP_TASK` is a system task. btw. `3600` is a value from one of the examples. I think I'll set it to 0 or omit it altogether.

the difference vs `timeoutSeconds` - this timeout is applied only when the task is already in the `IN_PROGRESS` state. so this is what is intuitively understood as a timeout - we know the task is being performed, and we give it X seconds to complete.

https://netflix.github.io/conductor/metadata/#task-definition

conductor/server/workflows/decommission_device.json
```diff
@@ -0,0 +1,44 @@
...
+    "description": "Removes device info from all services.",
+    "version": 1,
+    "tasks": [
+        {
+            "name": "delete_device_inventory",
+            "taskReferenceName": "delete_device_inventory",
+            "inputParameters": {
+                "http_request": {
+                    "uri": "http://mender-inventory:8080/api/0.1.0/devices/${workflow.input.device_id}",
+                    "method": "DELETE"
```
bboozzoo: How about including X-Mender-RequestId in request headers? This would allow us to record that certain workflow and associated requests to individual service are made on behalf of some external request identified by requestid header.

mchalski: done, the workflow accepts a `request_id` param and passes it on via the usual `X-MEN-RequestID` header.

also @kjaskiewiczz

conductor/server/workflows/decommission_device.json
```diff
@@ -0,0 +1,54 @@
+{
+    "name": "decommission_device",
+    "description": "Removes device info from all services.",
+    "version": 1,
+    "tasks": [
```
bboozzoo: I think it should be possible to refactor these to FORK/JOIN with requests to respective service running in parallel.

mchalski: not straightforward at least, tried for some time but failed.

conductor/server/workflows/decommission_device.json
maciejmrowiec: pretty cool

PR #204, opened 2017-03-30
Add release_tool.py.

kacf:
This script is meant to aid in the release process. As we need to
submit this is in order to test it during a release, it doesn't carry
much documentation at the moment, but this will come in a later
commit.

Changelog: None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 
 tags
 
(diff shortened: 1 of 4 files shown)


kacf:
Submitting this straight to 1.0.x, since it will be tested there. Will cherry-pick to master later.

PR #207, opened 2017-04-06
start 'docker-compose up' sequentially

GregorioDiStefano:
Changelog: none

### DIFF
--- a/tests/common_docker.py
+++ b/tests/common_docker.py
@@ -19,6 +19,9 @@ import tempfile
 import time
 import conftest
 from common import *
+import random
+import inspect
+import filelock
 
 COMPOSE_FILES = [
     "../docker-compose.yml",
@@ -28,6 +31,8 @@ COMPOSE_FILES = [
 ]
 
 log_files = []
+lock = filelock.FileLock("mender_lock_file")
+
 
 def docker_compose_cmd(arg_list, use_common_files=True):
     extra_files = pytest.config.getoption("--docker-compose-file")
@@ -40,13 +45,17 @@ def docker_compose_cmd(arg_list, use_common_files=True):
         for file in COMPOSE_FILES + extra_files:
             files_args += " -f %s" % file
 
-    subprocess.check_call("docker-compose -p %s %s %s" % (conftest.docker_compose_instance, files_args, arg_list), shell=True)
+    with lock:
+        cmd = "docker-compose -p %s %s %s" % (conftest.docker_compose_instance,
+                                              files_args,
+                                              arg_list)
+        subprocess.Popen(cmd, shell=True).wait()
 
 
 def stop_docker_compose():
     # take down all COMPOSE_FILES and the s3 specific files
+    docker_compose_cmd(" -f ../docker-compose.storage.s3.yml -f ../extra/travis-testing/s3.yml kill")
     docker_compose_cmd(" -f ../docker-compose.storage.s3.yml -f ../extra/travis-testing/s3.yml down -v")
-
     set_setup_type(None)
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -23,13 +23,13 @@ logging.basicConfig(level=logging.INFO)
 logging.getLogger("requests").setLevel(logging.CRITICAL)
 logging.getLogger("paramiko").setLevel(logging.CRITICAL)
 logging.getLogger("urllib3").setLevel(logging.CRITICAL)
-
 docker_compose_instance = "mender" + str(random.randint(0, 9999999))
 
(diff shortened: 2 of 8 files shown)


GregorioDiStefano:
@kacf

tests/common_docker.py
```diff
@@ -40,19 +40,26 @@ def docker_compose_cmd(arg_list, use_common_files=True):
...
 
 
 def start_docker_compose():
     inline_logs = pytest.config.getoption("--inline-logs")
+
+    while os.path.isfile(conftest.lock_file):
+        logging.debug("waiting for lock")
+        time.sleep(1)
+
+    os.mknod(conftest.lock_file)
```
kacf: Why `mknod`? Is it to achieve atomic file creation?

tests/common_docker.py
```diff
@@ -40,19 +40,26 @@ def docker_compose_cmd(arg_list, use_common_files=True):
...
     docker_compose_cmd("up -d")
+    os.remove(conftest.lock_file)
```
kacf: Several things:

* Don't do this here. Instead, move it into the `docker_compose_cmd` function, so that all commands are covered. Tearing it down is likely to be equally buggy.
* Wrap the docker-compose command in a `try`, and make sure that the `os.remove` happens inside the `finally` of that `try`.
* The whole thing has a race condition: If the file is created between your test and the file creation, the invocation will simply fail. The correct way to do this is to attempt the file creation immediately, and then test for AlreadyExists exception and do the sleep there. Wrap the whole thing in a loop which you only exit if the call is successful.

tests/conftest.py
```diff
@@ -86,6 +88,13 @@ def pytest_unconfigure(config):
     if not config.getoption("--no-teardown"):
         stop_docker_compose()
 
+def pytest_xdist_node_collection_finished(node, ids):
```
kacf: Is this some special function that gets called with xdist?

If so, it won't work without xdist, because the lockfile will be used even if xdist isn't present. Isn't there some pytest initialization function we can move this to? One that executes before xdist starts forking processes.

tests/common_docker.py
bboozzoo: You'd actually want to use context manager and do something like this:

```
from contextlib import contextmanager
import os
import time
import logging

@contextmanager
def lock_file(path):
    while True:
        try:
            fd = os.open(path, os.O_CREAT | os.O_EXCL)
        except FileExistsError as e:
            logging.info('waiting for lock..')
            time.sleep(1)
        else:
            break
    os.close(fd)
    try:
        yield
    finally:
        os.unlink(path)

if __name__ == '__main__':
    logging.basicConfig(level=logging.DEBUG,
                        format='%(process)d %(asctime)s - %(name)s - %(levelname)s - %(message)s')

    with lock_file('/tmp/foo'):
        logging.info('doing..')
        time.sleep(10)
        logging.info('done')
```

kacf: Ah yes, that's even better!

But if @GregorioDiStefano is right that xdist may unconditionally kill the subprocess that invokes this, then an actual "flock" might be the only safe alternative, because we will rely on the kernel to clean things up if the killing happens before the unlink is reached.

GregorioDiStefano:
ignore this PR right now, doing some testing still.

GregorioDiStefano:
@kacf Updated and review ready

GregorioDiStefano:
For some reason, I need to be using the latest version of Docker for this to work. 
Otherwise, pytest hangs.

tests/conftest.py
```diff
@@ -82,10 +84,16 @@ def pytest_exception_interact(node, call, report):
...
 
+
 def pytest_unconfigure(config):
     if not config.getoption("--no-teardown"):
         stop_docker_compose()
 
 
+def pytest_runtest_teardown(item, nextitem):
+    if nextitem is None:
+        stop_docker_compose()
```
kacf: Clever!

tests/run.sh
```diff
@@ -67,14 +67,19 @@ else
 fi
 
 XDIST_ARGS="-n auto"
+MAX_FAIL_ARG="--maxfail=1"
```
kacf: Don't remove execute permission on `run.sh`!

GregorioDiStefano: Fixed

GregorioDiStefano:
Added 2 new commits

The first one removes sleeping for 5 minutes if a device is already rebooted: this isn't need anymore, since the rest of the code will block until connectivity works

The second commit results in numerous failed tests since the "pending" state is very shorted lived, it's possible we never "see" that state.

PR #210, opened 2017-04-18
Grouping test

kacf:


kacf:
@GregorioDiStefano: Grouping test finally finished and passing! I will merge now since I've already rebased this several times, but I would appreciate a lookover when you get back!

PR #214, opened 2017-04-25
Parallelization fixes

GregorioDiStefano:


GregorioDiStefano:
@kacf Using a lock file when using the mender-artifact tool seems to fix the issue regarding "internal error" being returned from the deployments micro service. 

Also, `docker kill` is what seems to make pytest hang when using older versions of docker.

PR #223, opened 2017-05-08
added device decommissioning test

GregorioDiStefano:
@kjaskiewiczz @maciejmrowiec 

can you think of anything else that should be added here?

i'm aware that decommissioning a device after it's made a deployment, should include a decommissioned device, but was hoping that will be added to the acceptance tests, so I don't have to perform an actual deployment here.

### DIFF
--- a/tests/tests/test_decommission.py
+++ b/tests/tests/test_decommission.py
@@ -22,16 +22,17 @@ from common_update import common_update_procedure
 from MenderAPI import inv, adm, deploy, deviceauth
 from mendertesting import MenderTesting
 
-@pytest.mark.usefixtures("standard_setup_one_client")
+# ugly way of passing a the "true" parameter to "standard_setup_one_client"
+@pytest.mark.parametrize("standard_setup_one_client", ["force_new"], indirect=True)
 class TestDeviceDecommissioning(MenderTesting):
 
 
     @MenderTesting.fast
-    def test_device_decommissioning(self):
+    def test_device_decommissioning(self, standard_setup_one_client):
         """ Decommission a device successfully """
 
         if not env.host_string:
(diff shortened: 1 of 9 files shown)


GregorioDiStefano:
ping

kjaskiewiczz:
Looks good to me. It's simplest decommissioning scenario, but other scenarios (like device with different admission statuses or device with active or finished deployment) can be tested by acceptance tests, imo.

GregorioDiStefano:
thanks, and yes, integration tests require a real device running, which slows things down significantly, i too think it's best to add those tests to acceptance

tests/tests/test_decommission.py
```diff
@@ -0,0 +1,90 @@
...
+        if not env.host_string:
+            execute(self.test_device_decommissioning, standard_setup_one_client, hosts=get_mender_clients())
+            return
+
+        adm.check_expected_status("pending", len(get_mender_clients()))
+        adm_id = adm.get_devices()[0]["id"]
+        device_id = adm.get_devices()[0]["device_id"]
+
+        # decommission device before it's accepted results in 404
+        deviceauth.decommission(adm_id, expected_http_code=404)
```
bboozzoo: IMO the status should be 200 as long as there was a single request from the device. If it's not then that's a bug in deviceauth

GregorioDiStefano: @kjaskiewiczz can you comment on this?

bboozzoo: hmm, you're using a wrong ID here ;) should be `device_id` instead of `adm_id`

GregorioDiStefano: changing this to device_id locally, still results in 404..

PR #234, opened 2017-05-16
test all exposed connections are using tls

GregorioDiStefano:


GregorioDiStefano:
@kacf this is for MEN-951, making sure that mender only exposes secure ports.

tests/tests/test_security.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @pytest.mark.usefixtures("standard_setup_without_client")
+    def test_ssl_only(self):
+        """ make sure we are not exposing any non-ssl connections"""
+
+        # get all exposed ports from docker
+        exposed_hosts = subprocess.check_output("docker ps | grep %s | grep -o -E '0.0.0.0:[0-9]*'" % (conftest.docker_compose_instance), shell=True)
+        print exposed_hosts.split()
+
+        for host in exposed_hosts.split():
```
bboozzoo: Works as long as we're exposing http over SSL which may not be the case in the future. IMO a better way would be to check if the port is secured like this:

```
import ssl
import socket
import contextlib

for host in exposed_hosts.split():
    with contextlib.closing(ssl.wrap_socket(socket.socket())) as sock:
       with pytest.raises(ssl.SSLError):
           sock.connect()
```    
on top of this you can force specific versions of the protocol, ex. `ssl.wrap_socket(socket.socket(), ssl_version=ssl.PROTOCOL_TLS)` and so on (NOTE: ymmv with python 2.x)

tests/tests/test_security.py
```diff
@@ -0,0 +1,54 @@
...
+            logging.info("%s: connect to host with TLS" % host)
+            response = requests.get('https://' + host, verify=False)
+            assert response.status_code == 200 or response.status_code == 403, "%s unexpected response when connecting ssl: %s" % (host, response.text)
+
+        # make sure we either redirect or return bad request
+        for host in exposed_hosts.split():
+            logging.info("%s: connect to host without TLS" % host)
+            response = requests.get('http://' + host, allow_redirects=False)
+            if "Location" in response.headers:
+                assert response.headers["Location"].startswith("https://"), "% redirect was not to https!"
```
bboozzoo: I think we can assume that redirect will be set up for API gateway only. No other port is guaranteed to have a redirect.

tests/tests/test_security.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class TestSecurity(MenderTesting):
+
+    @pytest.mark.usefixtures("standard_setup_without_client")
+    def test_ssl_only(self):
+        """ make sure we are not exposing any non-ssl connections"""
+
+        # get all exposed ports from docker
+        exposed_hosts = subprocess.check_output("docker ps | grep %s | grep -o -E '0.0.0.0:[0-9]*'" % (conftest.docker_compose_instance), shell=True)
```
bboozzoo: playing with python:

```
# make dict: service name -> ports, grab lines, split on ';', grab service name, split on '_', pick middle element
In [59]: d = {srv[0].split('_')[1]: srv[1] for srv in [l.split(';') for l in subprocess.check_output('docker ps --format "{{.Names}};{{.Ports}}"', shell=True).decode().splitlines()]}

In [60]: d
Out[60]: 
{'mender-api-gateway': '0.0.0.0:443->443/tcp',
 'mender-conductor': '8080/tcp',
 'mender-deployments': '8080/tcp',
 'mender-device-adm': '8080/tcp',
 'mender-device-auth': '8080/tcp',
 'mender-dynomite': '8101-8102/tcp, 22122/tcp, 22222/tcp',
 'mender-elasticsearch': '9200/tcp, 9300/tcp',
 'mender-gui': '80/tcp',
 'mender-inventory': '8080/tcp',
 'mender-useradm': '8080/tcp',
 'minio': '9000/tcp',
 'mongo-common': '27017/tcp',
 'storage-proxy': '0.0.0.0:9000->9000/tcp'}
# grab public ports
In [62]: re.findall(r'0.0.0.0:[0-9]*', ' '.join(d.values()))
Out[62]: ['0.0.0.0:9000', '0.0.0.0:443']
```
Don't know if you can have docker print only public ports

bboozzoo:
@GregorioDiStefano will you be updating this PR or should we merge it as it is now?

PR #235, opened 2017-05-18
Better assertion error logging

GregorioDiStefano:


GregorioDiStefano:
@kacf hm, but if the s3cmd command fails/doesn't exist, no exception is thrown, ret != 0 and I'll print a warning to the screen, isn't that enough?

kacf:
This is uploading to a live cloud service. I really think we shouldn't even attempt that unless it's reasonable to assume you have the right credentials. Why don't you check the `UPLOAD_OUTPUT` variable? This is what's used for the old Jenkins build, and we can set it on the new one as well.

PR #236, opened 2017-05-18
Add --integration-versions-including argument to release tool.

kacf:
Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -3,7 +3,6 @@
 import argparse
 import copy
 import json
-import operator
 import os
 import re
 import shutil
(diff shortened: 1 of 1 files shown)


extra/release_tool.py
```diff
@@ -173,14 +175,21 @@ def docker_compose_files_list(dir):
...
+    data = {}
+    for json_str in json_list:
+        deepupdate(data, yaml.load(json_str))
+    return data
+
 def get_docker_compose_data(dir):
     """Return the Yaml as data from all the docker-compose YAML files."""
-    data = {}
+    json_list = []
     for file in docker_compose_files_list(dir):
```
GregorioDiStefano: 'file' is a built in, not sure if it's good idea to use it as a variable

kacf: The linter didn't complain, and I've been using this a lot. What is it?

extra/release_tool.py
```diff
@@ -955,14 +967,67 @@ def do_set_version_to(args):
...
+        candidates.append(line)
+
+    # Now look at each docker compose file in each branch, and figure out which
+    # ones contain the version of the service we are querying.
+    matches = []
+    for candidate in candidates:
+        yamls = []
+        files = execute_git(None, git_dir, ["ls-tree", "--name-only", candidate],
+                            capture=True).strip().split('\n')
+        for file in files:
```
GregorioDiStefano: same here

extra/release_tool.py
GregorioDiStefano: http://stackoverflow.com/a/2418007

kacf: Both fixed.

PR #239, opened 2017-05-24
fail with assert False in decommissioning test in order to get logs

GregorioDiStefano:
Changelog: none


GregorioDiStefano:
@kacf 

Failing with pytest.fail() doesn't upload logs to the s3 bucket, so I changed some relevant pytest.fail's to assert False

kacf:
Didn't know there was such a subtle difference between them.

PR #242, opened 2017-05-25
extras/testing/run-test-environment: preserve a predictable 'acceptance' service name

mchalski:
...by running 'up' instead of 'run'. the latter ignores service names by design.

Changelog: None

@bboozzoo @maciejmrowiec


mchalski:
also @GregorioDiStefano please take a look. the thing is we'll have to run a stub/fake of the `tenantadm` service to test multitenant scenarios.  the easiest way is to hack it together in python, and - to avoid container sprawl - plop it into the `testing` container (i.e. the `acceptance` compose service).

the issue is that the testing service now needs to get a predictable hostname, to which we can refer from other containers in the network. it happens for all service ootb, but not for the `acceptance` service. docker has a quirk where services executed via `docker-compose run` get ad-hoc service names and are not trivially reachable via their names. in this pr we work around this issue by doing `up` and collecting the exit status manually. `up` honours the service names as usual.

a funny side-effect is that we lose colored test output from pytest.

PR #244, opened 2017-05-26
release_tool: Standardize on naming of build parameters.

kacf:
Changelog: None


kacf:
This only affects the internal Jenkins.

PR #259, opened 2017-06-08
update backend testing 

GregorioDiStefano:


tests/tests/test_update.py
```diff
@@ -0,0 +1,191 @@
...
+    for volume in docker_volumes:
+        ret = subprocess.call(["docker", "volume", "create", "--name=%s" % volume])
+        assert ret == 0, "failed to create docker volumes"
+
+
+def setup_upgrade_source(upgrade_from):
+    # TODO: change this
+    if not os.path.exists("upgrade-from"):
+        ret = subprocess.call(["git", "clone", "https://github.com/GregorioDiStefano/integration", "-b", upgrade_from, "upgrade-from"])
+        assert ret == 0, "failed to git clone old source version"
```
kacf: > TODO: change this

Indeed. But will this need to stay until this is merged?

tests/tests/test_update.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # TODO: remove TEST_BRANCH: 1.0.x once merged
+    # start upgrade source and bring up it's production environment
+    ret = subprocess.call(["bash", "run.sh", "--get-requirements"],
+                          cwd="upgrade-from/tests",
+                          env={"TEST_BRANCH": "1.0.1"})
```
kacf: Any idea how we will determine this version number in the future? Will we simply put in the latest of each release we support or do you have something automatic in mind?

tests/tests/test_update.py
```diff
@@ -0,0 +1,191 @@
...
+
+def perform_upgrade():
+    ret = subprocess.call(["cp", "-r", "tests/upgrade-from/keys-generated", "."], cwd="..")
+    assert ret == 0, "faled to copy keys from original environment"
+
+    ret = subprocess.call(["cp", "extra/production-environment/production-testing-env.yml", "."], cwd="..")
+    assert ret == 0, "copying extra/production-envrionment folder failed."
+
+    ret = subprocess.call(["docker-compose", "-p", "testprod", "-f", "docker-compose.yml", "-f", "docker-compose.storage.minio.yml", "-f", "production-testing-env.yml", "up", "-d"], cwd="..")
+    assert ret == 0
```
kacf: Again, you can get rid of *a lot* or asserts here by using `check_call` instead of `call`.

tests/tests/test_update.py
```diff
@@ -0,0 +1,191 @@
...
+                         "-pollfreq=1",
+                         "-wait=1",
+                         "-failcount=%s" % str(fail_count)])
+    return p
+
+
+def provision_upgrade_server():
+    # deploy update to 10 devices
+    ret = subprocess.Popen(["python", "production_test_env.py", "--test-deployment"], cwd="upgrade-from/tests", stdout=subprocess.PIPE)
+    time.sleep(120)
```
kacf: Can you replace this with a `ssh_is_opened` call to the client? Or something similar to it. The client is always the last to finish and this has worked really well for our other tests. This gives two great benefits: You both finish faster if you get a fast droplet, and will in fact wait longer (but not fail) if you get a crappy droplet.

tests/run.sh
```diff
@@ -5,20 +5,21 @@ RUN_S3=""
...
+    local artifact_version=$(../extra/release_tool.py --version-of mender-artifact)
+
+    curl "https://d25phv8h0wbwru.cloudfront.net/${artifact_version}/tip/mender-artifact" \
          -o downloaded-tools/mender-artifact \
          -z downloaded-tools/mender-artifact
 
     chmod +x downloaded-tools/mender-artifact
 
-
     curl "https://s3.amazonaws.com/mender/temp_${TEST_BRANCH}/core-image-full-cmdline-vexpress-qemu.ext4" \
```
kacf: Maybe same for client, use:

```
local client_version=$(../extra/release_tool.py --version-of mender)
```

instead of `TEST_BRANCH`?

tests/run.sh
```diff
@@ -5,17 +5,25 @@ RUN_S3=""
...
-    export PATH=$PWD/downloaded-tools:$PATH
-
 
     curl "https://s3.amazonaws.com/mender/temp_${TEST_BRANCH}/core-image-full-cmdline-vexpress-qemu.ext4" \
          -o core-image-full-cmdline-vexpress-qemu.ext4 \
          -z core-image-full-cmdline-vexpress-qemu.ext4
+
+   curl "https://s3-eu-west-1.amazonaws.com/stress-client/release/mender-stress-test-client" \
+        -o downloaded-tools/mender-stress-test-client \
+        -z downloaded-tools/mender-stress-test-client
```
kacf: We should not use downloaded tools in release builds, so we need to add this to the non-`get_requirements` code block as well (the `-n "$BUILDDIR"` block). I suppose we can simply clone the repository and build it, or?

tests/tests/test_update.py
GregorioDiStefano: Yes

tests/tests/test_update.py
GregorioDiStefano: I can't, because I'm using the fake clients here, there is no ssh to connect to

tests/tests/test_update.py
GregorioDiStefano: MenderTesting.upgrade_from contains the branch that we are testing with, so I think I can just use that variable there (but, I'll have to merge this first: https://github.com/mendersoftware/integration/pull/248)

tests/run.sh
GregorioDiStefano: Okay, I can fix this.

tests/common_docker.py
```diff
@@ -47,8 +47,8 @@ def docker_compose_cmd(arg_list, use_common_files=True):
                                               arg_list)
 
         logging.info("running with: %s" % cmd)
-        subprocess.Popen(cmd, shell=True).wait()
-
+        ret = subprocess.Popen(cmd, shell=True).wait()
+        assert ret == 0, "docker-compose failed, command returned exit code: " + str(ret)
```
GregorioDiStefano: I want to use this, so I can clearly indicate what failed, CalledProcessError only returns the exit code

> Run command with arguments. Wait for command to complete. If the return code was zero then return, otherwise raise CalledProcessError. The CalledProcessError object will have the return code in the returncode attribute.

tests/production_test_env.py
```diff
@@ -0,0 +1,65 @@
...
+                                           "CERT_STORAGE_CN": "localhost"},
+                                      cwd="../")
+
+    # copy production environment yml file
+    if not os.path.exists("../production-testing-env.yml"):
+        subprocess.call(["cp", "extra/production-testing-env.yml", "."],
+                        cwd="../")
+
+    # start docker-compose
+    subprocess.call(["docker-compose",
```
GregorioDiStefano: Same as above

tests/run.sh
GregorioDiStefano: Fixed

tests/tests/test_update.py
GregorioDiStefano: Same as above

tests/tests/test_update.py
kacf: Ok. I don't like unconditional sleeps, but I don't know enough about the setup here to suggest anything better, so I guess it stays.

PR #261, opened 2017-06-12
Release tool to detect mender

GregorioDiStefano:
@kacf 

I'm using `env` in the release_tool.py shebang, since python3 might be a different location then /usr/bin/python3

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -1,25 +1,31 @@
 #!/bin/bash
 set -x -e
 RUN_S3=""
-MENDER_BRANCH=$(cd .. && python extra/release_tool.py --version-of mender)
+MENDER_BRANCH=$(../extra/release_tool.py --version-of mender)
 
 if [[ $? -ne 0 ]]; then
     echo "Failed to determine mender version using release_tool.py"
     exit 1
 fi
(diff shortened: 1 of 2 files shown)


tests/run.sh
```diff
@@ -1,30 +1,31 @@
...
+    exit 1
+fi
+
+echo "Detected Mender branch: $MENDER_BRANCH"
 
 function get_requirements() {
     # Download what we need.
     mkdir -p downloaded-tools
-    curl "https://d25phv8h0wbwru.cloudfront.net/${TEST_BRANCH}/tip/mender-artifact" \
+    curl "https://d25phv8h0wbwru.cloudfront.net/${MENDER_BRANCH}/tip/mender-artifact" \
```
kacf: Careful, this is a different version, you need to get `--version-of mender-artifact` for this.

tests/run.sh
```diff
@@ -1,30 +1,31 @@
 #!/bin/bash
 set -x -e
 RUN_S3=""
+MENDER_BRANCH=$(cd .. && python extra/release_tool.py --version-of mender)
```
kacf: You should not need to call python directly anymore after your other fix.

It should also not be necessary to `cd`, since the script will detect its own location and work from there.

GregorioDiStefano:
Thanks, fixed.

PR #262, opened 2017-06-13
Test aborting deployment during downloading state instead of installing.

pasinskim:
After introducing state scripts some states have been changed. Among
the other changes, installing at the moment is just flip-flopping
partitions, which is too fast operation to be able to get the status
and abort while still being in installing state.
Now, downloading is the state where downloading and storing new image
is happening, so testing aborting during downloading state is possible.

Changelog: Title

Signed-off-by: Marcin Pasinski <marcin.pasinski@mender.io>


pasinskim:
This was merged together with integration tests for state scripts functionality.

PR #287, opened 2017-06-27
compose: set ALLOWED HOSTS

mchalski:
the user will have to substitute this value with the DNS name he allocates
for the gateway.

the gateway's docker entrypoint will pick up this env var and modify the Host
whitelist in nginx.conf.

related to host header injection hardening.

Issues: MEN-1262

Changelog: None


GregorioDiStefano:
Force merging, only failed because the branch isn't rebased.

PR #288, opened 2017-06-28
testing: when using signed-client yml, use client container with :pr tag

GregorioDiStefano:
Changelog: none

### DIFF
--- a/extra/signed-artifact-client-testing/docker-compose.signed-client.yml
+++ b/extra/signed-artifact-client-testing/docker-compose.signed-client.yml
@@ -5,7 +5,6 @@ services:
     # mender-client
     #
     mender-client:
-        image: mendersoftware/mender-client-qemu:master
         networks:
             - mender
         volumes:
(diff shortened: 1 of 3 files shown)


extra/signed-artifact-client-testing/docker-compose.signed-client.yml
```diff
@@ -5,7 +5,7 @@ services:
     # mender-client
     #
     mender-client:
-        image: mendersoftware/mender-client-qemu:master
+        image: mendersoftware/mender-client-qemu:pr
```
kacf: This is added on top of the other docker compose files, right? In that case `image` should not be specified at all, you should just rely on the existing one.

GregorioDiStefano: didn't know that, thanks.

kacf:
@GregorioDiStefano: Doesn't the token stuff belong in #289?

GregorioDiStefano:
Ooops, fixed now :) Just included a commit by accident.

kacf:
This has been fixed already.

PR #292, opened 2017-06-29
tests/run.sh: whitelist all IPs/DNS names in the gateway for tests

mchalski:
tests access the gateway via IP, not a well-defined alias like docker.mender.io.
now that we have a host whitelist (server_name: ...) in the gateway, we have to
set it - so set it to a catchall regex because we don't know anything in advance.
otherwise pretty much everything fails with a nasty connection termination error.

btw '*' doesn't work here, yaml doesn't allow it (nginx doesn't allow it either for that matter). 

Issues: MEN-1262

Changelog: None

this is a followup to a previous PR to `integration` that we force merged to make https://github.com/mendersoftware/mender-api-gateway-docker/pull/65 green (but it was not enough obviously). @GregorioDiStefano can we force a merge here too if need be? (EDIT - the build is green so no need to force, but please have a look regardless).


mchalski:
@bboozzoo tried, failed, don't remember if it was nginx complaining and killing the container, or requests not going through

PR #293, opened 2017-07-01
[WIP] State scripts integration tests

pasinskim:
Do not merge; testing only


tests/tests/test_state_scripts.py
```diff
@@ -673,59 +652,59 @@ class TestStateScripts(MenderTesting):
...
-                # "ArtifactRollbackReboot_Enter_99",
-                # "ArtifactRollbackReboot_Leave_01",
-                # "ArtifactRollbackReboot_Leave_99",
-                # "ArtifactFailure_Enter_22",
-                # "ArtifactFailure_Enter_33",
-                # "ArtifactFailure_Leave_44",
-                # "ArtifactFailure_Leave_55",
+                # since version is corrupted from now on, no more scripts
+                # will be executed, but rollback will be performed
+                "Idle_Enter_08_testing",
```
kacf: I realize now that I have to change this test. You're doing exactly the right thing here, but the intention of the test was to test what happens if the version is wrong for the new Mender, but then right again when you roll back (this would be in the case where you try to update to a new version of Mender that is incompatible with the old one). But I need to improve the test to catch this.

pasinskim:
@kacf cool! Thanks for checking.

kacf:
Outdated.

PR #309, opened 2017-08-03
Authenticate with new Jenkins

GregorioDiStefano:


GregorioDiStefano:
@kacf Authenticate with new Jenkins

extra/release_tool.py
```diff
@@ -1200,6 +1197,13 @@ def main():
                         + "pull requests triggering a failure)")
     args = parser.parse_args()
 
+    if not os.getenv("JENKINS_USER") or not os.getenv("JENKINS_PASSWORD"):
+        logging.warn("WARNING: JENKINS_USER and JENKINS_PASSWORD env. variables not set")
+
+    global JENKINS_USER, JENKINS_PASSWORD
+    JENKINS_USER = os.getenv("JENKINS_USER")
+    JENKINS_PASSWORD = os.getenv("JENKINS_PASSWORD")
+
```
kacf: I think this section can be moved into `do_release`, since it won't be used in any other section, and the tool is used a lot outside of that as well.

kacf: It gets rid of the warning when you are not using the release functionality.

kacf:
Never mind, I can fix this myself. Thanks a lot!

kacf:
Did you test it? I get:

```
/usr/lib/python3/dist-packages/urllib3/connectionpool.py:794: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
Request returned: 500: Server Error
```

PR #315, opened 2017-08-03
tests: remove orphaned containers when killing test-production enviro…

GregorioDiStefano:
…nment

Changelog: none

### DIFF
--- a/tests/production_test_env.py
+++ b/tests/production_test_env.py
@@ -105,4 +105,4 @@ if args.deploy:
     print("devices=%d" % len(devices))
 
 if args.kill:
-    subprocess.call(["docker-compose", "-p", "testprod", "down", "-v"])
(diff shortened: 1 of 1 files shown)


tests/production_test_env.py
```diff
@@ -105,4 +105,4 @@ def setup_docker_volumes():
     print("devices=%d" % len(devices))
 
 if args.kill:
-    subprocess.call(["docker-compose", "-p", "testprod", "down", "-v"])
+    subprocess.call(["docker-compose", "-p", "testprod", "down", "-v", "--remove-orphans"])
```
kacf: Won't this kill containers for other unrelated tests? It says "Remove containers for services not defined in the Compose file". It's not completely clear from the documentation whether this applies to containers in a different namespace or not, so this should be tested.

GregorioDiStefano: It won't, this is limited to the specific project

PR #320, opened 2017-08-07
add warning regarding https://localhost no longer loading webui

GregorioDiStefano:
Changelog: none

### DIFF
--- a/up
+++ b/up
@@ -49,6 +49,8 @@ setuphosts() {
 
 rundocker() {
     # sudo -E docker-compose scale mender-client=3
+    echo "The web interface will be accessible locally via https://docker.mender.io"
+    sleep 2
     sudo -E ./demo up
 }
 


GregorioDiStefano:
@mchalski @estenberg

mchalski:
I understand your frustration with the `localhost` issue, but mind you - me, personally, have nothing against adding it to `ALLOWED_HOSTS`, and I'd rather do that. yes it was me who removed it in the first place, but only to be 100% in line with our docs; there was also some vague security related argument for it, which I don't remember and it's irrelevant (demo is all about being not secure). so let's be sane and enable `localhost` again.

is anyone objecting? @mendersoftware/rndity ?

bboozzoo:
Why even bother adding [s3.]docker.mender.io to `/etc/hosts` then?

estenberg:
@maciejmrowiec your input would be helpful as well!

GregorioDiStefano:
@bboozzoo we originally needed s3.docker.mender.io added due to some requirement I can't remember right now. I believe it was a hack to get fakes3 working. Since we stared using minio, I'm not sure this is required anymore.

I don't know why we add docker.mender.io to the hosts file.

maciejmrowiec:
For demo setup we could just add that localhost to the whitelist on the gateway. There was a security argument to have it strict, but after all this is only demo environment. 

It would be nice if we could get rid of `/etc/hosts` modifications. Question is can we? - last commit is about fakes3, and I also suspect that in the past UI (aka local browser) had to talk to S3 directly and it was connected to how s3 links were constructed. There is still API that allows us to get that direct link (GET artifacts/:id/download) so returned link probably could stop working after removing that mapping however this is not used API today so not something to be strongly concerned with. No sure if switch to minio changed something or not, maybe @kjaskiewiczz would have some idea.

Not sure if this removal would impact connecting external devices to demo setup or not.

estenberg:
@maciejmrowiec by "add that localhost to the whitelist on the gateway", do you mean what is done in https://github.com/mendersoftware/integration/pull/322 ?

estenberg:
I am closing this as we are 1) fixing the root issue in https://github.com/mendersoftware/integration/pull/322 and 2) @GregorioDiStefano is investigating removal of the /etc/hosts modifications for 1.2.0 final.

PR #328, opened 2017-08-11
test: test docker container pulling

GregorioDiStefano:

### DIFF
--- a/tests/common_setup.py
+++ b/tests/common_setup.py
@@ -139,20 +139,21 @@ def standard_setup_with_short_lived_token():
 
 @pytest.fixture(scope="function")
 def running_custom_production_setup(request):
-    conftest.production_setup_lock.acquire()
+    with conftest.production_setup_lock.acquire():
+        # since we are starting a manual instance of the backend,
+        # let the script know the instance is called "testprod"
+        # so that is cleaned up correctly on test failure/error
 
-    # since we are starting a manual instance of the backend,
-    # let the script know the instance is called "testprod"
-    # so that is cleaned up correctly on test failure/error
+        def fin():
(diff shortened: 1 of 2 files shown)


tests/common_setup.py
```diff
@@ -139,20 +139,21 @@ def standard_setup_with_short_lived_token():
...
-    # so that is cleaned up correctly on test failure/error
+        def fin():
+            stop_docker_compose()
+            conftest.production_setup_lock.release()
 
-    def fin():
-        conftest.production_setup_lock.release()
-        stop_docker_compose()
+        # this additional lock is required
+        conftest.production_setup_lock.acquire()
```
kacf: This looks wrong, you're acquiring the lock twice here, once using `with`, and once here. When do you want to hold the lock, during the whole production environment being up, or only during initialization and teardown?

GregorioDiStefano:
Filelock python module uses a lock counter. There is a lock used when starting the backend (which is then released) and an additional lock used until the finalizer releases it.

kacf:
Ok... But what is that supposed to achieve? AFAICS you could get the exact same behavior, with less complexity, by getting rid of the `with` and acquiring the lock immediately after entering the `running_custom_production_setup`. This would also be easier to understand when reading it.

GregorioDiStefano:
Yes, your suggestion should work. Can you amend my PR and merge this?

kacf:
Sure, after I fix the tenant test breakage.

kacf:
> Yes, your suggestion should work. Can you amend my PR and merge this?

Actually, looking closer, this is exactly what the old version does, and also the correct thing to do. So no action required here. :-)

PR #331, opened 2017-08-15
Revert "tests: pull all containers in parallel before test run"

kacf:
We should not pull containers in the test run. This breaks local
testing and is dangerous in a Jenkins run because you might be pulling
containers that overwrite what you're trying to test.

This reverts commit 003c8bf29bb1030e2796d0a9ccfbc73a6e1c15a0.


kacf:
@GregorioDiStefano: FYI

PR #336, opened 2017-08-17
tests: Use individual execute() calls to preserve exception info.

kacf:
Changelog: None

### DIFF
--- a/tests/tests/test_state_scripts.py
+++ b/tests/tests/test_state_scripts.py
@@ -651,12 +651,7 @@ class TestStateScripts(MenderTesting):
         """Test that state scripts are executed in right order, and that errors
         are treated like they should."""
 
-        if not env.host_string:
-            execute(self.test_state_scripts, test_set,
-                    hosts=get_mender_clients())
-            return
-
-        client = env.host_string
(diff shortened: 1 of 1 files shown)


tests/tests/test_state_scripts.py
```diff
@@ -709,16 +704,17 @@ def test_state_scripts(self, test_set):
             # quite slow.
             subprocess.check_call(["tar", "czf", "../rootfs-scripts.tar.gz", "."], cwd=rootfs_script_dir)
             # Stop client first to avoid race conditions.
-            run("systemctl stop mender")
+            execute(run, "systemctl stop mender", hosts=[client])
```
pasinskim: Not sure what the difference is here, but if this is needed then looks OK.

kacf: I'm not exactly sure why, but if you recurse into `execute()`, sometimes, but not always, you lose all the exception info if the error happens in there. By calling it individually where we need it, we keep this info in all places we are not running `execute()`, which is most of the code.

kacf:
I think the tests are working now, so we don't need this.

PR #344, opened 2017-09-05
add script to add tenant token / hosted server to pre-existing .mende…

GregorioDiStefano:
…r / .sdimg file


GregorioDiStefano:
Not sure who would like to take a look, but @kacf @bboozzoo @maciejmrowiec ?

I'm worried that parsing the start sectors using `parted` might fail if the output is not consistent on different versions.

extra/add_tenant_token.sh
```diff
@@ -0,0 +1,125 @@
...
+  local start_sectors=$(sudo parted "$filename" "unit s print" | grep -i '^[[:blank:]]\{1,\}[0-9]' | awk '{ print $2}' | tr -d 's')
+  read -a start_sectors_array <<< $start_sectors
+
+  local tempfile=$(mktemp)
+  local jq_argument=$(printf ".TenantToken = \"%s\" | .ServerURL = \"%s\"" "$tt" "$HOSTED_MENDER_URL")
+
+  # mount every partition, and copy over the mender.conf file
+  for s in "${start_sectors_array[@]}"
+  do
+    sudo mount -o loop,offset=$((s * 512)) "$filename" "$mount_dir"
```
kacf: Why not `debugfs` here as well? I assume this will run on a SaaS server, and I really don't think sudo is a good idea there.

extra/add_tenant_token.sh
```diff
@@ -0,0 +1,125 @@
...
+  local tt=$2
+
+  if [ "$EUID" -ne 0 ]; then
+    echo "Please run as root"; exit 1
+  fi
+
+  local mount_dir=$(mktemp -d)
+
+  # ugly way to determine start sectors of sdimg
+  local start_sectors=$(sudo parted "$filename" "unit s print" | grep -i '^[[:blank:]]\{1,\}[0-9]' | awk '{ print $2}' | tr -d 's')
```
kacf: `sudo` should not be needed for a plain file, or?

extra/add_tenant_token.sh
kacf: Oh right, you will have to copy out the file in that case. Bah... Well, it's up to you. This is a stop gap solution anyway, since Marcin is working on proper udisks support.

bboozzoo:
Discussed this on Slack. IMO the best way forward would be to import https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-qemu/scripts/docker/setup-mender-configuration.py which does exactly this and more.

For handling `.mender` artifact the `mender-artifact` tool should probably be updated to allow extracting and repacking images. At least you'd know that the artifact made by mender-artifact is properly structured, while poking manually with tar is dubious (although deb is a tar, you wouldn't repack a deb manually, would you?).

kacf:
I agree that this duplication is not good, but @pasinskim is working on [a more permanent solution](https://tracker.mender.io/browse/MEN-1213) for this which uses Udisks over DBUS, and which will trump both this script and `setup-mender-configuration.py`. So I suppose we can live with a bit hacky solution for now.

bboozzoo:
I'm slightly confused, how do you plan on replacing `setup-mender-configuration.py` in docker container? there's no udisks, there's no dbus daemon, and you will not be able to mount anything.

kacf:
Right, this will be *inside* a Docker container in many cases. Good point, I did not think about that.

kacf:
I talked it over with Marcin and Greg: Basically we don't really have a choice, inside containers we have to keep using debugfs for the time being, but our user facing tool will be mender-artifact which uses Udisks. It's not an ideal solution, but at least the debugfs part of it will be mostly contained to our own workflow and not so much users'.

A couple of options exist if we want to fix the container case, all with some caveats:

* Do the necessary modifications outside the container instead of inside.
* "Inception" option: Launch QEMU inside the container, and use the mount command inside the container to modify the image over scp/rsync, then tear down and restart QEMU using modified image.

No, I didn't say any of them were good, just that they were possible. :-)

In any case, the script in this PR will be in use for a few weeks at most, before the Udisks version is out.

bboozzoo:
Are you sure that depending on udisks is the right thing to do? IIRC it's not really a part of a default server installation of Debian/Ubuntu (Fedora maybe?). It may be fine for a desktop installation but If you plan on using this in SaaS it may be a bit of a stretch. Not sure about stability of udisks if there happens to be a significant number of requests. On top of this you will need to load via loopback and mount an image that comes from untrusted source. You'll obviously need elevated privileges to create loopback devices. Mounting may modify the superblock of the image. Read-only file systems (as will be mounted read only) and raw flash file systems (need additional modules, not even sure udisks can handle this) cannot be easily supported, you'll still need additional code to handle this. Finally, testing this will be a mess. 

On the other hand, using e2fsprogs and mtools is nicely confined to the running process, no mounting necessary, no loopbacks, no daemons running as root. Testing is trivial (and we kind of do it already in meta-mender). FWIW wic also uses debugfs/mtools to modify the image.

kacf:
We want to go further than just being able to insert one file here and there, we want to be able to truly inspect and edit the image, and for this mounting is unavoidable. Udisks is just the lesser evil versus sudo. Loopback works just fine with udisks. The only thing I think doesn't work is automatic partition detection (what `kpartx` does), this requires sudo AFAIK. But we can get around it by using loopback offsets manually.

GregorioDiStefano:
https://github.com/mendersoftware/saas-tools/pull/1

PR #345, opened 2017-09-05
compose/main: run services with --automigrate in the base layer

mchalski:
this is the best place for it, not everyone uses the demo layer (e.g. tests).


GregorioDiStefano:
@mchalski I think we drop this PR, the issue I was seeing isn't related to this.

bboozzoo:
do we need these changes then?

GregorioDiStefano:
nope, closing.

PR #348, opened 2017-09-07
add new SaaS/multi-tenantcy related integration tests

GregorioDiStefano:

### DIFF
--- a/tests/common_setup.py
+++ b/tests/common_setup.py
@@ -17,6 +17,7 @@ from MenderAPI import auth, adm, reset_mender_api
 from common import *
 from common_docker import *
 import conftest
+import time
 
 @pytest.fixture(scope="function")
 def standard_setup_one_client(request):
@@ -168,6 +169,9 @@ def multitenancy_setup_without_client(request):
                         -f ../docker-compose.tenant.yml \
                         -f %s up -d" % (conftest.mt_docker_compose_file),
                         use_common_files=False)
+
+    # wait a bit for the backend to start
+    time.sleep(30)
 
     def fin():
         stop_docker_compose()
--- a/tests/common_setup.py
+++ b/tests/common_setup.py
@@ -19,6 +19,18 @@ from common_docker import *
 import conftest
 import time
 
+def wait_for_containers(expected_containers, defined_in):
+    out = subprocess.check_output("docker-compose -p %s %s ps -q" % (conftest.docker_compose_instance, "-f " + " -f ".join(defined_in)), shell=True)
+
+    for _ in range(60 * 5):
+        if len(out.split()) == expected_containers:
+            time.sleep(60)
+            return
+        else:
+            time.sleep(1)
+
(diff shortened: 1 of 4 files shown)


GregorioDiStefano:
@kacf @bboozzoo 

please take a look at these small changes and new test

tests/common_setup.py
```diff
@@ -169,6 +170,9 @@ def multitenancy_setup_without_client(request):
                         -f %s up -d" % (conftest.mt_docker_compose_file),
                         use_common_files=False)
 
+    # wait a bit for the backend to start
+    time.sleep(30)
```
kacf: I must reiterate. Don't use sleeps. This *will* bite us later, and is the reason our tests don't work on crappy droplets. I strongly urge you to *detect* that the server is up instead of relying on a sleep. It could have an eventual timeout of course, but this should be *much* longer.

tests/tests/test_multi_tenancy.py
```diff
@@ -89,19 +89,17 @@ def set_correct_tenant_token(token):
     @pytest.mark.usefixtures("multitenancy_setup_without_client")
     def test_artifacts_exclusive_to_user(self):
         users = [
-            {"email": "greg1@greg1.com", "password": "hunter2hunter2", "username": "greg1"},
-            {"email": "greg2@greg2.com", "password": "hunter2hunter2", "username": "greg2"},
-            {"email": "greg3@greg2.com", "password": "hunter2hunter2", "username": "greg3"}
+            {"email": "foo1@foo1.com", "password": "hunter2"*2, "username": "foo1"},
+            {"email": "bar2@bar2.com", "password": "hunter2"*2, "username": "bar2"},
```
kacf: The `*2` syntax is just weird... Isn't this harder to read than what was already there?

tests/common_setup.py
GregorioDiStefano: oops, yes, good point. Fixed

bboozzoo: I see the fix and I don't think it's complete. Even though the container is up, the service may still be starting up. On the other hand, I don't see us implementing proper synchronization anyway.

GregorioDiStefano: so what do you suggest I do here?

bboozzoo: Ideally we should have some health check endpoint in each service. But here, I don't think we can do anything reasonable here without the changes in other components of the system.

kacf: Maybe just try an endpoint that is supposed to give 401 Forbidden or something? Or some other expected failure.

GregorioDiStefano: is there anything that can be done here?

kacf: Just make a request to some random endpoint and make sure you get the expected response? (even if that response is an error)

GregorioDiStefano: @kacf I checked with @bboozzoo and there is no way of doing this. 

What we currently have is wait for all the containers to be up, and after 30 seconds run tests. I don't know what else we can do here, without an endpoint verifying if containers are indeed up and running, and connected to their db

PR #360, opened 2017-09-26
tests: create user only in useradm

mchalski:
the user will be propagated to tenantadm via useradm,
so no need to call create-user in tenantadm. in fact this
will cause 'email already exists'.

needed for: https://github.com/mendersoftware/useradm/pull/94

changelog: none

### DIFF
--- a/tests/MenderAPI/authentication.py
+++ b/tests/MenderAPI/authentication.py
@@ -64,7 +64,6 @@ class Authentication:
                 tenant_id = self._create_tenant(self.username)
                 tenant_id = tenant_id.strip()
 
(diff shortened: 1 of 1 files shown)


tests/MenderAPI/authentication.py
```diff
@@ -108,12 +107,6 @@ def _create_tenant(self, username):
                                                                                              username)
         return docker_compose_cmd(cmd)
 
-    def _create_tenant_user(self, email, tenant_id):
```
bboozzoo: Is this called in one place only?

mchalski: to my knowledge, yes, grepped through everything

PR #363, opened 2017-09-27
tests: Open client SSH connection more gently, not knocking every second.

kacf:
We have some prior evidence that if you knock too often, the key
generation gets respawned too often to ever actually finish, and this
gets worse if resources are scarce.

Changelog: None

### DIFF
--- a/tests/common_docker.py
+++ b/tests/common_docker.py
@@ -139,8 +139,9 @@ def ssh_is_opened():
 
 
 @parallel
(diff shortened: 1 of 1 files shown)


GregorioDiStefano:
sincee this commit was introduced, integration tests take around 30-45 minutes longer than previously, and we are still seeing clients failing to establish an ssh connection

kacf:
Maybe reduce sleep time to 10 instead?

GregorioDiStefano:
yes

PR #367, opened 2017-09-28
tests: create larger test file when testing 'out of space' artifacts

GregorioDiStefano:
Changelog: none


GregorioDiStefano:
the test was creating a 200mb image, but the inactive partition is 208mb, looks like the verification of no reboot taking place _sometimes_ works as intended, which is why this test sporadically fails

kacf:
:+1:

PR #370, opened 2017-09-28
release_tool: Fix bug causing integration version references to be wr…

kacf:
…ong.

The symptom was that the docker-compose files diff would be missing a
few lines, which was subtle and hard to spot. Make sure we *always* do
the update.

Changelog: None


kacf:
Not used in testing, merging immediately.

PR #371, opened 2017-09-28
tests: lower connection timeout

GregorioDiStefano:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -105,7 +105,7 @@ def pytest_configure(config):
 
     env.connection_attempts = 50
     env.eagerly_disconnect = True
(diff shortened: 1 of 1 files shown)


GregorioDiStefano:
just testing how much faster tests are when ssh timeout is lower

kacf:
Yes, this is useful. They should not ever time out in theory, and I think that if you hit the key generation window, a reconnect might sometimes be required. Better to not have to wait 3 minutes in that case.

GregorioDiStefano:
precisely what i was thinking

GregorioDiStefano:
merging this, as its only failing due to other spurious fails

PR #387, opened 2017-10-09
if possible, pull docker-compose in parallel on first run

GregorioDiStefano:
Changelog: none

### DIFF
--- a/demo
+++ b/demo
@@ -1,10 +1,32 @@
 #!/bin/bash
 
-
 CLIENT="-f docker-compose.client.yml"
 if [ -n "${NO_CLIENT}" ]; then
     unset CLIENT
     echo "-- disabling client containers"
+fi
+
+if [ "$#" -eq 0 ]; then
+    echo "Arguments passed to this command are passed directly to docker-compose, if you want to run the demo, run:"
+    echo ""
+    echo "'./demo up'"
+    exit 1
+fi
+
+# speed up first start by pulling containers in parallel
+docker images | grep -q 'mendersoftware/deployments'
+
+if [[ "$?" -eq 1 ]]; then
+    docker_compose_output=$(docker-compose pull -h)
+    echo "$docker_compose_output" | grep -q -- '--parallel'
+
+    if [[ "$?" -eq 0 ]]; then
+        docker-compose pull --parallel
+    fi
+fi
+
+if [[ "$1" == "up" ]]; then
+    echo "Starting Mender demo environment.."
 fi
 
 exec docker-compose \


demo
```diff
@@ -1,12 +1,32 @@
...
 
+if [ "$#" -eq 0 ]; then
+    echo "Arguments passed to this command are passed directly to docker-compose, if you want to run the demo, run:"
+    echo ""
+    echo "'./demo up'"
+    exit 1
+fi
+
+# speed up first start by pulling containers in parallel
+if [ ! -e ".first-run" ]; then
```
kacf: I think a better check is whether the images already exist locally. It should be enough to check one of them.

GregorioDiStefano: Ok, updated!

demo
```diff
@@ -1,12 +1,30 @@
...
+if [[ "$(docker images | grep -q 'mendersoftware/deployments')" -eq 0 ]]; then
+    docker_compose_output=$(docker-compose pull -h)
+    check=$( echo "$docker_compose_output" | grep -q -- '--parallel')
+
+    if [[ "$check" -eq 0 ]]; then
+        docker-compose pull --parallel
+    fi
+fi
+
+echo "Starting Mender demo environment.."
```
kacf: Oh and this is wrong if you are calling `demo` with `down` or `pull`. I would just drop this, not really necessary, or?

demo
```diff
@@ -1,12 +1,30 @@
...
 
+if [ "$#" -eq 0 ]; then
+    echo "Arguments passed to this command are passed directly to docker-compose, if you want to run the demo, run:"
+    echo ""
+    echo "'./demo up'"
+    exit 1
+fi
+
+# speed up first start by pulling containers in parallel
+if [[ "$(docker images | grep -q 'mendersoftware/deployments')" -eq 0 ]]; then
```
kacf: `-eq 0` is not the right check, because `$()` will return empty string, not `0`. The correct thing is to drop the `[[` construct and call the command directly in the `if`.

demo
```diff
@@ -1,12 +1,30 @@
...
+    docker_compose_output=$(docker-compose pull -h)
+    check=$( echo "$docker_compose_output" | grep -q -- '--parallel')
+
+    if [[ "$check" -eq 0 ]]; then
```
kacf: Same here, you need to call the command in the `if` without the `$()` and the `[[ ]]`.

demo
GregorioDiStefano: Oops, yes, my bad.

PR #390, opened 2017-10-25
Fix Missing restart policy for some containers in docker-compose setup.

PiotrPrzybylak:
Issues: MEN-1556

Changelog: None

Signed-off-by: Piotr Przybylak piotr.przybylak@rndity.com


kacf:
This should have had `Changelog: Title`. It fixes a publicly facing bug, and it was reported on the mailing list.

PR #391, opened 2017-10-31
[DON'T MERGE] POC of conductor tasks/workers

mchalski:
This is a research POC for https://tracker.mender.io/browse/MEN-1571.
The short conclusion:
- the research was successful
-  task workers work as expected
- the python client works flawlessly under python3
- all workers can be placed in a single, generic docker container (see Gotchas)

How it works
- `create_organization` workflow fakes our main workflow (without service interactions)
- 2 tasks are defined - `prepare_welcome_email` and `send_email`, and a worker for each
- each worker needs a separate process/main(), but both live in a single container
- task 1 passes input params to 2 via the `outputs` field/mechanism

also, some compose extras and shell helpers are introduced to make testing easier

How to run
- do the usual `./demo up`, wait ~20 secs (hardcoded conductor sleep - see Gotchas)
- go to `localhost: 5000` to inspect workflow/task defs
- before launching the workflow, it's useful to attach to just the worker container: `./demo logs -f mender-conductor-workers ` (dynomite spams stdout completely...)
- launch the workflow via the helper script, e.g. `./conductor/server/workflows/create-organization tenant1 foo@tenant1.com somepass123`
- worker container's logs should indicate that the tasks were started/completed
- see `localhost: 5000` again to view the completed workflow/tasks

Gotchas
1. Process-per-worker 
   * can't run >1 worker in a single process
   * each ConductorWorker enters an infinite polling loop via `start()`
   * btw task distribution happens via http polling (not dynomite/redis) - ConductorWorker uses the dedicated `poll` api
   * for now it's worked around by having the entrypoint start one worker in the background, and another in the foreground - discuss if ok (if the background worker dies - will docker know/restart the container? should we user supervisord?)

2. Conductor startup time
   * conductor takes an indeterminate amount of time to start (10-20s), even though the container is started
   * in this time, workers might fail (max retries)
   * worked around by waiting 20s before starting each worker
   * to discuss: is there a saner way?

3. Various
   * workers *must* return well-formed responses containing *all* fields (see code) - even though they don't have anything to say; otherwise it's a silent failure and stuck workflow

### DIFF
--- a/conductor-workers/Dockerfile
+++ b/conductor-workers/Dockerfile
@@ -1,0 +1,18 @@
+FROM ubuntu:16.04
+
+RUN apt-get -y update && apt-get -y install git python3 python3-setuptools curl dnsutils
+
+RUN git clone https://github.com/Netflix/conductor.git
+
+RUN cd conductor/client/python && \
+       python3 setup.py install
+
+RUN mkdir -p /usr/bin/mender/prepare_welcome_email && \
+    mkdir -p /usr/bin/mender/send_email
+
+COPY prepare_welcome_email/* /usr/bin/mender/prepare_welcome_email/
+COPY send_email/* /usr/bin/mender/send_email/
+
+COPY entrypoint.sh /usr/bin/mender
+
+ENTRYPOINT ["/usr/bin/mender/entrypoint.sh"]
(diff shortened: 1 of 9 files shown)


kacf:
Outdated.

PR #399, opened 2017-11-08
MEN-1604: add conductor workers cherrypick to 1.3.x

mchalski:


kacf:
Dammit, there is an error due to a missing client image. I think I will have to start the build manually, forcing a client build. I think it's because this is the first 1.3.x build, so the client image is not up yet.

PR #400, opened 2017-11-15
Upgrade Conductor to 1.8.1

maciejmrowiec:
Main changes:
* Correct logging setup to stdout in docker image
* ElasticCache / Redis support
* Image size reduction comming from change of base images of docker to apline linux.
* Minor bug fixes and improvements.
* JSON_TRANSFORM system tasks.

Full change list: https://github.com/Netflix/conductor/compare/v1.7.6...v1.8.1

Changelog: Title

### DIFF
--- a/conductor/server/entrypoint/conductor-load
+++ b/conductor/server/entrypoint/conductor-load
@@ -1,6 +1,6 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
-import httplib
+import http.client
 import json
 import argparse
 import logging
@@ -23,7 +23,7 @@ def request(addr, method, url, body=None, headers=None):
     if not headers:
         headers = {}
 
-    with closing(httplib.HTTPConnection(addr)) as conn:
+    with closing(http.client.HTTPConnection(addr)) as conn:
         conn.request(method, url, body=body, headers=headers)
         resp = conn.getresponse()
         return (resp.status, resp.read())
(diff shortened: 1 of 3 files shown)


maciejmrowiec:
@mendersoftware/rndity can you guys think about any reason not to go for it?

Main things are logging in container is piped to stdout finally, 1.7.7 pipes it to local file. Dramatic size reduction of image which speeds up integration tests (542MB -> 84MB). Support for AWS hosted redis/elasticsearch.

maciejmrowiec:
Having some issues that need to debug, one is bash missing from alpine so need to switch to sh for our scripts, but seem to have problems with starting up, no useful messages.

kacf:
Make sure you are compiling with `CGO_ENABLED=0`, it doesn't work on alpine.

maciejmrowiec:
Thats JAVA app. Go error is from docker itself because entry point was using bash which is not installed.

maciejmrowiec:
@GregorioDiStefano  is there a Jenkins job?

GregorioDiStefano:
DigitalOcean ran out of high CPU droplets again, I manually intervened and started this job...

maciejmrowiec:
Looks like tests passed ;) yeyyy

maciejmrowiec:
Tested upgrade from 1.7.7 (there are no docs provided by conductor guys around upgrades).
Functionality still works, our workflow still works.

Downside: 
It seem we lose history of past jobs, conductor ui seem to not list any execution history from before the upgrade. There is nothing in application log around that so seem like expected behaviour. Not a big deal as decommissioning is all it have been doing so far.

PR #407, opened 2017-11-24
multitenancy s3 deployment test 

GregorioDiStefano:


GregorioDiStefano:
@kacf

tests/common_setup.py
```diff
@@ -192,3 +192,33 @@ def fin():
...
+                        use_common_files=False)
+
+
+    wait_for_containers(20, ["../docker-compose.yml",
+                             "../docker-compose.testing.yml ",
+                             "../docker-compose.tenant.yml",
+                             "../docker-compose.storage.minio.yml",
+                             "../docker-compose.storage.s3.yml"])
+
+    def fin():
```
kacf: Are you using French?? :-)

tests/tests/test_multi_tenancy.py
```diff
@@ -24,6 +24,7 @@
                           common_update_procedure
 from mendertesting import MenderTesting
 from tests import artifact_lock
+from boto3 import client
```
kacf: Any chance you could import this inside the function where you use it? The thing is that this will fail for anyone who doesn't have this library, and testing on Amazon will rarely be done locally.

GregorioDiStefano: yes, good point!

PR #410, opened 2017-12-01
Bump decomissioning worklfow version.

maciejmrowiec:
Was missed duriring previous fix, this fixes upgrade from previous version.

Changelog: none

### DIFF
--- a/conductor/server/workflows/decommission_device.json
+++ b/conductor/server/workflows/decommission_device.json
@@ -1,7 +1,7 @@
 {
     "name": "decommission_device",
     "description": "Removes device info from all services.",
-    "version": 2,
+    "version": 3,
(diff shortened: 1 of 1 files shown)


maciejmrowiec:
@GregorioDiStefano ugh, we forgot about this one. I would expect that without this patch after production upgrade from 1.2.x to 1.3.0 decommissioning would not fully work (workflow update would be skipped because version did not change). But it seem to be picked up by tests.

Will need to cherry-pick for 1.3.x

kacf:
@maciejmrowiec, @GregorioDiStefano: Has this been cherry-picked?

maciejmrowiec:
https://github.com/mendersoftware/integration/pull/433

PR #415, opened 2017-12-19
tests: Add missing setup_type for multitenancy with smtp. 

kacf:
Without its own setup_type a test may get the wrong kind of setup
before it starts.

Also add the optimization that skips initialization if the environment
is already set up.

Changelog: None


kacf:
@PiotrPrzybylak

PR #425, opened 2018-01-16
MEN-1192: Integration tests for spontanaeous-reboot hardening of the …

oleorhagen:
…client

Changelog: Title


tests/tests/test_state_scripts.py
```diff
@@ -632,6 +764,139 @@ class TestStateScripts(MenderTesting):
...
+            return
+
+        client = env.host_string
+        work_dir = "test_state_scripts.%s" % client
+
+        script_content = '#!/bin/sh\n\necho "$(basename $0)" >> /data/test_state_scripts.log\n'
+        # Even though this script is actually run twice, only show it once in the logs
+        script_stop_mender_on_entry = '#!/bin/sh\ntest "$(grep -c "$(basename $0)" /data/test_state_scripts.log)" -eq 0 && echo "$(basename $0)" >> /data/test_state_scripts.log && systemctl stop mender\nexit 0\n'
+        script_failure_content = script_content + 'sync\necho b > /proc/sysrq-trigger\n' # flush to disk before killing
+        script_reboot_once = '#!/bin/sh\ntest "$(grep -c "$(basename $0)" /data/test_state_scripts.log)" -eq 0 && echo "$(basename $0)" >> /data/test_state_scripts.log && sync && echo b > /proc/sysrq-trigger\necho "$(basename $0)" >> /data/test_state_scripts.log\nexit 0'
```
kacf: Could you break up these lines a bit? It's hard to read.

tests/tests/test_state_scripts.py
```diff
@@ -632,6 +764,139 @@ class TestStateScripts(MenderTesting):
...
+                test_set,
+                hosts=get_mender_clients())
+            return
+
+        client = env.host_string
+        work_dir = "test_state_scripts.%s" % client
+
+        script_content = '#!/bin/sh\n\necho "$(basename $0)" >> /data/test_state_scripts.log\n'
+        # Even though this script is actually run twice, only show it once in the logs
+        script_stop_mender_on_entry = '#!/bin/sh\ntest "$(grep -c "$(basename $0)" /data/test_state_scripts.log)" -eq 0 && echo "$(basename $0)" >> /data/test_state_scripts.log && systemctl stop mender\nexit 0\n'
```
kacf: I didn't get this. Why does it run twice?

tests/tests/test_state_scripts.py
```diff
@@ -632,6 +764,139 @@ class TestStateScripts(MenderTesting):
...
+        script_failure_content = script_content + 'sync\necho b > /proc/sysrq-trigger\n' # flush to disk before killing
```
kacf: Love it! :-)

tests/tests/test_state_scripts.py
kacf: Why are there commands after the sysrq call? They will never get executed, will they? Is it to make sure of that?

tests/tests/test_state_scripts.py
```diff
@@ -632,6 +764,139 @@ class TestStateScripts(MenderTesting):
...
+            # and the client will thus reboot straight after starting, and u-boot will
+            # fall back to the committed partition
+            if test_set.get("DoubleReboot", False):
+                token.verify_device_double_reboot()
+            else:
+                token.verify_reboot_performed(sleeptime=3, ntimes=3)
+
+            # if scripts are killed on a script that is run on the second partition,
+            # we need to wait for a second reboot
+            if test_set.get("DeviceDieSecondPartition", False) and len(
```
kacf: Hmm, is it possible to get rid of `DeviceDieSecondPartition`? I think you only test it here, so always together with `StopSecondPartition`, hence it is redundant.

tests/tests/test_state_scripts.py
```diff
@@ -632,6 +764,139 @@ class TestStateScripts(MenderTesting):
...
+                timeout = time.time() + 60
+                while timeout > time.time():
+                    logger.info("sleeping")
+                    time.sleep(3) # not in a hurry here, the client should be stopped
+                    output = run("journalctl -xn")
+                    logger.info(output)
+                    if "mender.service has finished shutting down" not in output:
+                        continue
+                    else:
+                        break
```
kacf: Have you actually seen the mender client still be active after `systemctl stop mender` has finished? I think it is supposed to be synchronous, so it should never happen.

If you still want to be sure, I have a better suggestion which is not timing dependent: check with `systemctl is-active mender`.

tests/tests/test_state_scripts.py
```diff
@@ -632,6 +764,139 @@ class TestStateScripts(MenderTesting):
...
+            while timeout > time.time():
+                logger.info("sleeping")
+                time.sleep(1)
+                if exists("/data/test_state_scripts.log"):
+                    break
+                else:
+                    pytest.fail("never found log files")
+
+            # make sure the client ended up on the right partition
+            if   "OtherPartition" in test_set.get("ExpectedFinalPartition", []):
```
kacf: Minor: extra spaces after if.

tests/tests/test_state_scripts.py
```diff
@@ -632,6 +764,139 @@ class TestStateScripts(MenderTesting):
...
+
+            # settle down
+            timeout = time.time() + 60
+            while timeout > time.time():
+                logger.info("sleeping")
+                time.sleep(1)
+                if exists("/data/test_state_scripts.log"):
+                    break
+                else:
+                    pytest.fail("never found log files")
```
kacf: What are you waiting for here? Given our greatly varying droplet performance, we have to use conditions when we wait instead of simply elapsed time, which is unreliable.

tests/tests/test_state_scripts.py
```diff
@@ -632,6 +764,139 @@ class TestStateScripts(MenderTesting):
...
+            broken_image=broken_image,
+            verify_status=True,
+            devices=[device_id],
+            scripts=[artifact_script_dir])[0]
+
+        try:
+
+            orig_part = Helpers.get_active_partition()
+
+            token = Helpers.place_reboot_token()
```
kacf: This has to be placed before `common_update_procedure` is called, otherwise there is a race condition.

tests/helpers.py
```diff
@@ -203,3 +206,32 @@ def verify_reboot_performed(self, max_wait=60*30):
...
+                    time.sleep(3)
+                    if not exists(self.dbootfile):
+                        logger.debug("reboot file has not reappeared, waiting")
+                        continue
+                    else:
+                        logger.debug("reboot file has reappeared. success!")
+                        return
+            except (BaseException):
+                logger.debug("no SSH connection. reboot in progress...")
+                continue
```
kacf: In general a double reboot detector is going to have race conditions because the agent might be reboot before we have time to check whether it rebooted. So this needs to be reworked somehow. Good news is that I'm deep into reboot semantics right now, so I can probably help you with this one. I'm also fairly certain that this is not the only place that tries to detect double reboot.

tests/tests/test_state_scripts.py
oleorhagen: This is for the case of a powerloss in commit_leave, in which case the script is rerun after it has booted.

oleorhagen: Because mender does not stop and start right at the execution point where it dies, it will in this case start over again from reboot_leave, as it is still on the same  partition as when it died (no reboot has happened).

tests/tests/test_state_scripts.py
oleorhagen: It will be rerun in the case of the device dying in commit_leave.

tests/tests/test_state_scripts.py
oleorhagen: I take it this no longer applies with the new implementation?

tests/tests/test_state_scripts.py
oleorhagen: I figured you would say this, and I agree. Initially I left it in for readability, but this is better handled with one parameter and a name change.

tests/tests/test_state_scripts.py
kacf: Oh, so you're fooling it into doing the same thing twice.

tests/tests/test_state_scripts.py
kacf: Ah right, I missed the conditional there. Another reason to break this up a bit for easier reading. :-)

tests/tests/test_state_scripts.py
kacf: The fact that it needs to be before `common_update_procedure` still applies. Translated to the new reboot detector this means that `common_update_procedure` should be inside the `with` scope.

tests/tests/test_state_scripts.py
```diff
@@ -632,15 +854,125 @@ class TestStateScripts(MenderTesting):
         "ArtifactFailure_Error_55", # Error for this state doesn't exist, should never run.
     ]
 
+    @pytest.mark.usefixtures("standard_setup_one_client_bootstrapped")
+    @pytest.mark.parametrize("description,test_set", REBOOT_TEST_SET)
```
oleorhagen: @kacf With the new reboot-detector, this turned out to be real simple, thus this is basically all there is to it,
the rest should look familiar to you, as I have merged this with ur master.

tests/tests/test_state_scripts.py
```diff
@@ -632,15 +854,125 @@ class TestStateScripts(MenderTesting):
...
+
+        # This is only needed in the case: die commit-leave,
+        # otherwise the device will get stuck in a boot-reboot loop
+        script_reboot_once =(
+        '''#!/bin/sh
+        if [ $(grep -c $(basename $0) /data/test_state_scripts.log) -eq 0 ]; then
+            echo "$(basename $0)" >> /data/test_state_scripts.log && sync && echo b > /proc/sysrq-trigger
+        fi
+        echo "$(basename $0)" >> /data/test_state_scripts.log
+        exit 0''')
```
oleorhagen: Readability should be significantly improved now.

oleorhagen:
@kacf added the final wait, and things should now be good for merging.

oleorhagen:
superseeded by: https://github.com/mendersoftware/integration/pull/432

PR #428, opened 2018-01-19
Small test improvements for organization testing

GregorioDiStefano:


GregorioDiStefano:
@kacf

tests/tests/test_create_organization.py
```diff
@@ -71,9 +71,12 @@ def test_duplicate_tenant(self):
     def test_duplicate_email(self):
         payload = {"request_id": "123456", "organization": "tenant-foo", "email":"some.user@example.com", "password": "asdfqwer1234", "g-recaptcha-response": "foobar"}
         rsp = requests.post("https://%s/api/management/v1/tenantadm/tenants" % get_mender_gateway(), data=payload, verify=False)
+        logging.info("first request to tenant adm returned: ", rsp.text)
```
kacf: Won't you get this information anyway if the assert fails?

If not I think I would make this `debug`.

GregorioDiStefano: no, the response isn't printed, we are only asserting on the statuscode

kacf: Ok.

PR #436, opened 2018-02-04
preauthorization tests 

mchalski:
https://tracker.mender.io/browse/MEN-1640

@mendersoftware/rndity @maciejmrowiec  @GregorioDiStefano 

NOTE: integration tests have been failing consistently on unrelated test cases - I'll keep restarting them   until maybe finally we'll get a green build. either way - the tests in this PR are green locally on my machine.

### DIFF
--- a/tests/tests/test_preauth.py
+++ b/tests/tests/test_preauth.py
@@ -24,11 +24,10 @@ from Crypto.PublicKey import RSA
 import time
 
 
-class TestPreauth(MenderTesting):
-    @pytest.mark.usefixtures("standard_setup_one_client")
-    def test_ok_preauth_and_bootstrap(self):
+class TestPreauthBase(MenderTesting):
+    def do_test_ok_preauth_and_bootstrap(self):
         """
-            Test the happy from preauthorizing a device to a successful bootstrap.
+            Test the happy path from preauthorizing a device to a successful bootstrap.
             Verify that the device/auth set appear correctly in admission API results.
         """
         client = get_mender_clients()[0]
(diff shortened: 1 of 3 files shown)


tests/tests/test_preauth.py
```diff
@@ -0,0 +1,214 @@
...
+
+        new_tenant_client("tenant-container", token)
+
+
+class Client:
+    """Wraps various actions on the client, performed via SSH (inside fabric.execute())."""
+
+    ID_HELPER = '/usr/share/mender/identity/mender-device-identity'
+    PRIV_KEY = '/data/mender/mender-agent.pem'
+    KEYGEN_TIMEOUT = 120
```
GregorioDiStefano: Increase the timeout

tests/tests/test_preauth.py
```diff
@@ -0,0 +1,214 @@
...
+        assert dev_preauth['device_identity'] == preauth_iddata_str
+        assert dev_preauth['key'] == preauth_key
+
+        # make one of the existing devices the preauthorized device
+        # by substituting id data and restarting
+        res = execute(Client.substitute_id_data, preauth_iddata, hosts=client)
+        res = execute(Client.restart, hosts=client)
+
+        # verify api results - after some time the device should be 'accepted'
+        time.sleep(30)
```
GregorioDiStefano: Increase this timeout

tests/tests/test_preauth.py
mchalski: what's your recommendation?

tests/tests/test_preauth.py
mchalski: what's your recommendation?

mchalski: btw this one should be instantaneous, I just restart the mender binary inside quemu (not the client itself)

mchalski: also, the tests are green. if we don't have good ideas for a precise number I'd leave it well alone for the time being.

GregorioDiStefano:
as expected, the 120 seconds timeout was exceeded in the second test run since this was merged, i increased it here:  https://github.com/mendersoftware/integration/pull/437/files

PR #446, opened 2018-02-09
Conductor event handler support

maciejmrowiec:
@mendersoftware/rndity 

Loading code + working example for the future references.


docker-compose.demo.yml
```diff
@@ -80,9 +80,12 @@ services:
             - ./conductor/server/entrypoint:/srv
             - ./conductor/server/workflows:/srv/workflows
             - ./conductor/server/tasks:/srv/tasks
+            - ./conductor/server/events:/srv/events
         entrypoint: /srv/start_conductor.sh
         environment:
             - CONFIG_PROP=config.properties
+        ports:
+            - "8080:8080"
```
maciejmrowiec: @GregorioDiStefano will this mess around with tests ? I find this very useful when testing

maciejmrowiec: on osx docker you cannot access containers directly, they need to be exposed (differs from linux docker)

GregorioDiStefano: https://github.com/mendersoftware/integration/blob/master/tests/run.sh#L21

Update this line to remove the 8080 port from being exposed while testing

maciejmrowiec: cool, done

conductor/server/events/demo_emit_post.json.skip
```diff
@@ -0,0 +1,25 @@
...
+                "name": "demo_emit_check_link",
+                "input": {
+                    "link": "${link}" 
+                }
+            },
+            "expandInlineJSON": true
+        }
+    ],
+    "active": true
+}
```
mchalski: nitpick: newline

conductor/server/tasks/demo_emit.json.skip
```diff
@@ -0,0 +1,12 @@
...
+        "name": "demo_emit",
+        "retryCount": 5,
+        "timeoutSeconds": 10,
+        "inputKeys": [],
+        "outputKeys": [],
+        "timeoutPolicy": "RETRY",
+        "retryLogic": "EXPONENTIAL_BACKOFF",
+        "retryDelaySeconds": 60
+    }
+]
```
mchalski: nitpick: newline

maciejmrowiec:
> to get it out of the way, demo_emit is actually a http task used down the line, and has nothing to do with eventing per se - correct? its definition doesn't say much about what it actually is, as usual - but the definition needs to be there.

Yes, task is there just so the `demo_emit_check_link` and `demo_emit_check_mender` workflows could properly execute on their own.

> an actual event definition lives in the demo_emit_start workflow which is what triggers it. the event will propagate as conductor:demo_emit_start:trigger_1, and it has 3 params. does anything happen with the ref_flow param that I'm missing? I can only see link being used later.

I think `ref_flow` it's just passed down to handler which could pass it down to action. Demo does not pass it down but at shows that it's available.

> the event will be caught by the demo_emit_post which will start 2 subordinate workflows via the start_workflow action. that's the end of the interesting stuff, since these just start some http tasks.

Pretty much.

> when I looked through the docs I saw that apart from start_workflow, supported actions are complete_task and fail_task. the doc is sparse, did you have a chance to play with these? I can't make any sense of them. could complete_task be used to e.g. send an event to a worker? I'm asking because this might be interesting for the future, if everything we can do is start workflows that's limited.

I haven't played with them, but from my understanding of docs they can be used to complete / fail another workflow (potentially here is where the ref_flow could be useful). There is also WAIT task type that waits fro external event to complete which sounds quite useful to sync completion of external workflows.

conductor/server/tasks/demo_emit.json.skip
maciejmrowiec: done

maciejmrowiec:
One more thing that I tested but it's not in example (to not introduce to much mess) is that you can have multiple event_handlers subscribing same event, and they both are being triggered. So relation between event and workflow/worker can be defined with worker event if multiple different ones interact with same trigger

mchalski:
@maciejmrowiec ok, thanks for the answers. so please just fix the newline and merge, I'm approving this.

PR #448, opened 2018-02-12
more random fixes

GregorioDiStefano:


tests/tests/test_preauth.py
```diff
@@ -205,10 +216,30 @@ def __wait_for_keygen():
...
+                sleepsec += 10
                 logging.info("waiting for key gen, sleepsec: {}".format(sleepsec))
             else:
+                time.sleep(5)
                 break
 
         assert sleepsec <= Client.KEYGEN_TIMEOUT, "timeout for key generation exceeded"
+
+    @staticmethod
+    def get_accepted_devices():
```
mchalski: the class is specifically for interacting with the client via ssh. this method should be added to `admission` perhaps? or as a helper to the test class. definitely not here.

PR #449, opened 2018-02-12
MEN-1767: Make the integration version available to the UI

oleorhagen:
Changelog: Title

The integration version is stored in a file, and made available to
the UI through an environment variable.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,6 +34,8 @@ services:
             service: mender-base
         networks:
             - mender
+        environment:
+            - INTEGRATION_VERSION
 
     #
     # mender-api-gateway
--- a/demo
+++ b/demo
@@ -29,6 +29,9 @@ if [[ "$1" == "up" ]]; then
     echo "Starting Mender demo environment.."
 fi
 
+# Pass this value on to the GUI container as an env variable
+export INTEGRATION_VERSION=$(git describe --tags --abbrev=0)
+
 exec docker-compose \
      -f docker-compose.yml \
      -f docker-compose.storage.minio.yml \


oleorhagen:
https://tracker.mender.io/browse/MEN-1767

extra/release_tool.py
```diff
@@ -721,6 +721,14 @@ def generate_new_tags(state, tag_avail, final):
         if not next_tag_avail[repo.git]['already_released'] and final:
             next_tag_avail[repo.git]['already_released'] = True
 
+    # Make the integration version available to the UI through an environment variable
+    # and store
+    for repo in REPOS.values():
+        if repo.git == "integration":
+            os.environ('INTEGRATION_VERSION', next_tag_avail[repo.git]['build_tag'])
```
GregorioDiStefano: Any point of having this line?

oleorhagen: Acceptance criteria

- [x] During release the integration version is stored in a file (e.g. json)
- [x] This file or its contents is mounted in a place accessible to the UI or made available to the UI *through environment variables*

docker-compose.yml
```diff
@@ -34,6 +34,8 @@ services:
             service: mender-base
         networks:
             - mender
+        volumes:
+          - ./extra/integration-version.json:/var/www/mender-gui/dist/integration-version.json
```
oleorhagen: @GregorioDiStefano something like this?
Not sure where it should be mounted in the container though.

estenberg:
Thanks! Will this works for Hosted as well, @maciejmrowiec ?

maciejmrowiec:
sure, doe this will correlate to hosted releases

kacf:
My proposition is: Use `git describe` somewhere in the `demo` script to pass the version of integration into gui using an environment variable. If we don't want the additional Git SHA which is added if the user has made subsequent commits, then we can use `git describe --abbrev=0`.

oleorhagen:
fyi I stopped the jenkins build

PR #467, opened 2018-04-04
release-tool: add mender-conductor

GregorioDiStefano:
Signed-off-by: Gregorio Di Stefano <greg.distefano@gmail.com>
Changelog: none

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -81,13 +81,14 @@ REPOS = {
     # indexing purposes.
     "mender-artifact": RepoName("mender-artifact", "mender-artifact", "mender-artifact", False),
     "mender-integration": RepoName("mender-integration", "integration", "integration", False),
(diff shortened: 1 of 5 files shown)


tests/tests/test_grouping.py
```diff
@@ -61,6 +61,7 @@ def validate_group_responses(self, device_map):
     def test_basic_groups(self):
         """Tests various group operations."""
 
+        time.sleep(10)
```
kacf: We should not add unconditional sleeps while waiting for conditions. We've had so many problems with this in the past. If we need to wait for something, wait for the condition to be true, not for a fixed time. Like you do in `admission.py` in this PR.

GregorioDiStefano: Yes! This was a just a PoC, the real fix is in admission.py!

PR #480, opened 2018-04-26
Switch default client container type to qemux86-64.

kacf:
Changelog: Title


kacf:
Merging immediately, because integration tests will *definitely* be broken without it, now that the containers have been updated.

PR #503, opened 2018-06-30
test_security.py: Ignore return code of grep. 

drewmoseley:
In the case that grep returns nothing, it also sets a return code
which will fail the test.  Ignore that by piping through cat and
just process the empty return string.

Changelog: Title


drewmoseley:
Testing my morty builds with this.

drewmoseley:
OK.  I'm having trouble triggering builds using release_tool.py to include this in my test run for https://github.com/mendersoftware/meta-mender/pull/539 but it seems like this would allow the integration tests to succeed.  @kacf thoughts?

PR #504, opened 2018-06-30
test_security.py: Ignore return code of grep.

drewmoseley:
In the case that grep returns nothing, it also sets a return code
which will fail the test.  Ignore that by piping through cat and
just process the empty return string.

Changelog: Title


drewmoseley:
Master branch version of https://github.com/mendersoftware/integration/pull/503

PR #505, opened 2018-07-09
FIX: remove reboot-hardening test for reboot-enter

oleorhagen:
Changelog: Title


kacf:
And remember also: No `Changelog: title` for tests! It's completely useless to users. Think about whether the commit changes the production software and whether it affects the user.

oleorhagen:
Turning off the formatter -_-

PR #508, opened 2018-07-11
demo: suppress warning on newer docker-compose versions

mirzak:
docker-compose version 1.21.0 and greater has deprecated
the --parallel option printing the following if used:

    WARNING: --parallel option is deprecated and will be
    removed in future versions.

This is because "--parallel" is the default mode as of
version 1.21.0.

Suppress warning by only using the --parallel option
if --no-parallel option does not exist because this
was added together with deprecation of --parallel.

Changelog: Title

Signed-off-by: Mirza Krak <mirza.krak@northern.tech>

### DIFF
--- a/demo
+++ b/demo
@@ -45,12 +45,20 @@ done
 docker images | grep -q 'mendersoftware/deployments'
 
 if [[ "$?" -eq 1 ]]; then
+    compose_args=""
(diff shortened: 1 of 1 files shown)


GregorioDiStefano:
Thanks!

PR #513, opened 2018-07-23
setup for backend-only tests

mchalski:
https://tracker.mender.io/browse/QA-26

### DIFF
--- a/backend-tests/run
+++ b/backend-tests/run
@@ -1,0 +1,55 @@
+#!/bin/bash
+set -e
+
+INTEGRATION_PATH=$(readlink -f ../$(dirname $0))
+export INTEGRATION_PATH
+
+COMPOSE_CMD="docker-compose -p backend-tests \
+            -f $INTEGRATION_PATH/docker-compose.yml \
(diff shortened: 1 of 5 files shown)


backend-tests/run
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+set -e
+
+INTEGRATION_PATH=$(readlink -f ../$(dirname $0))
+export INTEGRATION_PATH
+
+COMPOSE_CMD="docker-compose -p backend-tests \
+            -f $INTEGRATION_PATH/docker-compose.yml \
+            -f $INTEGRATION_PATH/docker-compose.storage.minio.yml \
+            -f $INTEGRATION_PATH/docker-compose.demo.yml \
```
maciejmrowiec: Demo will have a simulated device, is that intended ?

mchalski: I double checked - `docker-compose.demo.yml` doesn't contain the client.

It's the `./demo` script that pulls in the client layer, unless `NO_CLIENT` is specified, but we're not using that. So I think we're good here.

PR #516, opened 2018-08-07
parametrize backend tests

mchalski:
issues: https://tracker.mender.io/browse/QA-27

it was spotted during writing https://github.com/mendersoftware/mender-conductor/pull/31 that we have no backend test parametrization - no way of selecting compose files, or selecting the tests themselves (pytest -k etc.)

this PR fixes those issues.

it is not 100% clear how we'll deal with backend test selection (marks, regexes) but this should allow us to do whatever. it allows for freeform compose file selection and passing any args to the py.test command.

### DIFF
--- a/backend-tests/tests/test_dummy.py
+++ b/backend-tests/tests/test_dummy.py
@@ -16,3 +16,6 @@ import pytest
 
 def test_dummy():
     print('test_dummy executed')
+
+def test_dummy_multitenant():
+    print('test_dummy_multitenant executed')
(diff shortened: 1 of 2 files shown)


backend-tests/tests/test_dummy.py
```diff
@@ -16,3 +16,6 @@
 
 def test_dummy():
     print('test_dummy executed')
+
+def test_dummy_multitenant():
+    print('test_dummy_multitenant executed')
```
kacf: Did you mean to include this?

kacf:
#517 has been merged. Please rebase this to get the tests to run.

PR #520, opened 2018-08-23
remove admission service from the setup

kjaskiewiczz:
Issues: https://tracker.mender.io/browse/MEN-1957


extra/release_tool.py
```diff
@@ -102,7 +101,6 @@ def __init__(self, container, docker, git, has_container):
     "mender-conductor": "MENDER_CONDUCTOR_REV",
     "mender-conductor-enterprise": "MENDER_CONDUCTOR_ENTERPRISE_REV",
     "deployments": "DEPLOYMENTS_REV",
-    "deviceadm": "DEVICEADM_REV",
```
kacf: In addition to this, can you add

```
"DEVICEADM_REV": "master",
```

to `EXTRA_BUILDPARAMS`? The reason is that during release builds, Jenkins still has to clone the repo, even though it won't be used, because of the older versions that still have it. It can't clone selectively, unfortunately.

kjaskiewiczz: done

PR #521, opened 2018-08-24
backend-tests: account suspension

mchalski:
https://tracker.mender.io/browse/MEN-1715


kacf:
Just for my own curiosity: The plan is to run these using Travis, not Jenkins, correct?

mchalski:
@kacf correct

PR #538, opened 2018-09-27
QA-37: Fix inability to query repos that have been removed.

kacf:
We need to be able to query repos that have been in the product
earlier, but which have been removed, both when using the `--list`
argument and the `--version-of` argument. To achieve this we turn the
existing static list into a candidate map instead, and fill it based
on which versions are available either in the current working
directory or, if given the `--in-integration-version` argument, that
version.

Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -66,20 +66,20 @@ class RepoName:
         self.git = git
         self.has_container = has_container
 
-# All our repos, and also a map from docker-compose image name to all
+# All our repos, including repos that were part of our release before, and
+# aren't anymore. It's also a map from docker-compose image name to all
 # names. The key is container name, and thereafter the order is the order of the
 # RepoName constructor, just above.
 #
-# This is the main list of repos that will be used throughout the script. If you
-# add anything here, make sure to also update REPO_ALIASES (if there are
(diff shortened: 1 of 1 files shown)


extra/release_tool.py
```diff
@@ -66,20 +66,20 @@ def __init__(self, container, docker, git, has_container):
...
-REPOS = {
+# If you add anything here, make sure to also update GIT_TO_BUILDPARAM_MAP
+# (which tells the tool how to trigger Jenkins jobs.
+REPO_MAP = {
     "api-gateway": RepoName("mender-api-gateway", "api-gateway", "mender-api-gateway-docker", True),
     "mender-client-qemu": RepoName("mender-client", "mender-client-qemu", "mender", True),
     "mender-conductor": RepoName("mender-conductor", "mender-conductor", "mender-conductor", True),
     "mender-conductor-enterprise": RepoName("mender-conductor", "mender-conductor-enterprise", "mender-conductor-enterprise", True),
     "deployments": RepoName("mender-deployments", "deployments", "deployments", True),
+    "deviceadm": RepoName("mender-device-adm", "deviceadm", "deviceadm", True),
```
kacf: @kjaskiewiczz: It's back! :-)

But it's treated as a candidate now, and will be included or excluded depending on if it's present in `docker-compose.yml`.

PR #541, opened 2018-10-04
client: Use KVM automatically if available. Remove "./demo --kvm" option.

kacf:
For this we need to always run the client in a privileged container.

Changelog: Title

### DIFF
--- a/docker-compose.client.yml
+++ b/docker-compose.client.yml
@@ -10,3 +10,4 @@ services:
             - mender
         stdin_open: true
         tty: true
+        privileged: true
--- a/demo
+++ b/demo
@@ -2,10 +2,8 @@
 
 usage() {
     cat <<EOF
-$(basename $0) [--kvm] docker-options
+$(basename $0) [options] docker-options
 
---kvm
-	Enable KVM hardware acceleration of client VM.
 --no-client
 	Disable emulated client.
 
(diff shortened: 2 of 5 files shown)


drewmoseley:
Looks good.

docker-compose.client.yml
```diff
@@ -10,3 +10,4 @@ services:
             - mender
         stdin_open: true
         tty: true
+        privileged: true
```
mirzak: Have no idea how docker-compose and yml works, but there is an `--device` option to docker which is "more secure" then `--privileged`. It probably does not matter in the demo environment.

But one can use `--device /dev/kvm`, to provide access to KVM inside the docker container. This is what I normally do in my local setups working with QEMU. 

But again probably do not need it in demo environment but wanted to share it.

mirzak: Here is [link](https://docs.docker.com/engine/reference/commandline/run/#add-host-device-to-container---device) to docs.

kacf: I tried this but the problem is that if `/dev/kvm` isn't available (e.g. you're using VirtualBox), then the device cannot be mapped into the container, and it fails to launch instead of falling back to the emulated variant. I don't think it's possible to do any if logic in the docker-compose file to detect this. It would be more secure indeed, but I think we can live with the `--privileged` option since it's only a demo device.

mirzak:
Go for it.

PR #547, opened 2018-11-12
Mender changed version numbers to drop -ce/-ee and this breaks Mender

jgitlin-bt:
Mender changed version numbers to drop -ce/-ee (see: https://github.c…om/docker/docker-ce-packaging/pull/206) so we should just check to see if we're on version >2.x.x


jgitlin-bt:
Thanks @kacf apologies for missing that. I updated the commit message as requested and clarified the description.

For @estenberg: This is just a suggestion of how to handle this. My team discussed a number of options when the recent Docker update broke our Mender install, and this is the one we came up with. I believe the logic is sound: The new regex will match any version >= 2.x.x, including the `-ee` / `-cc` versions. But ultimately as long as version numbers of the recent Docker release work, I'm happy! 👍

estenberg:
Tested with no regression on string "Server Version: 17.05.0-ce".

This test might have outplayed its role right now (Mender works with all recent docker versions) and we can remove it in the future. But for thanks for maintaining @jgitlin-bt !

PR #550, opened 2018-11-28
backend-tests: devauth v2 api: preauth

mchalski:
https://tracker.mender.io/browse/MEN-1986

this is just a part of this task - 
preauth v2 ST + MT tests, plus some refactoring of common funcs.

still TODO:
- dev mgmt tests
- auth set mgmt tests

### DIFF
--- a/backend-tests/tests/test_devauth_v2.py
+++ b/backend-tests/tests/test_devauth_v2.py
@@ -144,6 +144,21 @@ class TestPreauthBase:
 
         assert r.status_code == 200
 
+        # device and authset changed status to 'accepted'
+        r = devauthm.with_auth(utoken).call('GET',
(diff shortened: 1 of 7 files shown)


backend-tests/tests/test_devauth_v2.py
```diff
@@ -0,0 +1,277 @@
...
+                                              pub,
+                                              priv,
+                                              tenant_token)
+
+        r = devauthd.call('POST',
+                          deviceauth_v1.URL_AUTH_REQS,
+                          body,
+                          headers=sighdr)
+
+        assert r.status_code == 200
```
kjaskiewiczz: what about checking if the aset and device status was updated after device request?

mchalski: fixed (stacked commit)

PR #553, opened 2018-12-03
MEN-2131: Added integration tests for the inventory update forcing fu…

oleorhagen:
…nction.

Changelog: None

### DIFF
--- a/tests/tests/test_basic_integration.py
+++ b/tests/tests/test_basic_integration.py
@@ -19,7 +19,7 @@ from common import *
 from common_setup import *
 from helpers import Helpers
 from common_update import update_image_successful, update_image_failed
-from MenderAPI import deploy, image
+from MenderAPI import deploy, image, inv
 from mendertesting import MenderTesting
 import shutil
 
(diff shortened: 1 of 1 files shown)


oleorhagen:
https://github.com/mendersoftware/mender/pull/336

tests/tests/test_basic_integration.py
```diff
@@ -139,12 +139,50 @@ def deployment_callback():
                 template = run(catcmd)
                 wait_count = 0
 
+    @pytest.mark.timeout(1000)
```
kacf: Why the timeout?

tests/tests/test_basic_integration.py
```diff
@@ -139,12 +139,50 @@ def deployment_callback():
...
+            execute(self.test_forced_inventory_update_from_clien,
+                    hosts=get_mender_clients())
+            return
+
+        # Give the image a really large wait interval.
+        sedcmd = "sed -i.bak 's/%s/%s/' /etc/mender/mender.conf" % ("\(.*PollInter.*:\)\( *[0-9]*\)", "\\1 1800")
+        out = run(sedcmd)
+        if out.return_code != 0:
+            logger.error(out)
+            pytest.fail("failed to set a large polling interval for the client.")
```
kacf: If `run()` receives a non-zero return code, it throws an exception, so this code will never be hit.

tests/tests/test_basic_integration.py
```diff
@@ -139,12 +139,50 @@ def deployment_callback():
...
+
         def deployment_triggered_callback():
-            output = run("mender --check-update")
+            output = run("mender --inventory-update")
             if output.return_code != 0:
                 logger.error(output)
-                pytest.fail("Forcing the update check failed")
-            logger.info("mender client has forced an update check")
+                pytest.fail("Forcing the inventory udpate failed")
+            logger.info("mender client has forced an inventory update")
```
kacf: You ended up removing `deployment_triggered_callback` and the `update_image_successful` call from the already existing `test_forced_update_check_from_client`, which probably was not the intention.

More generally I don't really see how this test is supposed to work. Why do you have to trigger an artifact update to send inventory? I also think the pass condition is too weak, it only checks return code, and does not check that anything actually worked. We should separate better between this test case and the previous one (update check), and check the log and make sure that the forced check was carried out, and at the same time, make sure that the other check was *not* carried out.

tests/tests/test_basic_integration.py
oleorhagen: If the client doesn't work, then the test will be stuck.

tests/tests/test_basic_integration.py
oleorhagen: http://docs.pyinvoke.org/en/1.2/api/runners.html#invoke.runners.Runner.run
http://docs.pyinvoke.org/en/1.2/api/runners.html#invoke.runners.Result
?

tests/tests/test_basic_integration.py
oleorhagen: True. I'm sorry. Was a bit too quick on this one.

oleorhagen:
This should be more to your liking, @kacf

tests/tests/test_basic_integration.py
kacf: No, it's ok, I'd rather have a long timeout, given our history with crappy droplets. I was just curious!

tests/tests/test_basic_integration.py
```diff
@@ -148,3 +148,69 @@ def deployment_triggered_callback():
...
+        output = run("cd /usr/share/mender/inventory && echo '#!/bin/sh\necho host=foobar' > mender-inventory-test && chmod +x mender-inventory-test")
+        if output.return_code != 0:
+            pytest.fail("Failed to create the extra inventory data script")
+
+        # Now that the client has settled into the wait-state, run the command, and check if it does indeed exit the wait state,
+        # and send inventory.
+        output = run("mender --send-inventory")
+        if output.return_code != 0:
+            logger.error(output)
+            pytest.fail("Forcing the inventory udpate failed")
```
kacf: The same would then apply here as well.

tests/tests/test_basic_integration.py
```diff
@@ -148,3 +148,69 @@ def deployment_triggered_callback():
...
+        logger.info("mender client has forced an inventory update")
+
+        # Give the client some time to send the inventory.
+        time.sleep(5)
+
+        # Make sure the client is forced from sleep.
+        client_log = run(catcmd)
+        if out.return_code != 0:
+            pytest.fail("Failed to get the client log")
```
kacf: And here.

PR #564, opened 2019-01-16
tests: un-skip TestCreateOrganization::test_success

mchalski:
after ES update this test passes locally.

changelog: none


kjaskiewiczz:
would be good to run integration tests on Jenkins so that we will be sure that it's working not only locally

PR #572, opened 2019-02-14
Client db migration tests

pasinskim:
Two tests for testing the migration from old client DB to the new one (MEN-23111).

### DIFF
--- a/legacy-client.yml
+++ b/legacy-client.yml
@@ -1,0 +1,14 @@
+version: '2'
+services:
+
+    #
+    # mender-legacy-client 
+    #
+    mender-client:
+            image: mendersoftware/mender-client-qemu:1.7.0
+            networks:
+                - mender
+            stdin_open: true
+            tty: true
+            privileged: true
+
--- a/legacy-client.yml
+++ b/legacy-client.yml
@@ -6,9 +6,5 @@ services:
     #
     mender-client:
             image: mendersoftware/mender-client-qemu:1.7.0
-            networks:
-                - mender
-            stdin_open: true
-            tty: true
-            privileged: true
+            
 
--- a/tests/common_setup.py
+++ b/tests/common_setup.py
@@ -99,6 +99,20 @@ def standard_setup_without_client():
 
     set_setup_type(ST_NoClient)
 
+@pytest.fixture(scope="function")
+def setup_with_legacy_client():
+    stop_docker_compose()
+    reset_mender_api()
+
+    docker_compose_cmd("-f ../docker-compose.yml \
+                        -f ../legacy-client.yml \
+                        -f ../docker-compose.storage.minio.yml \
+                        -f ../docker-compose.testing.yml up -d",
+                        use_common_files=False)
+
+    ssh_is_opened()
+    auth_v2.accept_devices(1)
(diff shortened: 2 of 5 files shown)


legacy-client.yml
```diff
@@ -0,0 +1,14 @@
...
+    #
+    # mender-legacy-client 
+    #
+    mender-client:
+            image: mendersoftware/mender-client-qemu:1.7.0
+            networks:
+                - mender
+            stdin_open: true
+            tty: true
+            privileged: true
```
kacf: Only specify what you need to override. The `image` alone should be enough here. See also comment in `common_setup.py`.

tests/common_setup.py
```diff
@@ -99,6 +99,20 @@ def standard_setup_without_client():
...
 
+@pytest.fixture(scope="function")
+def setup_with_legacy_client():
+    stop_docker_compose()
+    reset_mender_api()
+
+    docker_compose_cmd("-f ../docker-compose.yml \
+                        -f ../legacy-client.yml \
+                        -f ../docker-compose.storage.minio.yml \
+                        -f ../docker-compose.testing.yml up -d",
```
kacf: A better way is to keep the `docker-compose.client.yml` and then list `legacy-cient.yml` afterwards. Then it will only override those settings that are specified there.

legacy-client.yml
```diff
@@ -0,0 +1,14 @@
+version: '2'
```
kacf: Suggest putting this file inside `tests` to keep things tidy.

tests/run.sh
```diff
@@ -154,15 +154,15 @@ HTML_REPORT="--html=report.html --self-contained-html"
 UPGRADE_TEST_ARG=""
 SPECIFIC_INTEGRATION_TEST_ARG=""
 
-if ! pip2 list |grep -e pytest-xdist >/dev/null 2>&1; then
+if ! pip2 list --format=legacy |grep -e pytest-xdist >/dev/null 2>&1; then
```
kacf: Ah! I was wondering for a long time about this!

tests/tests/test_db_migration.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+
+        if not env.host_string:
+            execute(self.test_migrate_from_legacy_failure,
+                    hosts=get_mender_clients(),
+                    install_image=install_image)
+            return
+
+        dirpath = tempfile.mkdtemp()
+        script_content = '#!/bin/sh halt -p'
```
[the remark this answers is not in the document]
pasinskim: True. Fixing.

legacy-client.yml
pasinskim: Fixed!

tests/common_setup.py
pasinskim: Fixed!

legacy-client.yml
pasinskim: Not sure where we are using this. I can remove it, but it is the same as `docker-compose.client.yml`.

kacf: Yes, `version` must stay. I meant the file, not that line specifically.

PR #573, opened 2019-02-28
logo pushed as in the case of mendersoftware/mender repo

merlin-northern:
logo in the main README.md was broken.


merlin-northern:
forced push with signoff (no changes)

PR #574, opened 2019-03-04
MEN-2195: Modify test logic to reveal infinite loop bug

lluiscampos:
The test logic now will go through all the executed scripts and verify
that the expected sequence is found at least once but at a maximum of 20
times, so that an infinite cannot pass unnoticed again

Changelog: Reveal bug MEN-2195 in the integration test

### DIFF
--- a/tests/tests/test_state_scripts.py
+++ b/tests/tests/test_state_scripts.py
@@ -191,7 +191,7 @@ TEST_SETS = [
         "Failure_in_Download_Leave_script",
         {
             "FailureScript": ["Download_Leave_14"],
-            "ExpectedStatus": None,
+            "ExpectedStatus": "failure",
             "ScriptOrder": [
                 "Idle_Enter_08_testing",
                 "Idle_Enter_09",
--- a/tests/tests/test_state_scripts.py
+++ b/tests/tests/test_state_scripts.py
@@ -952,7 +952,7 @@ class TestStateScripts(MenderTesting):
         work_dir = "test_state_scripts.%s" % client
         deployment_id = None
         try:
-            script_content = '#!/bin/sh\n\necho "$(basename $0)" >> /data/test_state_scripts.log\n'
+            script_content = '#!/bin/sh\n\necho "echo `date --rfc-3339=seconds | sed -e e/\n//` $(basename $0)" >> /data/test_state_scripts.log\n'
             script_failure_content = script_content + "exit 1\n"
 
             old_active = Helpers.get_active_partition()
@@ -1114,6 +1114,9 @@ class TestStateScripts(MenderTesting):
 
     def verify_script_log_correct(self, test_set, log):
         expected_order = test_set['ScriptOrder']
+
+        # First remove timestamps from the log
+        log = [l.split(" ")[-1] for l in log]
 
         # Iterate down the list of expected scripts, and make sure that the log
         # follows the same list.
--- a/tests/tests/test_state_scripts.py
+++ b/tests/tests/test_state_scripts.py
@@ -62,8 +62,6 @@ TEST_SETS = [
                 "ArtifactCommit_Enter_05",
                 "ArtifactCommit_Leave_01_extra_string",
                 "ArtifactCommit_Leave_91",
-                "Idle_Enter_08_testing",
-                "Idle_Enter_09",
             ],
         },
     ),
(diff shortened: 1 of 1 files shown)


tests/tests/test_state_scripts.py
```diff
@@ -1170,17 +1137,21 @@ def verify_script_log_correct(self, test_set, log):
...
+                    and not expected_order[expected_pos % len(expected_order)].startswith("Idle_")):
                     # The Idle/Sync sequence is allowed to "wrap around" and start
                     # over, because it may take a few rounds of checking before the
                     # deployment is ready for the client.
-                    expected_pos = idle_pos
+                    expected_pos = 0
+
+                if expected_pos == len(expected_order):
+                    num_iterations = num_iterations + 1
+                    expected_pos = 0
```
kacf: I have to admit that I have some trouble following: You now set `expected_pos` back to zero, which will start from the beginning of the test_set? But won't that be wrong? We don't want the entire sequence to repeat, at least not the ones that get beyond `Download_Enter`. Can you explain a bit more?

lluiscampos: Right, I will then document the code inline.

The point here is that I want to measure the total number of iterations that the `log` list wraps around `expected_order` list (also taking into account the Idle/Sync wrapping allowed inside each iteration).

The previous code will only ensured that `log` list included `expected_order`, but didn't care about how many times.

lluiscampos: > We don't want the entire sequence to repeat, at least not the ones that get beyond `Download_Enter`. Can you explain a bit more?

Exactly, we don't want it to repeat. Hence we save the number of iterations (repetitions!) of the entire sequence and then verify that it has repeated only _few_ times.

Then, the "few" concept is quite complicated because sometimes the test frameworks takes very long to stop the client, so this can go several iterations of 5s each through the sequence. Hence my upper limit of 20 iterations below.

Does it make sense?

lluiscampos:
Added inline doc and removed the `Changelog` entry :+1:

tests/tests/test_state_scripts.py
kacf: I understand what you mean, but all the states starting with `Artifact` should execute exactly once, and never more than once, not even when the tests are slow. That's why I'm questioning whether setting `expected_pos` is correct, because we should not allow any repeated sequences of those states.

Do you agree that it is only the `Idle` / `Sync` sequence that should repeat?

lluiscampos: Hi again,

So I reworked the test following our conversation couple of days ago.

Working on the code, I figured out that I didn't need a new parameter in the test sets: using `ExpectedStatus` will actually say if the test is expect to loop (i.e. expected None) or not (i.e. expected success/failure). The test logic uses now this parameter to strictly verify when the case is allowed to loop or not.

There is an additional change I need you to verify. I think that the status expectation for `Failure_in_Download_Leave_script` was wrong, as it should be "failure". Please confirm or argue back :)

lluiscampos:
New change set pushed. Running integration tests again in jenkins.

tests/tests/test_state_scripts.py
```diff
@@ -1170,17 +1140,26 @@ def verify_script_log_correct(self, test_set, log):
...
                     # deployment is ready for the client.
-                    expected_pos = idle_pos
+                    expected_pos = 0
 
-                if (expected_pos < len(expected_order)
-                    and not expected_order[expected_pos - 1].startswith("Idle_")
-                    and expected_order[expected_pos].startswith("Idle_")):
-                    # New Idle sequence entered.
-                    idle_pos = expected_pos
+                if expected_pos == len(expected_order):
```
kacf: Minor: Is it possible to move this if-block up, to right after the `expected_pos = expected_pos + 1`. Then you will not need the `expected_pos % len(expected_order)` wraparound guard, and it should be a bit easier to read.

tests/tests/test_state_scripts.py
kacf: > I think that the status expectation for `Failure_in_Download_Leave_script` was wrong, as it should be "failure". Please confirm or argue back

You're absolutely right! I wondered why it passed in the first place, but I guess it wasn't checked at all if it was `None`.

As for the rest, yes, I believe you nailed it now. Clever to use `ExpectedStatus` as an indication for number of iterations!

tests/tests/test_state_scripts.py
lluiscampos: I made the change and run all state script tests locally :+1:

tests/tests/test_state_scripts.py
lluiscampos: Thanks! Good that we understood each other at the end :)

PR #575, opened 2019-03-05
MEN-2311: DB migration tests

kacf:


kacf:
@pasinskim: Check out what I did on top of yours. A few adjustments, but not much.

tests/tests/test_db_migration.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/python
+# Copyright 2017 Northern.tech AS
```
lluiscampos: 2019.

At least for new files we should try to keep them up to date

tests/tests/test_db_migration.py
```diff
@@ -0,0 +1,90 @@
...
+class TestDBMigration(MenderTesting):
+
+    @pytest.mark.usefixtures("setup_with_legacy_client")
+    def test_migrate_from_legacy_mender_v1_failure(self, install_image=conftest.get_valid_image()):
+        """
+            Start a legacy client (1.7.0) first and update it to the new one.
+
+            The test starts a setup with the 1.7.0 client and then updates it to
+            the current version. The update is failing first (due to reboot
+            command inside the artifact commit enter state script).
```
lluiscampos: Where is this `reboot` actually happening? Because in the ArtifactCommit_Enter is only an `exit 1` but not reboot.

tests/tests/test_db_migration.py
```diff
@@ -0,0 +1,90 @@
...
+                version=2)
+
+    @pytest.mark.usefixtures("setup_with_legacy_client")
+    def test_migrate_from_legacy_mender_v1_success(self, install_image=conftest.get_valid_image()):
+        """
+            Start a legacy client (1.7.0) first and update it to the new one.
+
+            The test starts a setup with the 1.7.0 client and then updates it to
+            the current version. After the first successful update, we are updating 
+            the client for the second time.
```
lluiscampos: Why do we need to want to verify the update twice?

kacf:
All issues fixed / clarified with comments.

PR #578, opened 2019-03-08
Release tool minor cleanup

lluiscampos:


extra/release_tool.py
```diff
@@ -816,13 +814,17 @@ def check_tag_availability(state):
...
         try:
             execute_git(state, repo.git(), ["rev-parse", state[repo.git()]['version']],
                         capture=True, capture_stderr=True)
             # No exception happened during above call: This is a final release
             # tag.
             tag_avail[repo.git()]['already_released'] = True
             tag_avail[repo.git()]['build_tag'] = state[repo.git()]['version']
+        except FileNotFoundError as err:
+            print(err)
+            missing_repos = True
```
kacf: Are you sure it produces a `FileNotFoundError`, and not a similar error, but from Git? I haven't checked, so you may be right. I just wanna make sure.

extra/release_tool.py
```diff
@@ -660,7 +658,7 @@ def execute_git(state, repo_git, args, capture=False, capture_stderr=False):
     else:
         git_dir = os.path.join(state['repo_dir'], repo_git)
 
-    if (not PUSH and is_push) or (DRY_RUN and is_change):
+    if (not PUSH and (is_push or is_change)):
```
kacf: This is wrong. By taking away the `DRY_RUN` condition, you are entering the dry-run section every time. You can remove the entire definition of `is_change` I think, since it is only used together with `DRY_RUN`. And maybe update the comment for this function as well.

kacf: Oh, actually no, I didn't take the switched parentheses into account. But the `is_change` removal and comment update still stand.

extra/release_tool.py
lluiscampos: What it fails is the `cd` into the directory.

If the directory exists but it is not a git repo, then I guess Git will fail with different error and then the exception will be raised, right?

extra/release_tool.py
lluiscampos: I don't really understand the intention here, actually. I might very well skip this change from the PR is we are not certain.

Why do we want to differentiate between the git/docker `push` that other commands now?

I think the original intention is that we still would like to have the git/docker changes locally made, then inspect manually, and finally push manually. But now I broke that, making `--no-push`, in effect, a `--dry-run` flag..., do you agree? If so, I suggest I revert my change and only modify the documentation of the dry-run flag

kacf: Yes, maybe that's better. And yes, the intention is roughly: `--dry-run` means make no local nor remote changes, `--no-push` means make no remote changes.

extra/release_tool.py
kacf: Ah, I see. Yes, you're right.

lluiscampos:
Reverted change regarding `--dry-run`

PR #580, opened 2019-03-14
Relax the max iterations criteria to 50 after failures in Jenkins

lluiscampos:
These failures are due to resource capacity and not actual test
failures. Anyhow, if the "infinite" loop bug shows again, the num of
iterations will be in the order of thousands.

Changelog: None

### DIFF
--- a/tests/tests/test_state_scripts.py
+++ b/tests/tests/test_state_scripts.py
@@ -952,7 +952,7 @@ class TestStateScripts(MenderTesting):
         work_dir = "test_state_scripts.%s" % client
         deployment_id = None
         try:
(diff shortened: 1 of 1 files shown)


lluiscampos:
New failures have been reported and the new criteria will not be enough anyway. Closing this PR as we require to debug the issue further.

lluiscampos:
The other failure report turned out to be another issue (related but not the same). Reopening PR

lluiscampos:
Pushed a second fix for error logging on this test, in case you want to have a look.

lluiscampos:
The test failure is due to another known issue now fixed: the master client container image being out of date.

Merging

PR #584, opened 2019-03-28
MEN-2408: Resolve docker credentials problems in integration

lluiscampos:
The user shall have rights to access the docker daemon, as required in
the official Mender docs, to be able to run this scripts.

Removed all the sudo calls from the Docker-related scripts and added a
check for it in verify-docker-versions script.

Changelog: Title

### DIFF
--- a/verify-docker-versions
+++ b/verify-docker-versions
@@ -24,10 +24,21 @@ compare_versions() {
   return 0
 }
 
+# test that user has rights to use docker
+
+DOCKER_ENGINE_INFO_STDERR=$(docker info 2>&1 >/dev/null)
+DOCKER_ENGINE_INFO_RETVAL=$?
+if [[ $DOCKER_ENGINE_INFO_RETVAL -ne 0 ]]; then
+  >&2 echo "Local user has insufficient permissions to connect to the docker daemon."
+  >&2 echo "Please add user to 'docker' group to be able to access it or check the documentation for your host OS."
+  >&2 echo "Be aware that if you were just added to the group, you need to log out and in again."
+  >&2 echo $DOCKER_ENGINE_INFO_STDERR
+  exit 1
+fi
 
 # test Docker Engine version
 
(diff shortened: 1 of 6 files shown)


lluiscampos:
@mirzak please have a look

verify-docker-versions
```diff
@@ -24,10 +24,19 @@ compare_versions() {
   return 0
 }
 
+# test that user has rights to use docker
+
+DOCKER_ENGINE_INFO_STDERR=$(docker info 2>&1 >/dev/null)
+if [[ -n "$(echo $DOCKER_ENGINE_INFO_STDERR | grep 'permission denied')" ]]; then
+  >&2 echo "Local user has insufficient permissions to connect to the docker daemon."
+  >&2 echo "Please add user to 'docker' group to be able to access it or check the documentation for your host OS."
```
kacf: I would add here that if you were just added to the group, you need to log out and in again. This is a fairly common problem for first-installers.

lluiscampos: Good suggestion. Added

verify-docker-versions
```diff
@@ -24,10 +24,20 @@ compare_versions() {
   return 0
 }
 
+# test that user has rights to use docker
+
+DOCKER_ENGINE_INFO_STDERR=$(docker info 2>&1 >/dev/null)
+if [[ -n "$(echo $DOCKER_ENGINE_INFO_STDERR | grep 'permission denied')" ]]; then
+  >&2 echo "Local user has insufficient permissions to connect to the docker daemon."
```
kacf: Btw, it just occurred to me that this won't work in any non-English locale. Is there a better way we can detect this? Is simply return code enough? `docker info` should never fail otherwise right?

lluiscampos: You are right! There are other parts of the script that use docker info output, though.

I don't know if there are other reasons for `docker info` to fail, that is why I originally decided on using the stderr output

Anyway, fixed now. Have a new look and I will merge

PR #588, opened 2019-04-16
set HSTS max-age via env/demo config files

mchalski:
https://tracker.mender.io/browse/MEN-2440

gateway has a new setting now - so set it in demo: https://github.com/mendersoftware/mender-api-gateway-docker/pull/103.

proxy also needs max-age=0, but it doesn't have the substitution logic that the gateway has - so simply have 2 config files, one for demo, one for prod.

### DIFF
--- a/template/prod.yml
+++ b/template/prod.yml
@@ -86,6 +86,7 @@ services:
         volumes:
             - ./template/keys-generated/certs/storage-proxy/cert.crt:/var/www/storage-proxy/cert/cert.crt:ro
             - ./template/keys-generated/certs/storage-proxy/private.key:/var/www/storage-proxy/cert/private.key:ro
+            - ./storage-proxy/nginx.conf:/usr/local/openresty/nginx/conf/nginx.conf
 
     mender-deployments:
         command: server --automigrate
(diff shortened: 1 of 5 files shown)


template/prod.yml
```diff
@@ -86,6 +86,7 @@ services:
         volumes:
             - ./template/keys-generated/certs/storage-proxy/cert.crt:/var/www/storage-proxy/cert/cert.crt:ro
             - ./template/keys-generated/certs/storage-proxy/private.key:/var/www/storage-proxy/cert/private.key:ro
+            - ./storage-proxy/nginx.conf:/usr/local/openresty/nginx/conf/nginx.conf
```
mzedel: just as a side note: this seems to be covered by https://github.com/mendersoftware/integration/blob/14c12f7aeb91ffba397a11b571515f91c75f6303/docker-compose.storage.minio.yml#L28 - but the change here should not interfere with the download links/ uploads anyway, as the proxy doesn't seem to be used in the S3 scenario.

mchalski: @mzedel I was actually meaning to remove the mount from minio.yml - fixed now.

the reason is that when there are competing mounts - say in demo and minio, or more impornantly prod and minio - which one will be selected depends on the order of files in the docker-compose statement...

now - minio has no mount, and it's up to demo/prod/testing to mount one

mzedel: sounds good, the prod and demo distinction makes it more explicit than hidden in the minio mount

PR #594, opened 2019-04-24
backend tests: remove inventory regex filtering tests

kjaskiewiczz:
regex filtering was an experimental feature and will not be present
in the inventory service anymore


kacf:
I think the failures may (again) have something to do with the client image. I can take it from here!

kacf:
Superseded by rebased https://github.com/mendersoftware/integration/pull/596.

PR #596, opened 2019-04-29
backend tests: remove inventory regex filtering tests

kacf:
regex filtering was an experimental feature and will not be present
in the inventory service anymore

changelog: none

Signed-off-by: Krzysztof Jaskiewicz <krzysiekjaskiewicz@gmail.com>


kacf:
[Passed now](https://mender-jenkins.mender.io/job/mender-builder/3297/). The red crosses are from earlier builds.

PR #597, opened 2019-05-02
Release tool fixes

kacf:


extra/release_tool.py
```diff
@@ -362,6 +358,14 @@ def yml_components(self):
     "mender-qa": "MENDER_QA_REV",
 }
 
+class JenkinsParam:
+    type = None
+    value = None
```
lluiscampos: Why do you want these defaults?

kacf: I just need something. Got better ones? :-)

PR #600, opened 2019-05-03
Men 1365 reenable ui tests

mzedel:


tests/e2e_tests/run
```diff
@@ -0,0 +1,110 @@
...
+    fi
+
+    for var in "${COMPOSE_FILES[@]}"
+    do
+        COMPOSE_CMD+=" -f ${var}"
+    done
+}
+
+run_tests() {
+    $COMPOSE_CMD build && $COMPOSE_CMD up -d
```
kacf: One disadvantage of this method is that you can't spin up new containers for a new test, so you need to ensure that:

* All tests need the same set of containers
* No test changes the containers in any way that affects a subsequent test

I can definitely see the advantages to this method also though (reproducibility, ease of setup, etc), so I'm not saying this is bad. But have you given it any thought?

tests/e2e_tests/test_ui.py
```diff
@@ -342,10 +354,11 @@ def get_args():
...
-    test.test_deployment_successful()
-    #pytest.main(args=[os.path.realpath(__file__)])#, "--url", args.url])
+    # TODO adjust to match correct ui elements
+    # test.test_artifact_upload()
+    # test.test_authorize_all()
+    # TODO adjust to match correct ui elements
+    # test.test_basic_inventory()
+    # test.test_deploy()
+    # test.test_deployment_in_progress()
+    # test.test_deployment_successful()
```
lluiscampos: Apart from removing these from here, we should mark them as skip for pytest

So for each of these tests, please add `@pytest.mark.skip` in the method declarations. You will also need to import pytest

tests/e2e_tests/Dockerfile.gui
```diff
@@ -0,0 +1,23 @@
+FROM python:2
```
lluiscampos: Is there any hard requirement for Python 2? The `test_ui.py` claims to be compatible with Python 3

tests/e2e_tests/test_ui.py
```diff
@@ -18,6 +18,8 @@
 import selenium
```
lluiscampos: Missing Northern.tech Copyright disclaimer, like the rest of Python test files

tests/e2e_tests/Dockerfile.gui
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /usr/src/app
+ENV CHROME_BIN="/usr/bin/google-chrome"
+ENV DISPLAY=":99.0"
+RUN apt-get -q update && apt-get -yq install chromium=73.0.3683.75-1~deb9u1
+RUN wget -q "https://chromedriver.storage.googleapis.com/73.0.3683.68/chromedriver_linux64.zip" -O /tmp/chromedriver.zip \
+  && unzip /tmp/chromedriver.zip -d /usr/bin/ \
+  && rm /tmp/chromedriver.zip
+RUN pip install -U --user paramiko 'fabric<2.0' PyYAML pytest requests pytest-xdist filelock selenium psutil
```
lluiscampos: I don't think we are using Fabric. If I am not mistaken, it should be dropped (which is probably the reason to use Python 2).

tests/e2e_tests/Dockerfile.gui
```diff
@@ -0,0 +1,23 @@
...
+RUN ln -s /usr/bin/chromedriver .
+COPY tests .
+RUN mkdir -p downloaded-tools && \
+  curl -SL --fail "https://mender.s3-accelerate.amazonaws.com/temp_master/core-image-full-cmdline-qemux86-64.ext4" \
+  -o core-image-full-cmdline-qemux86-64.ext4 && \
+  curl -SL --fail "https://d1b0l86ne08fsf.cloudfront.net/mender-artifact/master/mender-artifact" \
+  -o /usr/bin/mender-artifact && \
+  curl -SL --fail "https://stress-client.s3-accelerate.amazonaws.com/release/mender-stress-test-client" \
+  -o downloaded-tools/mender-stress-test-client && \
```
lluiscampos: I don't think we need the mender-stress-test-client neither

tests/e2e_tests/docker-compose.e2e-tests.yml
```diff
@@ -0,0 +1,16 @@
...
+      context: .
+    networks:
+      - mender
+    depends_on:
+      - mender-api-gateway
+      - mender-gui
+  mender-gui:
+    build:
+      dockerfile: ${GUI_REPOSITORY}/Dockerfile
+      context: ${GUI_REPOSITORY}
```
lluiscampos: I definitely need to learn `docker-compose` :wink:

This setup covers a very similar need I have with the `.deb` packaging task and I gave up (using `docker run --network ...` directly from my tests).

tests/e2e_tests/run
```diff
@@ -0,0 +1,110 @@
...
+             -f tests/e2e_tests/docker-compose.e2e-tests.yml"
+
+# by default just add minio, with COMPOSE_CMD_BASE this creates the standard onprem ST setup
+COMPOSE_FILES_DEFAULT=( "${INTEGRATION_PATH}/docker-compose.storage.minio.yml" )
+COMPOSE_FILES=()
+
+usage() {
+    echo "runner script for gui-specific integration tests"
+    echo ""
+    echo "./gui-tests"
```
lluiscampos: Name is incorrect.

I prefer to use `$0` to avoid these kind of typos.

tests/e2e_tests/run
```diff
@@ -0,0 +1,110 @@
...
+    echo -e "\t-h --help"
+    echo -e "\t-c --skip-cleanup \tleave containers running after tests"
+    echo -e "\t-f=<FILE>         \tspecify custom compose file(s); default files will not be used,"
+    echo -e "\t                  \tmake sure to specify all files you need"
+    echo -e "\t other args will be passed to the testing container's py.test command"
+    echo ""
+    echo -e "examples:"
+    echo -e "run all tests, default ST setup:"
+    echo -e "\t./run"
```
lluiscampos: If you replace above with `@0`, please do it here also

tests/e2e_tests/run
```diff
@@ -0,0 +1,110 @@
...
+
+usage() {
+    echo "runner script for gui-specific integration tests"
+    echo ""
+    echo "./gui-tests"
+    echo -e "\t-h --help"
+    echo -e "\t-c --skip-cleanup \tleave containers running after tests"
+    echo -e "\t-f=<FILE>         \tspecify custom compose file(s); default files will not be used,"
+    echo -e "\t                  \tmake sure to specify all files you need"
+    echo -e "\t other args will be passed to the testing container's py.test command"
```
lluiscampos: The container is not using `py.test` (which I don't know why).

Two options, we make the container to actually use `pytest`, or at least we modify here the comment.

tests/e2e_tests/run
mzedel: I am not sure that I fully understand your question do you mean the mandatory building part or using the composition as a whole?
Regarding the mandatory building:
To make good use of the pipeline I would pass on the build results and mount them in the gui container. same could go for the test container though...
Regarding the composition:
The existing tests basically build on top of each other (one uploading an artifact, the next using this to create a deployment, ...) so they do change the containers + the containers used are the whole demo setup. There also is no mocking/ initialization/ tear down per test, but rather a walkthrough for a new user.

The UI unit tests will happen apart from that in the future and will not be bound to the infrastructure.

lluiscampos:
I also created this task as a follow-up:

https://tracker.mender.io/browse/MEN-2498

tests/e2e_tests/Dockerfile.gui
mzedel: this is unfortunately required in one of the MenderAPI self checks, but as far as I can see it's just to check if there is an executable at that path...

tests/e2e_tests/Dockerfile.gui
mzedel: fabric is unfortunately required in one of the helper function - thus the python 2 requirement

tests/e2e_tests/Dockerfile.gui
mzedel: the `test_ui.py` file should be, but the `helpers.py` and some more transitive requirements also require fabric - the changes in the fabric api (to v2 or better v3) would probably require quite a few more changes to a range of tests..

lluiscampos: Okey, I understood now, thanks. I didn't realize the dependency on MenderAPI and helpers.

Yes, we have already tasks in QA project to migrate to Fabric >2 (and then Python 3), so we don't need to do them now.

tests/e2e_tests/Dockerfile.gui
lluiscampos: :+1:

tests/e2e_tests/Dockerfile.gui
lluiscampos: Ok. It makes little sense but let's not change MenderAPI in this task.

tests/e2e_tests/docker-compose.e2e-tests.yml
mzedel: let me know whenever you have a question :)

tests/e2e_tests/run
```diff
@@ -0,0 +1,109 @@
...
+    echo ""
+    echo "./$0"
+    echo -e "\t-h --help"
+    echo -e "\t-c --skip-cleanup \tleave containers running after tests"
+    echo -e "\t-f=<FILE>         \tspecify custom compose file(s); default files will not be used,"
+    echo -e "\t                  \tmake sure to specify all files you need"
+    echo ""
+    echo -e "examples:"
+    echo -e "run all GUI tests, from root of this repo, with default ST setup:"
+    echo -e "GUI_REPOSITORY='<path-to-local-gui-repo>' INTEGRATION_PATH=. ./tests/e2e_tests/$0"
```
lluiscampos: `$0` will already have the relative or absolute path for the script. It will represent whatever the user has written (for example `./run`, `./tests/e2e_tests/run` of `/home/lluis/full/path/to/repo/tests/e2e_tests/run`.

So, basically, you don't need to add anything here or in line 16.

tests/e2e_tests/test_ui.py
lluiscampos: As discussed in person, we cannot run them using pytest due to the dependencies. But can we add the decorators anyway?

tests/e2e_tests/run
mzedel: ah thanks, I thought this would just return the filename - should be fixed now...

PR #605, opened 2019-05-28
WIP: MEN-2433: Automate upload of demo application artifact in demo s…

oleorhagen:
…erver

Changelog: Title

### DIFF
--- a/demo
+++ b/demo
@@ -45,7 +45,7 @@ done
 # Check if the demo-artifact has been downloaded,
 # or if there exists a newer one in storage.
 DEMO_ARTIFACT_NAME="mender-demo-artifact.mender"
-wget -q --timestamping https://dgsbl4vditpls.cloudfront.net/mender-demo-artifact.mender
+wget -q --timestamping https://dgsbl4vditpls.cloudfront.net/${DEMO_ARTIFACT_NAME}
 
 retval=$?
 if [ $retval -ne 0 ]; then
@@ -79,12 +79,15 @@ fi
 export INTEGRATION_VERSION=$(git describe --tags --abbrev=0)
 
 if [[ "$1" != "up" ]]; then
+    # exec steals the shell, so unless docker-compose is not found,
+    # exit 1 will never happen.
     exec docker-compose \
          -f docker-compose.yml \
          -f docker-compose.storage.minio.yml \
          -f docker-compose.demo.yml \
-         $client \
+         $CLIENT \
          "$@"
+    exit 1
 fi
 
 # ------------------------------------------------------------------------------
@@ -93,22 +96,32 @@ fi
 #
 # ------------------------------------------------------------------------------
 
+# Make sure that the demo environment is brought down on SIGINT
+trap "docker-compose \
+    -f docker-compose.yml \
+    -f docker-compose.storage.minio.yml \
+    -f docker-compose.demo.yml \
+    $CLIENT \
+    down" SIGINT
+
+
 echo "Starting the Mender demo environment..."
 
 docker-compose \
     -f docker-compose.yml \
     -f docker-compose.storage.minio.yml \
     -f docker-compose.demo.yml \
-    $client \
+    $CLIENT \
     "$@" -d
 
-# DC_PID=$!
+USER='mender-demo@example.com'
+PASSWORD=$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c12)
 
-USER='myusername@example.com'; PASSWORD='mysecretpassword'
 MENDER_SERVER_USER="${USER}:${PASSWORD}"
 MENDER_SERVER_URI='https://localhost'
 
 # Block until the useradm service returns an HTTP response.
+# Which means that it is up.
 echo "Waiting for the useradm service to start up"
 curl --silent -k -X POST -u $MENDER_SERVER_USER\
      --fail\
@@ -120,12 +133,11 @@ curl --silent -k -X POST -u $MENDER_SERVER_USER\
 
 retval=$?
 case $retval in
-    22) ;; # Server 400 error, ie, server is up - continue.
     0)  ;; # User exists - continue.
-    *)
-       echo "Failed to access the mender-useradm service within the allotted two minutes."
+    22) ;; # Server 400 error, ie, server is up - continue.
+    *) echo "Failed to access the mender-useradm service"
        echo "Please check the docker-compose setup."
-       echo "exit code" $retval
+       echo "curl exit code" $retval
        exit $retval
        ;;
 esac
@@ -138,11 +150,11 @@ docker exec integration_mender-useradm_1 /usr/bin/useradm create-user --username
 retval=$?
 case $retval in
      0) echo "New user created." ;;
-     5) ;; # User already exists.
+     5) echo "The user already exists"; exit 0;; # No need to continue if a demo user exists
      *) exit $retval ;;
 esac
 
-echo "Getting JWT Token"
+echo "Getting JWT Token..."
 
 JWT=$(curl --silent -k -X POST -u $MENDER_SERVER_USER\
            --fail\
@@ -175,18 +187,19 @@ errout=$(curl -k --request POST\
 retval=$?
 
 case "$errout" in
-   " Artifact not unique"*) echo "Artifact already exists on the server" ;; # Artifact is already on the server.
+   " Artifact not unique"*) echo "Artifact already exists on the server" ;;
     "") echo "Artifact uploaded to the demo server" ;;
     *) echo "Uploading the demo artifact failed with error: " $errout
        exit $retval ;;
 esac
 
-trap "docker-compose \
-    -f docker-compose.yml \
-    -f docker-compose.storage.minio.yml \
-    -f docker-compose.demo.yml \
-    down" SIGINT
-
-sleep 2
+echo "****************************************"
+echo
+echo "Login password: " $PASSWORD
+echo
+echo "****************************************"
+echo "Please keep the password available, it will not be cached by the login script."
+echo "Backend ready. Press Enter to show logs or Ctrl-C to quit."
+read -se
 
 docker-compose logs --follow
--- a/demo
+++ b/demo
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -e
 
 usage() {
     cat <<EOF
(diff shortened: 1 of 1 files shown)


demo
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
```
oleorhagen: Unsure about this one. It now exits when one of the curl calls fails, and then I guess the background docker-compose process will be killed also.

demo
```diff
@@ -35,6 +36,13 @@ while [ -n "$1" ]; do
     shift
 done
 
+# Check if the demo-artifact has been downloaded,
+# or if there exists a newer one in the s3-storage.
+DEMO_ARTIFACT_NAME="core-image-full-cmdline-qemux86-64.mender"
+# wget -q --timestamping file://home/olepor/yocto/qemu-setup/build/tmp/deploy/images/qemux86-64/core-image-full-cmdline-qemux86-64.mender
```
oleorhagen: From what I've read it seems that `wget --timestamp` will do what is required, as long as the remote is up and running serving the demo-artifact.

demo
```diff
@@ -67,4 +75,56 @@ exec docker-compose \
...
      $CLIENT \
-     "$@"
+     "$@" &
+
+DC_PID=$!
+
+MENDER_SERVER_USER='myusername@example.com:mysecretpassword'
+MENDER_SERVER_URI='https://localhost'
+# Block until the useradm service returns an HTTP response.
+echo "Blocking until the useradm service is up!"
```
oleorhagen: The echo's are all here for debugging purposes, I guess they should be removed before the final PR. However, they do help follow the flow of the program here.

demo
```diff
@@ -67,4 +75,56 @@ exec docker-compose \
...
+     --form "artifact=@${DEMO_ARTIFACT_NAME}"\
+     "${MENDER_SERVER_URI}/api/management/v1/deployments/artifacts"
+
+echo "Demo Artifact was successfully uploaded to the server"
+
+# CURL_PID=$!
+
+# wait ${CURL_PID}
+# wait ${CURL_JWT_PID}
+wait ${DC_PID}
```
oleorhagen: Do not exit the script before the docker-compose call is finished.

demo
```diff
@@ -67,4 +75,56 @@ exec docker-compose \
...
+     --retry 60 \
+     --retry-max-time 120\
+     --header "Authorization: Bearer ${JWT}"\
+     --form "size=${ARTIFACT_SIZE_BYTES}"\
+     --form "artifact=@${DEMO_ARTIFACT_NAME}"\
+     "${MENDER_SERVER_URI}/api/management/v1/deployments/artifacts"
+
+echo "Demo Artifact was successfully uploaded to the server"
+
+# CURL_PID=$!
```
oleorhagen: Will be cleaned up.

demo
```diff
@@ -35,6 +36,13 @@ while [ -n "$1" ]; do
     shift
 done
 
+# Check if the demo-artifact has been downloaded,
+# or if there exists a newer one in the s3-storage.
+DEMO_ARTIFACT_NAME="core-image-full-cmdline-qemux86-64.mender"
+# wget -q --timestamping file://home/olepor/yocto/qemu-setup/build/tmp/deploy/images/qemux86-64/core-image-full-cmdline-qemux86-64.mender
+
+ARTIFACT_SIZE_BYTES=$(ls -l ${DEMO_ARTIFACT_NAME} | awk '{print $5}')
```
kacf: A more elegant way is:

```
ARTIFACT_SIZE_BYTES=$(stat -c %s ${DEMO_ARTIFACT_NAME})
```

demo
```diff
@@ -67,4 +75,56 @@ exec docker-compose \
      -f docker-compose.storage.minio.yml \
      -f docker-compose.demo.yml \
      $CLIENT \
-     "$@"
+     "$@" &
```
kacf: You will need to check whether `up` is one of the arguments passed, because `demo` script can be used for any docker-compose command, and especially `down` is frequently used. Something aka `echo "$@" | grep '\bup\b'` maybe?

demo
```diff
@@ -67,4 +75,56 @@ exec docker-compose \
...
+
+echo "New user created!"
+
+echo "Getting JWT Token"
+
+JWT=$(curl -k -X POST -u $MENDER_SERVER_USER\
+           --retry-connrefused\
+           --retry 60\
+           --retry-delay 10\
+           --retry-max-time 120\
```
oleorhagen: The wait times are also somewhat arbitrary. What would be good defaults?

demo
kacf: I think it won't actually. I think maybe we should resort to manual error checking for this file. The artifact already being uploaded is a very real possibility, if the environment is restarted, and this should be handled gracefully, with a message like "already uploaded" or something.

demo
oleorhagen: nice! strong unix foo

demo
oleorhagen: Yes, I agree. However, this brings us to the log-frequency of the containers. Most likely this will never be seen by an end-user, as the containers have already spewed out several hundred lines of logs. Any good ideas on this one? Could docker compose write to a file? and then after the initial setup (curls and wgets ) is done, then we can `tail -f  ` the file or something?

kacf: Hmm, good question. My initial idea for this implementation was to do this uploading in a container of its own, which would enable you to view its logs separately by using `docker logs`. But that is a more complicated solution and I kind of like the simplicity of your approach as well. It is possible to call `docker-compose up -d` which defers it to the background, including the logs. Later you can "resume" viewing the logs with `docker-compose logs -f`. But the great disadvantage with this method is that Ctrl-C stops working, and you can no longer stop the demo environment by pressing that combo. All you will do then is interrupt the log viewing.

A couple of suggestions, none of which are particularly good:

* Use the `docker-compose up -d` trick, but implement a `trap SIGINT` in the shell code to catch Ctrl-C and call `docker-compose down` manually.
* Simply accept that the log will scroll away together with the rest
* Move the code to an execute-and-exit container

demo
```diff
@@ -35,6 +35,19 @@ while [ -n "$1" ]; do
     shift
 done
 
+# Check if the demo-artifact has been downloaded,
+# or if there exists a newer one in storage.
+DEMO_ARTIFACT_NAME="mender-demo-artifact.mender"
+wget -q --timestamping https://dgsbl4vditpls.cloudfront.net/mender-demo-artifact.mender
```
oleorhagen: Now working with the demo artifact url!

demo
```diff
@@ -55,16 +68,118 @@ if [[ "$?" -eq 1 ]]; then
...
 # Pass this value on to the GUI container as an env variable
 export INTEGRATION_VERSION=$(git describe --tags --abbrev=0)
 
-exec docker-compose \
-     -f docker-compose.yml \
-     -f docker-compose.storage.minio.yml \
-     -f docker-compose.demo.yml \
-     $CLIENT \
-     "$@"
+if [[ "$1" != "up" ]]; then
```
oleorhagen: All other cases than up is handled like it used to.

demo
```diff
@@ -55,16 +68,118 @@ if [[ "$?" -eq 1 ]]; then
...
+retval=$?
+
+case "$errout" in
+   " Artifact not unique"*) echo "Artifact already exists on the server" ;; # Artifact is already on the server.
+    "") echo "Artifact uploaded to the demo server" ;;
+    *) echo "Uploading the demo artifact failed with error: " $errout
+       exit $retval ;;
+esac
+
+trap "docker-compose \
```
oleorhagen: In the end I went with your suggestion of `docker-compose -d`, and it seems to work fine :)

demo
```diff
@@ -55,16 +68,118 @@ if [[ "$?" -eq 1 ]]; then
...
+    -f docker-compose.yml \
+    -f docker-compose.storage.minio.yml \
+    -f docker-compose.demo.yml \
+    down" SIGINT
+
+sleep 2
```
oleorhagen: Give the user a little time to read before  docker logs swarm the terminal. Could possibly be a little longer.

demo
```diff
@@ -55,16 +68,118 @@ if [[ "$?" -eq 1 ]]; then
...
+       echo "Please check the docker-compose setup."
+       echo "exit code" $retval
+       exit $retval
+       ;;
+esac
+
+echo "Creating a new user..."
+
+# Create a new user.
+docker exec integration_mender-useradm_1 /usr/bin/useradm create-user --username=${USER} --password=${PASSWORD}
```
kacf: @estenberg: This is something we didn't think about: If the demo artifact is to be uploaded, then a user is needed first, which means we can no longer wait for it to be done manually. Is this OK? I suppose we can just remove this step. One step less in the instructions is actually better, right?

estenberg: Indeed, makes sense.

I have a couple of recommendations, though:
* use mender-demo@example.com as user email (for cosmetics)
* generate random password of 12 characters & output it to the user (for security)
* ouput a link to where user can log in to change user name and password

I am not sure how feasible this is, and am aware that this is "just" a demo server. However, default credentials is probably the most common way to get compromised and I don't think we should fall in this pitfall.

oleorhagen: Smart! However, should the password remain the same over multiple sessions, or a new one generated at each `up` command invocation?

kacf: I think if the user already exists, we can assume that the artifact is already uploaded and bail out early. This also saves us from complicated and almost certainly insecure methods of caching the generated password.

demo
kacf: Suggestion, use `read` instead:

```
echo "Backend ready. Press Enter to show logs or Ctrl-C to quit."
```

demo
```diff
@@ -55,16 +68,118 @@ if [[ "$?" -eq 1 ]]; then
...
+    "") echo "Artifact uploaded to the demo server" ;;
+    *) echo "Uploading the demo artifact failed with error: " $errout
+       exit $retval ;;
+esac
+
+trap "docker-compose \
+    -f docker-compose.yml \
+    -f docker-compose.storage.minio.yml \
+    -f docker-compose.demo.yml \
+    down" SIGINT
```
kacf: Btw, this should be moved up to right below the `up`, so that it works no matter where in the execution you press Ctrl-C.

oleorhagen:
@kacf, see what you think now.

demo
```diff
@@ -35,6 +35,19 @@ while [ -n "$1" ]; do
     shift
 done
 
+# Check if the demo-artifact has been downloaded,
+# or if there exists a newer one in storage.
+DEMO_ARTIFACT_NAME="mender-demo-artifact.mender"
+wget -q --timestamping https://dgsbl4vditpls.cloudfront.net/${DEMO_ARTIFACT_NAME}
```
kacf: Btw, one thing I forgot earlier: Should we do some quick `wget --help` and `curl --help` checks, and advice to install them if not available?

demo
```diff
@@ -55,16 +68,131 @@ if [[ "$?" -eq 1 ]]; then
...
+              --silent\
+              --show-error\
+              --retry-connrefused\
+              --retry-delay 10\
+              --retry 60 \
+              --retry-max-time 120\
+              --header "Authorization: Bearer ${JWT}"\
+              --form "size=${ARTIFACT_SIZE_BYTES}"\
+              --form "artifact=@${DEMO_ARTIFACT_NAME}"\
+              "${MENDER_SERVER_URI}/api/management/v1/deployments/artifacts" | jq .error)
```
kacf: `jq` should also be checked and documented I think.

oleorhagen:
> I reckon we should check for the tools we use. Otherwise it looks good. Looking forward to the test! :-)

Yes, I forgot! This would add weight to your container argument though. But I guess we can expect the user to have or install these tools on a demo host.

kacf:
> This would add weight to your container argument though. But I guess we can expect the user to have or install these tools on a demo host.

Yes, I was particularly thinking of `jq`, which is a bit more uncommon. But then I remembered that some of our update modules require this even on the client, so it should be ok for the backend as well. As long as our messages are clear and helpful, I think no one will be bothered by it!

oleorhagen:
@kacf, if u r happy now, I'm gonna go ahead and rebase and merge.

oleorhagen:
Superseeded by: https://github.com/mendersoftware/integration/pull/607

PR #606, opened 2019-05-29
MEN-2363: Disable virtual QEMU client by default.

kacf:
It can be turned on again by giving the `--client` argument to the
`demo` script.

Changelog: Commit

### DIFF
--- a/demo
+++ b/demo
@@ -4,8 +4,9 @@ usage() {
     cat <<EOF
 $(basename $0) [options] docker-options
 
---no-client
-	Disable emulated client.
+--client
+	Enable emulated client. To enable more than one client, you can use:
+	  $(basename $0) --client scale mender-client=2
 
 All other arguments passed to this command are passed directly to
 docker-compose, if you want to run the demo, run:
--- a/demo
+++ b/demo
@@ -15,6 +15,11 @@ docker-compose, if you want to run the demo, run:
 EOF
 }
 
+if [ "$#" -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+    usage
+    exit 1
+fi
+
 CLIENT=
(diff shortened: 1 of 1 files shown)


estenberg:
Make sense.

Is there a way to start one or more client afterwards? Would be nice as a separate step for demo purposes. :) Maybe a small script can be made to do this (e.g. ./demo-client 15 would ensure 15 clients are running). Not a big deal and can be handled later too.

kacf:
The `--client` argument already allows this, but it's not completely clear from the commit message I wrote, so I'll reword it.

demo
```diff
@@ -14,17 +14,22 @@ docker-compose, if you want to run the demo, run:
...
 while [ -n "$1" ]; do
-    if [ -n "${NO_CLIENT}" -o "$1" = "--no-client" ]; then
-        unset CLIENT
-        echo "-- disabling client containers"
-        if [ -n "$NO_CLIENT" ]; then
-            unset NO_CLIENT
-            continue
-        fi
+    if [ "$1" = "--no-client" ]; then
+        echo "--no-client argument is deprecated. Client is now disabled by default and can be enabled with --client"
```
mzedel: should this also be reflected in the `usage` description? I saw that the `--client` arg would be capable of defining the number of devices, but don't seem to understand how this would work - maybe put that instead?

kacf: Oops, you're right. Will fix that.

kacf:
@mzedel: This good? https://github.com/mendersoftware/integration/pull/606/commits/42a8c0de71e5e4a21ad9fa1f2527bdf74ac14e94

mzedel:
yes, that explains it - I wondered if this was passed into the client service, but this clarifies things!

kacf:
Thanks!

kacf:
Fixed another issue with `usage` not being displayed (preexisting).

PR #607, opened 2019-06-04
Men 2423 

oleorhagen:


oleorhagen:
Based off of: https://github.com/mendersoftware/integration/pull/605

demo
```diff
@@ -62,16 +80,144 @@ if [[ "$?" -eq 1 ]]; then
...
 # Pass this value on to the GUI container as an env variable
 export INTEGRATION_VERSION=$(git describe --tags --abbrev=0)
 
-exec docker-compose \
-     -f docker-compose.yml \
-     -f docker-compose.storage.minio.yml \
-     -f docker-compose.demo.yml \
-     $CLIENT \
-     "$@"
+if [[ "${!#}" != "up" ]]; then
```
oleorhagen: Done it like this, so that the command can take extra arguments, ie, -p `proj-name`. However, this is not currently consistent throughout the script. The previous method is `$1`. So I guess my question is: Which do you prefer?

demo
```diff
@@ -62,16 +80,144 @@ if [[ "$?" -eq 1 ]]; then
...
+    exit 1
+fi
+
+# ------------------------------------------------------------------------------
+#
+#       The following code will only be run in the case ./demo up [args]
+#
+# ------------------------------------------------------------------------------
+
+DOCKER_COMPOSE_PROJECT_NAME="${DOCKER_COMPOSE_PROJECT_NAME:-integration}"
```
oleorhagen: Needed since the integration test-framework does give custom project names when executing docker-compose.

demo
```diff
@@ -62,16 +80,144 @@ if [[ "$?" -eq 1 ]]; then
...
+
+echo "Docker_compose_proj name"
```
oleorhagen: Debug leftovers. Removed now, but not pushed.

demo
```diff
@@ -62,16 +80,144 @@ if [[ "$?" -eq 1 ]]; then
...
+
+docker-compose \
+    -f docker-compose.yml \
+    -f docker-compose.storage.minio.yml \
+    -f docker-compose.demo.yml \
+    $CLIENT \
+    "$@" -d
+
+USER='mender-demo@example.com'
+PASSWORD=$(< /dev/urandom tr -dc A-Za-z0-9 | head -c12)
```
oleorhagen: Acceptible generation of random passwords?

tests/conftest.py
```diff
@@ -131,6 +131,13 @@ def pytest_runtest_setup(item):
     logger = log.setup_custom_logger("root", item.name)
     logger.info("%s is starting.... " % item.name)
 
+    # Enable incremental tests for classes
+    # (see https://docs.pytest.org/en/latest/example/simple.html#incremental-testing-test-steps)
+    if "incremental" in item.keywords:
```
oleorhagen: Removed in head. Not pushed.

tests/conftest.py
```diff
@@ -149,6 +156,13 @@ def pytest_exception_interact(node, call, report):
 
 @pytest.mark.hookwrapper
 def pytest_runtest_makereport(item, call):
+
+    # Enables incremental test-runs in test-classes
```
oleorhagen: Removed in head. Not pushed.

tests/tests/test_demo_artifact.py
```diff
@@ -0,0 +1,115 @@
...
+
+# NOTE - The password is set on a per test-basis, as it is generated on the fly by
+# the demo script.
+auth = authentication.Authentication(
+    username='mender-demo', email='mender-demo@example.com')
+authv2 = auth_v2.auth = auth
+deploy = deployments.Deployments(auth, authv2)
+
+
+# @pytest.mark.incremental
```
oleorhagen: For now, the tests do not pass when run simultaneously. Any good solutions to this? I've read pytest has some dependency solutions. Thoughts?

tests/tests/test_demo_artifact.py
```diff
@@ -0,0 +1,115 @@
...
+                'up'
+            ],
+            cwd="..",
+            stdout=subprocess.PIPE,
+            env=test_env)
+        logging.info('Started the demo script')
+        password = ""
+        for line in iter(proc.stdout.readline, ''):
+            if "Login password:" in line.strip():
+                password = line[-13:-1]
```
oleorhagen: Kinda fugly, but works. Would you rather have it parsed via say a regex?

demo
kacf: TBH I don't really understand how it works. Man page says "The entire command line typed so far". Won't this always fail to match then, since "!=" compares the whole string?

demo
kacf: :+1:

No need to connect to a quantum randomness source for this, it's just a demo. ;-)

tests/tests/test_demo_artifact.py
kacf: Why is that? Doesn't the namespace keep them separate? Or is it the downloaded file which is the problem?

tests/tests/test_demo_artifact.py
kacf: I think it's fine. A change to that string is likely to break it no matter what we do, so we might as well KISS (no, not the HR-violating kind).

tests/tests/test_demo_artifact.py
```diff
@@ -0,0 +1,115 @@
...
+                break
+
+        logging.info('The login password:')
+        logging.info(password)
+        assert len(password) == 12
+
+        auth.password = password
+
+        yield proc
```
kacf: After you have pushed the `addfinalizer` you can get rid of the rest of this function. `addfinalizer` is generally preferred over `yield`, since it will clean up even if the fixture crashes midway through, `yield` won't.

tests/tests/test_demo_artifact.py
```diff
@@ -0,0 +1,115 @@
...
+        proc.send_signal(signal.SIGINT)
+        proc.wait()
+        assert proc.returncode == 0
+
+    # Give the test a timeframe, as the script might run forever,
+    # if something goes awry, or the script is not brought down properly.
+    @pytest.mark.timeout(1000)
+    def test_demo_artifact_upload(self, run_demo_script):
+        """Tests that the demo script does indeed upload the demo Artifact to the server."""
+        proc = run_demo_script
```
kacf: `proc` is unused.

tests/tests/test_demo_artifact.py
oleorhagen: fixed in https://github.com/mendersoftware/integration/pull/607/commits/943d0d996b9ed48886213c7f95c87699af7e6e92

demo
```diff
@@ -42,6 +47,19 @@ while [ -n "$1" ]; do
...
+DEMO_ARTIFACT_NAME="mender-demo-artifact.mender"
+wget -q --timestamping https://dgsbl4vditpls.cloudfront.net/${DEMO_ARTIFACT_NAME}
+
+retval=$?
+if [ $retval -ne 0 ]; then
+    echo "Failed to download the demo artifact"
+    exit $retval
+fi
+
+ARTIFACT_SIZE_BYTES=$(stat -c %s ${DEMO_ARTIFACT_NAME})
```
lluiscampos: As we just learn in [another PR](https://github.com/mendersoftware/mender/pull/404), this will not work for Mac OS X.

demo
```diff
@@ -62,16 +80,138 @@ if [[ "$?" -eq 1 ]]; then
...
+
+echo "Creating a new user..."
+
+# Create a new user.
+docker exec ${DOCKER_COMPOSE_PROJECT_NAME}_mender-useradm_1 /usr/bin/useradm create-user --username=${USER} --password=${PASSWORD}
+
+retval=$?
+case $retval in
+     0) echo "New user created." ;;
+     5) echo "The user already exists"; exit 1;; # No need to continue if a demo user exists
```
lluiscampos: I don't understand. Do you assume that if the demo user exists the artifact has been already uploaded?

demo
```diff
@@ -62,16 +80,138 @@ if [[ "$?" -eq 1 ]]; then
...
+              --retry-connrefused\
+              --retry-delay 2\
+              --retry 60 \
+              --retry-max-time 120\
+              --header "Authorization: Bearer ${JWT}"\
+              --form "size=${ARTIFACT_SIZE_BYTES}"\
+              --form "artifact=@${DEMO_ARTIFACT_NAME}"\
+              "${MENDER_SERVER_URI}/api/management/v1/deployments/artifacts" | jq .error)
+
+retval=$?
```
lluiscampos: The retval here is for the `jq .error` command, I don't think that is your intention. You coud set this `pipefail` flag, but I don't know if that is POSIX standard or not. @kacf ?

demo
oleorhagen: The number of arguments is the last argument ;) tad hacky mby

lluiscampos:
Reminder: we need to update mender-docs together with this PR

lluiscampos:
Note: but please do not merge this PR nor the mender-docs one until we branch out for the release

demo
oleorhagen: Yes. That is the assumption. You want it explicitly checked?

oleorhagen: The reason it is done like this, is that a password is generated randomly for each user, so if the user exists already, we don't have the password. Maybe there should be an option for passing in a password through the command line?

demo
oleorhagen: I will split this up into two parts. And handle the errors independently. `pipefail` is a bashism btw ;)

demo
lluiscampos: Now I see the problem, good point. I was just wondering about what if the user is created correctly but the Artifact is not uploaded for some reason. Then the user will need to run `./demo down` before retrying?

Maybe we can add such information in the error message: "run this if you want to delete existing user and then rerun this to create a new one".

oleorhagen: Yes, good idea. I'll incorporate the error message now.

demo
```diff
@@ -62,16 +80,138 @@ if [[ "$?" -eq 1 ]]; then
...
+       exit $retval ;;
+esac
+
+echo "****************************************"
+echo
+echo  "Login password: ${PASSWORD}"
+echo
+echo "****************************************"
+echo "Please keep the password available, it will not be cached by the login script."
+echo "Backend ready. Press Enter to show logs or Ctrl-C to quit."
```
lluiscampos: So my point from yesterday demo.

The verb "quit" here IMO is not clear about what it will do. My first read is like now is ready, running in the background (because I don't see anything in the console), so I can "quit" the script and go to the website in my Firefox.

For the text to be a bit clearer, I suggest something like: "Backend ready and running in the background. Pres Enter to show logs. Pres Ctrl-C to stop the backend and quit."

I leave it up to you, but one important thing to me is to separate the actions with something more than "or".

oleorhagen: +1 Incorporated

demo
kacf: Suggestion: `echo "$@" | grep '\bup\b'`

demo
kacf: Indeed, good catch. The `curl` error will go unnoticed here. Feel free to use `pipefail` and other bashisms, the script is explicitly called with `bash`, not `sh`.

oleorhagen:
@kacf @lluiscampos This should be pretty much good to go now. If you guys are pleased, I am going to rebase and squash. I am also thinking that this should solve both 2433 and this task 2423. Thoughts?

oleorhagen:
Docs changes: https://github.com/mendersoftware/mender-docs/pull/707

tests/run.sh
```diff
@@ -176,8 +176,8 @@ if [[ -n $SPECIFIC_INTEGRATION_TEST ]]; then
 fi
 
 if [ $# -eq 0 ]; then
-    py.test $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT --runfast --runslow $UPGRADE_TEST_ARG $SPECIFIC_INTEGRATION_TEST_ARG $DEFAULT_TESTS
+    py.test $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT --runfast --runslow $UPGRADE_TEST_ARG $SPECIFIC_INTEGRATION_TEST_ARG $DEFAULT_TESTS -k 'test_demo'
     exit $?
 fi
 
-py.test $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT "$@" $DEFAULT_TESTS
+py.test $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT "$@" -k 'test_demo'
```
kacf: Don't forget to remove these two!

PR #608, opened 2019-06-04
allowed access to qemu container from localhost through port 3000

mzedel:
Changelog: None

### DIFF
--- a/demo
+++ b/demo
@@ -21,7 +21,7 @@ if [ "$#" -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
 fi
 
 CLIENT=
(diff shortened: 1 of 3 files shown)


lluiscampos:
The error in the test seems related to this change, we need to fix before merging/releasing

mzedel:
yep, but according to the integration repository there should not be another exposed 3000... is there some jenkins config using 3000?

lluiscampos:
A lot cleaner now.

@kacf Should we wait for the integ tests? Seeing the full change-set now, Not necessary IMO

kacf:
This failed last time. I think we should wait, to be sure.

PR #610, opened 2019-06-07
MEN-2571: Unify demo scripts

kacf:
```
commit dfb621a5a36b2772f16e7eddf48c3910e612c745
Date:   Fri Jun 7 09:51:01 2019

    Unify demo scripts by removing all except the one called 'demo'.
    
    We don't actually need the `/etc/hosts` meddling which exists in the
    `up` script, since Docker works just fine with its own internal DNS
    server. The other scripts provide no extra functionality and are
    completely redundant.
    
    The only thing we os keep the `verify-docker-versions`, which was not
    in `demo` before.
    
    MEN-2571
    
    Changelog: Title
    
commit 7f255be4ff7cedfb4bf34a28dc8902b5ce71e832
Date:   Thu Jun 6 11:21:23 2019

    Make sure `docker-compose stop` always brings down the client too.
    
    Changelog: None
    
```


kacf:
I tested this with our Getting Started guide. No ill effects AFAICS.

kacf:
Also [docs for this](https://github.com/mendersoftware/mender-docs/pull/710).

lluiscampos:
By the way, regardless of the release, this is great clean up. Thank you!

kacf:
> So if Docker has its own DNS, why did we had earlier the hook on /etc/hosts?

TBH I don't know. That part of the script is ancient, from the first year of Mender. I suspect that it was part of experimentation, and we never formally concluded whether we needed it or not. There is also the chance that I'm wrong, and we do need it for something I haven't thought of. But I haven't found anything, and we really need this fix...

kacf:
Merged into 2.1.x as well.

lluiscampos:
Thanks!

PR #611, opened 2019-06-12
 Fix bash syntax in variable assignment

lluiscampos:
Changelog: None


lluiscampos:
The two failing tests are due to the spurious error at setup. The changes are definitely safe :)

PR #612, opened 2019-06-17
Add option to skip download of requirements

lluiscampos:
For GitLab runners, we will prepare the native tools in the path prior
to the execution of the script, while we will not have the yocto
environment avaiable neither.

Changelog: None


lluiscampos:
Spurious error. Merging.

PR #615, opened 2019-06-21
FIX: Give the test_demo_artifact test a clean test environment 

oleorhagen:
Changelog: None


oleorhagen:
The `test_demo_artifact` test did not clean up the docker environment prior to starting the tests. This led to a fair amount of spurious fails in the integration tests, dependent upon the test ordering that pytest chose at the given run.

lluiscampos:
Good catch :+1:

PR #618, opened 2019-06-24
MEN-2035: Read-only filesystem support and test

kacf:


tests/common.py
```diff
@@ -23,34 +23,10 @@
...
-ST_ShortLivedAuthToken = 6
-ST_CustomSetup = 7
-ST_MultiTenancyNoClient = 8
-ST_OneClientsBootstrapped_AWS_S3_MT = 9
-ST_MultiTenancyNoClientWithSmtp = 10
-ST_Failover = 11
-ST_LegacyClient = 12
-ST_OneDockerClientBootstrapped = 13
-
-
```
lluiscampos: Good clean-up :+1:

tests/run.sh
```diff
@@ -186,4 +194,4 @@ if [ $# -eq 0 ]; then
     exit $?
 fi
 
-py.test $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT $pass_args $DEFAULT_TESTS
+python -m pytest $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT $pass_args $DEFAULT_TESTS
```
lluiscampos: :+1:

tests/run.sh
```diff
@@ -104,34 +100,46 @@ function get_requirements() {
...
-        eval "$(cd "$BUILDDIR" && bitbake -e core-image-minimal | grep '^export PATH=')":"$PATH"
-    fi
-
-    cp -f "$BUILDDIR/tmp/deploy/images/$MACHINE_NAME/core-image-full-cmdline-$MACHINE_NAME.ext4" .
-
-    # mender-stress-test-client is here
-    export PATH=$PATH:~/go/bin/
-
-elif [[ -n "$DOWNLOAD_REQUIREMENTS" ]]; then
+if [[ -n "$DOWNLOAD_REQUIREMENTS" ]]; then
```
lluiscampos: I think this is not correct. For a Jenkins build, it will download mender-artifact and mender-stress-client from AWS instead of using the built ones from the current build run.

In most of the cases won't matter, though, but it is a change from the previous logic. Is it intentional?

tests/run.sh
```diff
@@ -104,34 +100,46 @@ function get_requirements() {
...
     get_requirements
 fi
 
+mkdir -p output
+ret=0
+docker run --rm --privileged --entrypoint /extract_fs -v $PWD/output:/output \
+       mendersoftware/mender-client-qemu:$(../extra/release_tool.py -g mender-client-qemu) || ret=$?
+if [ $ret -eq 0 ]; then
+    # There is `extract_fs` support. Get the R/O image too.
+    docker run --rm --privileged --entrypoint /extract_fs -v $PWD/output:/output \
```
lluiscampos: Now I understand the script :)

Nice design. I'll modify the GitLab Pipeline accordingly.

tests/tests/test_basic_integration.py
```diff
@@ -27,17 +27,23 @@ class TestBasicIntegration(MenderTesting):
 
 
     @MenderTesting.fast
-    @pytest.mark.usefixtures("standard_setup_one_client_bootstrapped")
-    def test_double_update(self):
+    @pytest.mark.usefixtures("standard_setup_one_rofs_client_bootstrapped")
+    def test_double_update_rofs(self):
         """Upload a device with two consecutive upgrade images"""
```
lluiscampos: Can we add to the comment that the test is using a ro fs?

tests/run.sh
kacf: Ah, good catch, I was a bit too quick here. I'm still testing this, so I'll probably get an error soon.

tests/tests/test_basic_integration.py
kacf: Yes! I meant to do it, but it got lost along the way.

tests/run.sh
```diff
@@ -104,34 +100,46 @@ function get_requirements() {
...
 if [[ $1 == "--get-requirements" ]]; then
     get_requirements
     exit 0
 fi
 
 dd if=/dev/zero of=large_image.dat bs=300M count=0 seek=1
 
-if [[ -n "$BUILDDIR" ]]; then
-    # Get the necessary path directly from the build.
+# mender-stress-test-client is here
```
lluiscampos: Just for the record, mender-artifact and the Artifact generators are also here. These are all the dependencies for the integ tests, really.

lluiscampos:
Superseded by #625

Closing

PR #619, opened 2019-06-24
Remove the in-progress check for test_demo_artifact_install

oleorhagen:
Changelog: None

Signed-off-by: Ole Petter <olepor@matnat.uio.no>

### DIFF
--- a/demo
+++ b/demo
@@ -142,8 +142,11 @@ PASSWORD=$(< /dev/urandom 2>/dev/null tr -dc A-Za-z0-9 | head -c12)
 MENDER_SERVER_USER="${USER}:${PASSWORD}"
 MENDER_SERVER_URI="https://localhost"
 
+RETRY_LIMIT=5
+
 # Block until the useradm service returns an HTTP 4xx response
 echo "Waiting for the useradm service to start up..."
+RETRIES=0
 while :
 do
     curl --silent -k -X POST -u $MENDER_SERVER_USER\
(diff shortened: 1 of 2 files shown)


oleorhagen:
It turns out the build-slave, together with mender-artifact-v3 and the new client is too fast, so that the installation is done before the API requests for in progress have been processed. WOW!

oleorhagen:
@kacf If you could have a look at the demo script now, please :)
integration passes locally.

demo
```diff
@@ -144,84 +144,97 @@ MENDER_SERVER_URI="https://localhost"
...
     case $retval in
-        0)  USER_ADM_SERVICE_UP=true ;; # User exists - continue.
-        22) USER_ADM_SERVICE_UP=true ;; # Server 400 error, ie, server is up - continue.
+        0)  break ;; # User exists - continue.
+        22) break ;; # Server 400 error, ie, server is up - continue.
         *) echo "Failed to access the mender-useradm service"
            echo "Please check the docker-compose setup."
            echo "curl exit code" $retval
-           exit $retval
+           # exit $retval
```
kacf: This will retry infinitely. I think we should give up after three attempts or so. Also, I think we should change the message with the new behavior. Maybe something like "Failed to access the mender-useradm service. Retrying...", and then save the other part for when we give up.

The same goes for the other retry loops.

demo
```diff
@@ -144,84 +144,97 @@ MENDER_SERVER_URI="https://localhost"
...
            ;;
     esac
-    sleep 2
+    sleep 5
 done
 
 
-echo "Creating a new user..."
+echo "Trying to create a new user..."
```
kacf: I think the original message is better. Let's be optimistic! :-)

PR #621, opened 2019-06-26
Fix the inconsistency with the Docker project name in demo script

lluiscampos:
Some parts were relying on the env variable, but that assumes that the
caller passes a -p with the same name.

The design was broken as it cannot run in standalone mode.

Might this be the reason behing the enigmatic GitLab failures on these
tests?

Changelog: None


lluiscampos:
@oleorhagen Please have a look at this PR

demo
```diff
@@ -22,6 +22,10 @@ if [ "$#" -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
     exit 1
 fi
 
+# For the integration tests the name of the docker-compose project is
+# generated by pytest
+DOCKER_COMPOSE_PROJECT_NAME="${DOCKER_COMPOSE_PROJECT_NAME:-integration}"
+
```
lluiscampos: One missing detail, though, will be to check if the user has passed a "-p something" in the arguments and remove it, else you will be passing it two times to docker-compose

lluiscampos:
Superseded by #623 

Closing

PR #623, opened 2019-06-27
Demofix

oleorhagen:


oleorhagen:
Superseeds: https://github.com/mendersoftware/integration/pull/621

tests/tests/test_demo_artifact.py
```diff
@@ -27,6 +27,11 @@
 from mendertesting import MenderTesting
 
 
+class DemoScriptException(Exception):
```
lluiscampos: A simplest and more "pythony" way to do this is just declare it empty

```
class MyCoolException(Exception):
    pass
```

tests/tests/test_demo_artifact.py
```diff
@@ -69,7 +71,15 @@ def run_demo_script_up():
...
                     break
-            return proc
+            # Make sure that there are no errors in the script
+            if demo_script.poll() != None:
+                logging.error(
+                    "The demo script could not be brought up: error code: %d" %
+                    demo_script.returncode)
+                raise DemoScriptException(demo_script.returncode)
+            assert len(password) == 12, \
+                "The demo-script failed to initialize the environment for the tests"
```
lluiscampos: Good idea to check this here. I suggest to remove the other assert in line 72 to have the test more tidy

tests/tests/test_demo_artifact.py
```diff
@@ -79,31 +89,62 @@ def run_demo_script_up():
...
 
         logging.info("--------------------------------------------------")
         logging.info("Running test_demo_artifact_upload")
         logging.info("--------------------------------------------------")
-        self.demo_artifact_upload(run_demo_script)
-        stop_docker_compose()
+        demo_script = run_demo_script()
+        try:
+            self.demo_artifact_upload(demo_script)
+        except Exception as e:
```
lluiscampos: mmm... you want to except only `DemoScriptException`, right?

lluiscampos: Discussed in person.

Better to then use a try/finally. Better to double check in Google

tests/tests/test_demo_artifact.py
```diff
@@ -79,31 +89,57 @@ def run_demo_script_up():
...
+            raise e
+
+        try:
+            demo_script = run_demo_script()
+            self.demo_artifact_upload(demo_script)
+        except DemoScriptException as e:
+            if e.err_code != 0:
+                raise e
+            else:
+                pass
```
lluiscampos: IMO, this block will be nicer with just a simple assert e.err_code == 0

tests/tests/test_demo_artifact.py
```diff
@@ -79,31 +89,57 @@ def run_demo_script_up():
...
+        except Exception as e:
+            raise e
```
lluiscampos: This last exception is not necessary, by "default" it will raise everything that you don't except

oleorhagen:
@lluiscampos see what you think now. The commits are logically ordered.

oleorhagen:
Thank you, and yes I've seen it :/ Gonna do a couple of jenkins runs now, and monitor it.

demo
```diff
@@ -52,7 +59,7 @@ done
 # Check if the demo-Artifact has been downloaded,
 # or if there exists a newer one in storage.
 DEMO_ARTIFACT_NAME="mender-demo-artifact.mender"
-wget -q --timestamping https://dgsbl4vditpls.cloudfront.net/${DEMO_ARTIFACT_NAME}
+wget -q https://dgsbl4vditpls.cloudfront.net/${DEMO_ARTIFACT_NAME}
```
lluiscampos: Careful.

Without specifying the `-O name` a new version of the Artifact will never be downloaded

PR #631, opened 2019-07-04
Skip the test_demo_artifact tests

oleorhagen:
Changelog: Title. The test seems to cause some related test failures
in other tests. Most notably in test_state_scripts. However, the cause
is still unkown. Therefore the test is skipped for now, so that development
can continue. See MEN-2627.

### DIFF
--- a/tests/tests/test_demo_artifact.py
+++ b/tests/tests/test_demo_artifact.py
@@ -76,6 +76,7 @@ class TestDemoArtifact(MenderTesting):
     # Give the test a timeframe, as the script might run forever,
     # if something goes awry, or the script is not brought down properly.
     @pytest.mark.timeout(3000)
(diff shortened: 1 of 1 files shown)


oleorhagen:
https://tracker.mender.io/browse/MEN-2627

oleorhagen:
If this passes, I'm killing myself

kacf:
Seems to fail just as much without the tests. Maybe we should try to bisect this instead.

oleorhagen:
I guess that would be learning to debug 102 for me then :p

lluiscampos:
The tests have been passing in master for a while now, since @kacf fixed MEN-1939, and I think we have rolled out that these were the root of the instabilities.

I have rejected ticket MEN-2627 and I am closing this PR.

PR #632, opened 2019-07-04
Removed unused class variables from MenderAPI classes

oleorhagen:
auth_v2 and deployments have a declared a few unused class variables,
removing for clarity.

Changelog: None


oleorhagen:
@kjaskiewiczz I don't really see why these are needed?

PR #633, opened 2019-07-05
Fix the Mac issues in the demo script

oleorhagen:
Changelog: None


oleorhagen:
@kacf, these are the changes that are needed in order to run the demo script on a Mac, along with some minor cosmetics. The changed tests I have not pushed, as they are now skipped in master anyhow. However, https://github.com/mendersoftware/integration/pull/631 still failed with 8(!) errors in the integration tests. Maybe it is not all my fault? Or do you reckon the mere presence of the test file is enough to bring the monolith down?

demo
```diff
@@ -133,28 +137,59 @@ docker-compose \
...
+
+MENDER_SERVER_URI="https://${MENDER_TEST_GATEWAY:-localhost}"
+
+# On MacOS docker is run in a VM,
+# therefore curl calls to the container
+# IP address does not reach the containers
+# like they should. Therefore, default to localhost.
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    MENDER_SERVER_URI="https://localhost"
+fi
```
kacf: Do we actually need the non-Darwin variant here? Can't we just always connect to localhost?

demo
```diff
@@ -133,28 +137,59 @@ docker-compose \
...
 while :
 do
-    curl --silent -k -X POST -u $MENDER_SERVER_USER\
+    curl --silent -k -X POST -u ${USER}:${PASSWORD} \
          --fail\
-         --connect-timeout 120\
-         --retry 60\
-         --retry-delay 2\
-         --retry-max-time 120\
+         --connect-timeout 5\
```
kacf: Did you mean to reduce it to 5? It's quite short.

There are a couple of these.

demo
```diff
@@ -22,6 +22,10 @@ if [ "$#" -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
     exit 1
 fi
 
+# For the integration tests the name of the docker-compose project is
+# generated by pytest
+DOCKER_COMPOSE_PROJECT_NAME="${DOCKER_COMPOSE_PROJECT_NAME:-integration}"
```
kacf: I would put this and the related changes in a different commit. It's not related to Mac at all, is it?

demo
```diff
@@ -133,28 +137,59 @@ docker-compose \
...
 
+retval=$?
+if [[ $retval -ne 0 ]]; then
+    echo "Failed to start docker compose"
+    exit $retval
+fi
+
 USER='mender-demo@example.com'
-PASSWORD=$(< /dev/urandom 2>/dev/null tr -dc A-Za-z0-9 | head -c12)
+PASSWORD=$(base64 /dev/urandom | head -c 12)
```
kacf: :+1:

kacf:
> However, #631 still failed with 8(!) errors in the integration tests. Maybe it is not all my fault? Or do you reckon the mere presence of the test file is enough to bring the monolith down?

I can't say I'm sure, but right now I'm not convinced that your tests have anything to do with the current failures. I'm trying to get some bisecting going to see exactly where all this stuff broke (I'm also blocked by this in the upcoming thud update), but it might take some time.

demo
oleorhagen: Will that not send the request to all the running compose setups when the integration tests are run?

demo
oleorhagen: Yes, I did. Since it is already retried 5 times, that means 5x5 seconds is spent on each phase. But you are a much better judge of the default values than me. What would you like to see?

demo
oleorhagen: No, it was related to the tests. KK, I'm sorry, I will seperate it.

oleorhagen:
If you actually figure this out, I'll be mightily curious to see where the errors are stemming from!

demo
kacf: It shouldn't. The line [is removed](https://github.com/mendersoftware/integration/blob/master/tests/run.sh#L53) when running the tests, so only the actual demo test will use it.

demo
kacf: Ah, there is also the retries. I forgot about that. Then it's fine, I think! :-)

demo
oleorhagen: Ahh, beautiful! Paranoia started gripping me halfway through last week, so I started removing anything that could make it interfere with the other tests

kacf:
This PR unfortunately has a lot more errors than the "usual" errors I found in my other tests. Compare these two:

* [Release build with 2.0.1](https://s3-eu-west-1.amazonaws.com/mender-testing-reports/integration-reports-qemux86-64-uefi-grub/3971/report.html)
** All the failures are in the same place, in the `set_device_auth_set_status` fixture
* [This build](https://s3-eu-west-1.amazonaws.com/mender-testing-reports/integration-reports/3973/report.html)
** Failures are all over the place, many of them quite different from the first one

I'm going to run a master build as well, which may be a better comparison.

kacf:
I was wrong, it is actually passing in 2.0.x and failing all over the place in master. It's hard to say whether this PR introduces more errors or not, but we can't block this forever, so I'll just merge. If we find actual errors later they should be contained to the `test_demo` tests, so shouldn't be too hard to pinpoint.

PR #639, opened 2019-07-18
MEN-1939: Add request retrying for API calls.

kacf:
This is motivated by the extremely unpredictable issue where MongoDB
will either return I/O timeout or connection refused, both of which
seem to be caused by high load. It is a frequently occurring issue in
our parallel tests, but when testing manually the issue always seems
to go away on subsequent attempts.

Hence this fix to add a retry mechanism when doing API requests and
the server returns 500 Server Error (and only then).

Changelog: None


kacf:
Maybe, but the failures don't seem to be related to the client. Almost all the errors happen while we use the Python library to query the backends. The only influence from the client would be load.

kacf:
:-O Success! Unbelievable!

But I will not close the task yet, I will leave it to simmer a bit in the master branch, and see if it works as expected over time.

PR #641, opened 2019-07-24
make release tool aware of the enterprise repos

mchalski:
https://tracker.mender.io/browse/MEN-1611

@kacf your review is the most important here.
I think I filled in every section of the map, not sure about the `release_component` bool flag - should it be true or false?
also - will the minimal enterprise compose layer do the job?

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -56,6 +56,11 @@ class Component:
                 "docker_container": ["mender-deployments"],
                 "release_component": True,
             },
+            "deployments-enterprise": {
+                "docker_image": ["deployments-enterprise"],
+                "docker_container": ["mender-deployments"],
+                "release_component": False,
+            },
             "deviceadm": {
                 "docker_image": ["deviceadm"],
                 "docker_container": ["mender-device-adm"],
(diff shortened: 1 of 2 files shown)


extra/release_tool.py
```diff
@@ -256,8 +266,8 @@ class Component:
                 "release_component": False,
             },
             "mender-useradm": {
-                "git": ["useradm"],
-                "docker_image": ["useradm"],
+                "git": ["useradm", "useradm-enterprise"],
+                "docker_image": ["useradm", "useradm-enterprise"],
                 "release_component": True,
```
merlin-northern: why here under mender-useradm key you put useradm-enterprise?

mchalski: I have to defer to @kacf here.
this was just a semi-intelligent copy-paste based on other components.

merlin-northern: just as mine yesterday :) https://github.com/mendersoftware/integration/pull/640

kacf: It's because the `useradm-enterprise` image is a *replacement* for the `mender-useradm` container, not an extra container.

PR #645, opened 2019-07-26
enterprise layer

mchalski:
https://tracker.mender.io/browse/MEN-2656

must be tested with: https://github.com/mendersoftware/mender-qa/pull/277

build: https://mender-jenkins.mender.io/job/mender-builder/label=googlecloud%20&&%20mender_servers/4098/console


kacf:
Take a look at the `KeyError: 'email-sender'` [here](https://mender-jenkins.mender.io/job/mender-builder/4096/label=googlecloud%20&&%20mender_servers/console). What confuses me is that in the release-tool `COMPONENT_MAPS`, email-sender is marked as belonging to conductor, yet it is in the `docker-compose.mt.yml` and `docker-compose.tenant.yml` files you just removed. So where does it belong?

mchalski:
@kacf yeah, I can see that.

it belongs to conductor as far as github goes. it lives inthe same repo.

it has a separate docker image though, and is a separate service.

that it was in both files was imo redundant, but the reasoning was:
- 'tenant' provided the service def
- 'mt' - just extra config

it wasn't split into those files on purpose - this was an abandoned refactor from way back. hence why i'm trying to collate them.

kacf:
Ok. Just make sure the `image: mendersoftware/email-sender:master` line appears in some yml file then, even if it's just `other-components.yml`. Otherwise you'll keep getting those `KeyError`s early on in the build.

mchalski:
it's right there in the enterprise layer -https://github.com/mendersoftware/integration/pull/645/files#diff-474ec45aa4ecd41f79429ccbb1baa9ccR38

i'm confused as to what we're hitting here.
(btw I checked - mender-qa changes will be important, but not this early in the build, so it's not that)

mchalski:
@kacf ok I know what this is - in a single file, I'm trying to both define a service and add some config to it - in >1 sections.

this works with multiple compose files - docker will merge service defs across files.

but since it's a single file - I'm actually overwriting each of these services - with 'no image', 'no network', no anything.

PR #647, opened 2019-07-29
tests: enterprise fixture names

mchalski:
no too much refactoring - enterprise layers are already used
in several fixtures pertaining to MT.

renamed them as 'enterprise', shortened fixture names and that's about it.

changelog: none

https://tracker.mender.io/browse/MEN-2657

### DIFF
--- a/tests/common_setup.py
+++ b/tests/common_setup.py
@@ -202,7 +202,7 @@ def running_custom_production_setup(request):
 
 
 @pytest.fixture(scope="function")
-def multitenancy_setup_without_client(request):
+def enterprise_no_client(request):
     if not conftest.run_tenant_tests:
         pytest.skip("Tenant tests disabled")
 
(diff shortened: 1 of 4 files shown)


tests/common_setup.py
```diff
@@ -202,7 +202,7 @@ def fin():
 
 
 @pytest.fixture(scope="function")
-def multitenancy_setup_without_client(request):
+def enterprise_no_client(request):
     if not conftest.run_tenant_tests:
```
lluiscampos: What about changing also the conf parameter to something like `conftest.run_enterprise_tests`?

mchalski: we'll be getting back to this soon, don't want to touch too much at once.

we'll be introducing a new env var, maybe replacing this arg, or maybe doing something clever to it.

PR #650, opened 2019-07-30
extend conductor configuration with logging configuartion file

kjaskiewiczz:
Conductor should use INFO as a default log level, but with default configuartion, it's logging DEBUGs.

Issues: https://github.com/Netflix/conductor/issues/1127, https://tracker.mender.io/browse/MEN-2644

### DIFF
--- a/conductor/server/config/log4j.properties
+++ b/conductor/server/config/log4j.properties
@@ -1,0 +1,9 @@
+# Set root logger level to DEBUG and its only appender to A1.
+log4j.rootLogger=INFO, A1
+
+# A1 is set to be a ConsoleAppender.
+log4j.appender.A1=org.apache.log4j.ConsoleAppender
+
+# A1 uses PatternLayout.
+log4j.appender.A1.layout=org.apache.log4j.PatternLayout
+log4j.appender.A1.layout.ConversionPattern=%-4r [%t] %-5p %c %x - %m%n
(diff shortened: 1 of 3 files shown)


conductor/server/config/log4j.properties
```diff
@@ -0,0 +1,9 @@
+# Set root logger level to DEBUG and its only appender to A1.
```
merlin-northern: the above comment is not true considering the below line

conductor/server/config/log4j.properties
```diff
@@ -0,0 +1,9 @@
...
+log4j.rootLogger=INFO, A1
+
+# A1 is set to be a ConsoleAppender.
+log4j.appender.A1=org.apache.log4j.ConsoleAppender
+
+# A1 uses PatternLayout.
+log4j.appender.A1.layout=org.apache.log4j.PatternLayout
+log4j.appender.A1.layout.ConversionPattern=%-4r [%t] %-5p %c %x - %m%n
```
merlin-northern: could you please add here time & date? add %d at least, it would be perfect to make the format as in other services:
[2019-07-29 20:18:22,105] [INFO    ] log message

0lmi:
Logging level is already set as INFO in mender/master brach, timestamp is also changed there to use ISO8601 format instead of ms from contactor instance start  - https://github.com/mendersoftware/conductor/blob/mender/master/server/src/main/resources/log4j.properties

kjaskiewiczz:
ehh, it looks like I'm using a different branch as a base for 2.11.0 update... (2.2.0-java-opts, not mender/master)

kjaskiewiczz:
looks like we need to update mender/master, rebuild images and run tests once again https://github.com/Netflix/conductor/pull/1225
@0lmi @merlin-northern

PR #651, opened 2019-07-30
mark MT tests as -Enterprise for test suite selection

mchalski:
https://tracker.mender.io/browse/MEN-2658

we're unifying test selection mechanism - in both integration and backend-integration, tests will be tagged as Enterprise, and we'll select based on this suffix via pytest -k.

all tests using the enterprise setup are therefore tagged as such.
some code review reminded me that we already have a test selection mechanism - the $SPECIFIC_INTEGRATION_TEST env var accepted by run.sh. so basically it seems we're all set for test selection.

minor remarks:
- the name of the var could be changed to $SPECIFIC_INTEGRATION_TEST***S***, but I don't want to touch it - it would ripple through jenkins config, possibly mender-qa, god knows what else. it's 'good enough'
- I refactored slightly the stuff around 'run_multitenant_tests' flag as it was confusing - see commit message

@kacf your review is important here

### DIFF
--- a/tests/common_setup.py
+++ b/tests/common_setup.py
@@ -203,8 +203,8 @@ def running_custom_production_setup(request):
 
 @pytest.fixture(scope="function")
 def enterprise_no_client(request):
-    if not conftest.run_tenant_tests:
-        pytest.skip("Tenant tests disabled")
+    if conftest.is_integration_branch:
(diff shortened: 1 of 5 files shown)


mchalski:
@lluiscampos oh absolutely, it's under 'closing backend services' epic

PR #655, opened 2019-08-07
DO NOT MERGE: conductor update tests

kjaskiewiczz:


docker-compose.enterprise.yml
```diff
@@ -3,10 +3,12 @@ services:
 
     # subsitute services with 'enterprise' versions
     mender-deployments:
+        #image: mendersoftware/deployments:master
         image: mendersoftware/deployments-enterprise:master
 
     mender-conductor:
-        image: mendersoftware/mender-conductor-enterprise:master
+        #image: mendersoftware/mender-conductor-enterprise:master
+        image: mendersoftware/mender-conductor-enterprise:2.11.0-test
```
kacf: Just FYI, I think the tag at the end is always overwritten with `pr`, which is the tag that Jenkins builds. However the build step for enterprise might not be in place yet.

PR #659, opened 2019-08-14
tests: Update to new link for mender-artifact.

kacf:
Changelog: None


merlin-northern:
@kacf  thanks

PR #662, opened 2019-08-14
rename backend tests to '-Enterprise'

mchalski:
https://tracker.mender.io/browse/MEN-2659

needs https://github.com/mendersoftware/mender-qa/pull/285


backend-tests/tests/infra/cli.py
```diff
@@ -18,8 +18,19 @@ class CliUseradm:
     def __init__(self):
         self.cid = docker.getid('mender-useradm')
 
+        # is it an open useradm, or useradm-enterprise?
```
lluiscampos: I would document that when both are available, the enterprise version is used (according to the code)

lluiscampos:
Stopping the GitLab build, as the backend integ tests are not run there anyway.

backend-tests/tests/infra/cli.py
```diff
@@ -18,8 +18,19 @@ class CliUseradm:
...
+        for path in ['/usr/bin/useradm', '/usr/bin/useradm-enterprise']:
```
kacf: I'm comfortably removed from most of the complications here, so maybe you have good reasons, but I want to point out: Is it a good idea to use different binary names? My thinking is that if useradm-enterprise is going to be a drop in replacement for useradm, then it makes sense to keep the same name. This might help us later in tests that test both of them, for example. Just a gut feeling. :-)

backend-tests/tests/infra/cli.py
```diff
@@ -18,8 +18,19 @@ class CliUseradm:
...
+                self.path=path
+            except:
+                continue
+
+        if self.path is None:
+            raise RuntimeError('no runnable binary found in mender-useradm')
+
     def create_user(self, username, password, tenant_id=''):
-        cmd = ['/usr/bin/useradm',
+        cmd = [self.path,
```
kacf: This is exactly the type of change you wouldn't need if we used the same name.

backend-tests/tests/infra/cli.py
kjaskiewiczz: IMO, it's a different service so the binary name should be different;
I know useradm-enterprise is a replacement for useradm (same as for other enterprise services except for tenantadm), but to me, it's better to know exactly which service I'm using (open or enterprise) and the binary name is a good differentiator
I'm also aware that this will complicate tests and documentation so I don't have really strong opinion here...

mchalski: both points are valid and it could be decided either way. I'd keep the -enterprise distinction for now, mostly for the purpose of time - I feel we should proceed quickly forward now.

I do admit that at some point I may do a quick wholesale refactor of both useradm-enterprise and deployments-enterprise. Especially when this gives us problems in other areas, like docs.

PR #668, opened 2019-08-17
QA-84: Support trigger builds on GitLab from release_tool.py

lluiscampos:
```
The tool, by default, will still trigger the builds on Jenkins CI, while
this behaviour is now configurable using -c cli parameter.

The implementation basically follows the same logic as for Jenkins, only
deciding in which server to trigger the build in the last stage. One
notably difference is that the default params are fetched from the API
for Jenkins but parsed from .gitlab-ci.yaml for GitLab.

Changelog: None

```

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -35,6 +35,12 @@ JENKINS_CREDS_MISSING_ERR = """Jenkins credentials not found. Possible locations
 - JENKINS_USER / JENKINS_PASSWORD environment variables
 - 'pass' password management storage."""
 
+GITLAB_SERVER = "https://gitlab.com/api/v4"
+GITLAB_JOB = "projects/Northern.tech%2FMender%2Fmender-qa"
+GITLAB_TOKEN = None
+GITLAB_CREDS_MISSING_ERR = """GitLab credentials not found. Possible locations:
+- GITLAB_TOKEN environment variable"""
+
 # What we use in commits messages when bumping versions.
 VERSION_BUMP_STRING = "Bump versions for Mender"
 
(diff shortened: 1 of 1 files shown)


extra/release_tool.py
```diff
@@ -35,13 +35,22 @@
 - JENKINS_USER / JENKINS_PASSWORD environment variables
 - 'pass' password management storage."""
 
+GITLAB_SERVER = "https://gitlab.com/api/v4"
+GITLAB_JOB = "projects/Northern.tech%2FMender%2Fmender-qa"
+GITLAB_TOKEN = None
+GITLAB_PASSWORD = None
```
lluiscampos: This is unused. I'll remove

lluiscampos:
Thank you for your review. Merging!

extra/release_tool.py
```diff
@@ -35,13 +35,21 @@
...
+- GITLAB_TOKEN environment variable"""
+
 # What we use in commits messages when bumping versions.
 VERSION_BUMP_STRING = "Bump versions for Mender"
 
 # Whether or not pushes should really happen.
 PUSH = True
 # Whether this is a dry-run.
 DRY_RUN = False
+# Wheter we are using GitLab
```
kacf: Nitpick: `s/Wheter/Whether/`

extra/release_tool.py
```diff
@@ -1317,6 +1376,44 @@ def trigger_jenkins_build(state, tag_avail):
...
+    init_gitlab_creds()
+    if not GITLAB_TOKEN:
+        raise SystemExit(GITLAB_CREDS_MISSING_ERR)
+
+    headers = {'PRIVATE-TOKEN': GITLAB_TOKEN}
+
+    # Prepare json POST data
+    # See https://docs.gitlab.com/ee/api/pipelines.html#create-a-new-pipeline
+    # mender-qa ref hardcoded to master, it will not work for PR-like refs
+    postdata = {"ref": "master", "variables": []}
```
kacf: It should be pretty easy to make a special case for `MENDER_QA_REV` and `ref` here, don't you think? I'm using PRs with mender-qa pretty often.

lluiscampos: Yep, you are right. Thanks for doing this in PR #673

PR #670, opened 2019-08-22
Python 3 compatibility fixes and cleanups

lluiscampos:
Small fixes to make the test framework (part of it, at least!) Python 3 compatible. It should work just fine with Python 2 as well.

Best reviewed commit by commit.

### DIFF
--- a/tests/MenderAPI/auth_v2.py
+++ b/tests/MenderAPI/auth_v2.py
@@ -13,8 +13,16 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+import time
+import json
+import requests
+import pytest
 
-from MenderAPI import *
+from . import logger
+from .requests_helpers import requests_retry
+
+from ..common_docker import get_mender_gateway
+from ..common_docker import get_mender_clients
 
 class DeviceAuthV2():
 
--- a/tests/MenderAPI/auth_v2.py
+++ b/tests/MenderAPI/auth_v2.py
@@ -52,26 +52,25 @@ class DeviceAuthV2():
         starttime = time.time()
         sleeptime = 5
 
+        got_devices = False
         while starttime + max_wait >= time.time():
             time.sleep(sleeptime)
             # Linear backoff
             sleeptime += 5
-            try:
-                logger.info("getting all devices from :%s" % (device_status_path))
-                devices = requests_retry().get(device_status_path, headers=self.auth.get_auth_token(), verify=False)
-                assert devices.status_code == requests.status_codes.codes.ok
-                assert len(devices.json()) == expected_devices
+            logger.info("getting all devices from: %s" % (device_status_path))
+            devices = requests_retry().get(device_status_path, headers=self.auth.get_auth_token(), verify=False)
+            if devices.status_code == requests.status_codes.codes.ok and len(devices.json()) == expected_devices:
+                got_devices = True
                 break
-            except AssertionError:
+            else:
                 if devices is not None and getattr(devices, "text"):
                     logger.info("fail to get devices (payload: %s), will try for at least %d more seconds"
                                 % (devices.text, starttime + max_wait - time.time()))
                 else:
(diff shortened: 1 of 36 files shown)


lluiscampos:
@pasinskim could you have a look? You probably have done similar changes on your prototype for Python 3...

RootVirtual:
So amazing!! Congratulations!!

lluiscampos:
The tests passed in GitLab but failed in Jenkins. In both cases is Python 2, so investigations are needed.

PR on hold.

lluiscampos:
> The tests passed in GitLab but failed in Jenkins.

Incorrect. The GitLab tests also failed (which is good news). Only that these do not report to GitHub due to issue [QA-100](https://tracker.mender.io/browse/QA-100)

lluiscampos:
Okey, this PR is finally ready for review. The tests are now running just fine in Python 2 and in addition I am able to re-use parts of the common and MenderAPI code.

See [GitLab 78717270](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/78717270) for a CI run.

@kacf @pasinskim please have a look :)

tests/common_docker.py
```diff
@@ -77,6 +77,9 @@ def docker_compose_cmd(arg_list, use_common_files=True, env=None):
                 if "up -d" in arg_list:
                     store_logs()
 
+                # Return as string (Python 2/3 compatible)
+                if isinstance(output, bytes):
+                    return output.decode("utf-8")
```
kacf: Minor, mostly discussing code style: The default `decode()` is UTF-8. Normally I'd go for being explicit, but as this tends to be used *a lot* in Python 3, I would suggest using the shortest possible variant. What do you think?

tests/MenderAPI/auth_v2.py
```diff
@@ -56,21 +56,19 @@ def get_devices_status(self, status=None, expected_devices=1):
...
                 if devices is not None and getattr(devices, "text"):
                     logger.info("fail to get devices (payload: %s), will try for at least %d more seconds"
                                 % (devices.text, starttime + max_wait - time.time()))
                 else:
                     logger.info("failed to get devices, will try for at least %d more seconds"
                                 % (starttime + max_wait - time.time()))
-                continue
-        else:
+
+        if max_wait >= time.time():
```
kacf: I think this will always be false. `max_wait` is only `60*60`, so you need to add `starttime` like in the other calculations.

tests/tests/test_decommission.py
```diff
@@ -16,13 +16,11 @@
...
-import common_docker
-from helpers import Helpers
-from common_update import common_update_procedure
-from MenderAPI import inv, auth_v2
-from mendertesting import MenderTesting
+from ..common_setup import *
+from ..helpers import Helpers
+from .common_update import common_update_procedure
+from ..MenderAPI import inv, auth_v2
+from .mendertesting import MenderTesting
```
kacf: Quite a job you've done with all of these! Kudos!

tests/common_docker.py
lluiscampos: Agreed :+1:

tests/MenderAPI/auth_v2.py
lluiscampos: Of course it is wrong. Thanks

lluiscampos: And technically speaking, it would still be wrong unless I substract the last `sleeptime`. For instance when the request succeeds in the very last iteration. So I think I'll go simple defining a flag for this.

tests/tests/test_decommission.py
lluiscampos: Mostly search and replace, though. But thanks!

lluiscampos:
Fixed the two commented issues and rebased from latest master.

lluiscampos:
Thanks, I rewrote a bit that commit message and rebased on top of master.

Merging (finally!)

PR #674, opened 2019-08-27
release_tool: Make `--select-test-suite` compatible with version branches.

kacf:
We can't use the word 'master' for detecting which branch we're on,
because this doesn't work for version branches. Instead use a regex to
match well known names.

Changelog: None


lluiscampos:
This was [discussed in the original PR](https://github.com/mendersoftware/integration/pull/669#issuecomment-523456483). Thanks for addressing it.

I am not sure how to validate the change, though.

kacf:
I tested it with several combinations by calling the function directly in `main()`.

PR #675, opened 2019-08-28
Fix ValueError thrown when retreiving Jenkins password

lluiscampos:
Python throws ValueError if "/" is not found in the server string, which
happens for Jenkins url.


extra/release_tool.py
```diff
@@ -434,7 +434,7 @@ def get_value_from_password_storage(server, key):
         if server.startswith("https://"):
             server = server[len("https://"):]
         # Remove address part.
-        if server.index("/") >= 0:
+        if "/" in server:
```
kacf: Ah, yes, difference between `find` and `index`, I always get that wrong!

PR #677, opened 2019-08-29
tests/run: fix unquoted param expansion

mchalski:
...for SPECIFIC_INTEGRATION_TEST_ARG.

otherwise, the arg will be split on whitespaces, which
is critical to avoid here.

changelog: none

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -185,4 +185,4 @@ if [[ -n $SPECIFIC_INTEGRATION_TEST ]]; then
     SPECIFIC_INTEGRATION_TEST_ARG="-k $SPECIFIC_INTEGRATION_TEST"
 fi
 
-python2 -m pytest $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT $pass_args $SPECIFIC_INTEGRATION_TEST_ARG $DEFAULT_TESTS
+python2 -m pytest $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT $pass_args "$SPECIFIC_INTEGRATION_TEST_ARG" $DEFAULT_TESTS


tests/run.sh
```diff
@@ -185,4 +185,4 @@ if [[ -n $SPECIFIC_INTEGRATION_TEST ]]; then
     SPECIFIC_INTEGRATION_TEST_ARG="-k $SPECIFIC_INTEGRATION_TEST"
 fi
 
-python2 -m pytest $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT $pass_args $SPECIFIC_INTEGRATION_TEST_ARG $DEFAULT_TESTS
+python2 -m pytest $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT $pass_args "$SPECIFIC_INTEGRATION_TEST_ARG" $DEFAULT_TESTS
```
kacf: Will this work...? `SPECIFIC_INTEGRATION_TEST_ARG` starts with argument `-k`, which must not be concatenated with the next argument (which is what the quotes will do).

What is in `SPECIFIC_INTEGRATION_TEST` exactly? Is it `not Enterprise`? Then a better approach might be:

```
f [[ -n $SPECIFIC_INTEGRATION_TEST ]]; then
    SPECIFIC_INTEGRATION_TEST_FLAG="-k"
fi

python2 -m pytest \
    $XDIST_ARGS \
    $MAX_FAIL_ARG \
    -s \
    --verbose \
    --junitxml=results.xml \
    $HTML_REPORT \
    $pass_args \
    $SPECIFIC_INTEGRATION_TEST_FLAG "$SPECIFIC_INTEGRATION_TEST" \
    $DEFAULT_TESTS
```

lluiscampos: You are right: it is not working. Please see PR #678 for an alternative solution

PR #678, opened 2019-08-30
Amend quoting fix for test/run.sh script

lluiscampos:
The previous fix wasn't working properly on GitLab nor for Jenkins.

Actually, I think that integration tests for this specific selection case (not Enterprise) it has never worked in Jenkins. The wicked situation here is that any build launched with a PR in integration repo will always resolve to run "all" tests, hence this specific case can only be tested once the code hits master.

Commit follows:

```
commit 1ebbcd48c9f0830b2b5d744c85f58b29d6f3f356
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Fri Aug 30 09:36:29 2019 +0200

Amend quoting fix for test/run.sh script

So that it can correctly evaluate any combination of tests expressions
passed through SPECIFIC_INTEGRATION_TEST. This will work both when the
caller uses single quotes in the variable or escaped double quotes.

This commit amends incomplete fix from commit 743622a

Changelog: None

```

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -182,7 +182,16 @@ if ! pip2 list|grep -e pytest-html >/dev/null 2>&1; then
 fi
 
 if [[ -n $SPECIFIC_INTEGRATION_TEST ]]; then
-    SPECIFIC_INTEGRATION_TEST_ARG="-k $SPECIFIC_INTEGRATION_TEST"
+    SPECIFIC_INTEGRATION_TEST_FLAG="-k"
 fi
 
-python2 -m pytest $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT $pass_args "$SPECIFIC_INTEGRATION_TEST_ARG" $DEFAULT_TESTS
+python2 -m pytest \
+    $XDIST_ARGS \
+    $MAX_FAIL_ARG \
+    -s \
+    --verbose \
+    --junitxml=results.xml \
+    $HTML_REPORT \
+    $pass_args \
+    $SPECIFIC_INTEGRATION_TEST_FLAG "$SPECIFIC_INTEGRATION_TEST" \
+    $DEFAULT_TESTS


tests/run.sh
```diff
@@ -185,4 +185,4 @@ if [[ -n $SPECIFIC_INTEGRATION_TEST ]]; then
     SPECIFIC_INTEGRATION_TEST_ARG="-k $SPECIFIC_INTEGRATION_TEST"
 fi
 
-python2 -m pytest $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT $pass_args "$SPECIFIC_INTEGRATION_TEST_ARG" $DEFAULT_TESTS
+eval python2 -m pytest $XDIST_ARGS $MAX_FAIL_ARG -s --verbose --junitxml=results.xml $HTML_REPORT $pass_args $SPECIFIC_INTEGRATION_TEST_ARG $DEFAULT_TESTS
```
kacf: I think this also won't work, `eval` will just make everything unquoted. Isn't the discussion about `SPECIFIC_INTEGRATION_TEST` containing spaces that must remain quoted? If so you cannot escape having two separate variables, I think, because `-k` must not be quoted together with the rest.

lluiscampos: It will work because `SPECIFIC_INTEGRATION_TEST` value is single quoted, both in yoctobuild-build script and in travis-ci.yml (this part in a PR, not in master).

This variable is defined like `SPECIFIC_INTEGRATION_TEST="'not Something'"` so `eval` here will do the trick (it works locally at least).

From run.sh point of view, your suggestion is the safest and cleanest. I'll rework in both repos then

lluiscampos:
@kacf Thanks. Please have a look now.

Long story short: we need to merge this to master to be able to correctly verify the mender-qa PR (as it needs to run with integration master).

tests/run.sh
lluiscampos: See related commit in PR: https://github.com/mendersoftware/mender-qa/pull/290/commits/140f550ffe4f030ae21b66821cd56da190f29b86

PR #679, opened 2019-08-30
[WIP]: Initial skeleton

oleorhagen:
Changelog: None


oleorhagen:
@mchalski, it is not much, but it is something I think you might call the initial skeleton. To be honest I am not completely sure what exactly should be tested.

backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,257 @@
...
+        files=(("description", (None, description)),
+               ("size", (None, str(os.path.getsize(filename)))),
+               ("artifact", (filename, open(filename, 'rb'),
+                             "application/octet-stream"))))
+    assert r.status_code == 201
+
+
+@pytest.fixture(scope='function')
+def setup_deployments_enterprise_test(clean_mongo, nr_devices=100):
+    tenant = create_tenant('acme')
```
mchalski: cool, but it makes sense to test with at least 2 tenants.

this is to make sure we're not adding/screwing up data for the 'other' tenant.

I know it looks like it can't happen, but it can - say if the automatic database selection inside the storage layer is screwed up - and points to the single default db. then, 1 tenant will have all the symptoms of a passing test. for 2 tenants it will be immediately obvious that some data changed for both of them, when it shouldn't.

backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,257 @@
...
+    assert r.status_code == 200
+    api_devs = r.json()
+    assert len(api_devs) == 100
+    return (user, tenant, utoken, api_devs)
+
+
+class TestDeploymentsEndpointEnterprise(object):
+    test_cases = [
+        # Base-case
+        ({
```
mchalski: weird formatting, can you format like `# Phase, Full batch` and onwards?

backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,257 @@
...
+        }, {
+            "name": "second-test-case",
+            "artifact_name": "deployments-phase-testing",
+            "device_count": 100
+        }),
+        # Phase, Full batch
+        (
+            {
+                "name":
+                "third-test-case",
```
mchalski: move to single line `"name": ...`? couple other examples here too.

backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,257 @@
...
+                    "device_count": 0,
+                }, {
+                    "batch_size": 60,
+                    "device_count": 0,
+                }],
+            }),
+    ]
+
+    @pytest.mark.parametrize("test_case", test_cases)
+    def test_phased_deployments_success(self, test_case,
```
mchalski: nice idea. I don't remember us ever parametrizing test cases themselves.

backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,257 @@
...
+            api.client.GATEWAY_URL + "/api/management/v1/deployments")
+        # Set the device pool in the request body
+        request_body["devices"] = [device["id"] for device in devices]
+        resp = deploymentclient.with_auth(utoken).call(
+            'POST',
+            '/deployments',
+            body=request_body,
+        )
+        assert resp.status_code == 201
+        # Store the location for the GET /deployments/{id} request
```
mchalski: you can easily remove all these code comments. save them for major GOTCHAs

backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,257 @@
...
+        deployment_id = os.path.basename(resp.headers['Location'])
+        # GET /deployments (unfiltered)
+        resp = deploymentclient.with_auth(utoken).call('GET', '/deployments')
+        assert resp.status_code == 200
+        # POST /deployments and GET /deployments should yield equal results
+        response_body_dict = resp.json()[0]
```
mchalski: I know at least for now it's just one deployment (but I'm mentioning >1 in other comments) - but never rely on the returned order moving forward. just a note.

backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,257 @@
...
+        ({
+            "name": "first-test-case",
+            "artifact_name": "deployments-phase-testing",
+            "devices": ["is set in test body"],
+        }, {
+            "name": "first-test-case",
+            "artifact_name": "deployments-phase-testing",
+            "device_count": 100
+        }),
+        # Phase, set start time
```
mchalski: remove comments, have more descriptive "name", like `"name": "one phase, with start, no batch"`, and then
`"name": "one phase, with start, with batch"`
`"name": "> one phase, first no start, last no batch"`
`"name": "> one phase, first with start, last no batch"`, etc (just examples)

backend-tests/tests/test_deployments.py
mchalski: don't get the purpose of the base case?
also, the 'devices' thing?

backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,257 @@
...
+                "deployments-phase-testing",
+                "devices": ["set in test body"],
+                "phases": [
+                    {
+                        # Start time is optional in the first stage (default to now)
+                        "batch_size": 10,
+                    },
+                    {
+                        "start_ts":
+                        str((datetime.utcnow() +
```
mchalski: use predefined datetimes instead. don't play with time unless you have to.

mchalski:
@oleorhagen left 2 more inline comment. one general comment for the test cases - need some corner cases that I think are not there. here's a list of what I'd want:

for just 1 phase:
- defined ts, defined batch  (must be 100)
- undefined ts, defined batch (must be 100)
- defined ts, undefined batch  

let's make sure all 3 versions work, because this is what we promise.

moving forward to >1 phases - the next special case is actually 2 phases. why - because we make some promises regarding 1st phase and last phase (ts and batch size processing). this is the special case where there's nothing in between, make sure it works. again, couple of cases:
- first phase defined ts, last phase defined batch   
- first phase undefined ts, last phase defined batch  
- first phase undefined ts, last phase undefined batch  

next you could slap a full blown case of 3, or 4 or 5 phases, with a mix of ts/batch sizes. or, again, extend it to:
- first phase defined ts, last phase defined batch   
- first phase undefined ts, last phase defined batch  
- first phase undefined ts, last phase undefined batch

mchalski:
added @kjaskiewiczz as a reviewer, because of course ;)

oleorhagen:
@mchalski, should be more to your liking now

backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,494 @@
...
+                    "device_count": 0,
+                }],
+            }),
+    ]
+
+    @pytest.mark.parametrize("test_case", test_cases)
+    def test_phased_deployments_success(self, test_case,
+                                        setup_deployments_enterprise_test):
+
+        user, tenant, utoken = setup_deployments_enterprise_test
```
mchalski: I can see you create another tenant in setup_..., so that's cool.

but it's not used ever, right? it should be.

the logic should look like this:
* always have two tenants (one user per each)
* parametrize the whole test with a tenant - so that it's ran twice
* in the test - do the test for *this* tenant, but also - check that the other tenant's deployment count is still 3

as a result we have tested that evt works in multitenancy, regardless of tenant name.

you'll have to return tenants + users + tokens from setup. they can be just dicts `{tenant_name, user_name, user_token} x2` and you're all set. sometimes we use structs Tenant and User from common too but that's optional.

you already know how to use `pytest.mark.parametrize` so you're set.

backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,494 @@
...
+        resp = deploymentclient.with_auth(utoken).call(
+            'POST',
+            '/deployments',
+            body=request_body,
+        )
+        assert resp.status_code == 201
+        deployment_id = os.path.basename(resp.headers['Location'])
+        resp = deploymentclient.with_auth(utoken).call('GET', '/deployments')
+        assert resp.status_code == 200
+        assert len(resp.json()) == 4
```
mchalski: just a note - it would be useful to actually compare the existing deployments to those returned - so that we're 100% sure they're unchanged.

but, I think we can cut a corner and leave it like this. it would be good to keep in mind with next test tasks.

PR #680, opened 2019-09-02
Set default CI server for builds to be GitLab.

lluiscampos:


lluiscampos:
@kacf check it out. As discussed last Friday.

PR #682, opened 2019-09-04
Change the trigger build string to reflect the actual behavior.

pasinskim:
The default behavior for triggering the build is to not start it
but the message presented to the user is indicating that it will be
started by default.

Changelog: none

Signed-off-by: Marcin Pasinski <marcin.pasinski@northern.tech>

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -1299,7 +1299,7 @@ def trigger_build(state, tag_avail):
         for param in sorted(params.keys()):
             print(fmt_str % (param, params[param]))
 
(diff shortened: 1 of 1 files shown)


pasinskim:
Probably we don't need a whole test suite to run for this one.

kacf:
Indeed!

lluiscampos:
I did this change originally.

If you ask me "I am going to do this, ok?" I would reply, "ok". I think it still needs to hint what is expected from the user...

kacf:
Haha, this is why we need Michael to design user interfaces!

I think the next change will be yours, Lluis, if you feel motivated enough to change it. :-)

PR #684, opened 2019-09-04
Backend-test: Phased rollout

alfrunes:
Tests phased rollout for a set of 100 devices.
Checks that deployments exhausting a phase are rejected and DeploymentPhase
device_counter is updated correctly.

### DIFF
--- a/backend-tests/tests/test_deployments.py
+++ b/backend-tests/tests/test_deployments.py
@@ -1,24 +1,80 @@
 import pytest
-from datetime import datetime, timedelta
-import time
 import os
 import requests
-import json
+import random
+import time
+import util.crypto
+
+from datetime import datetime, timedelta
 
 import api.client
-from api.client import ApiClient
 import api.deviceauth as deviceauth_v1
 import api.useradm as useradm
 import api.inventory as inventory
+import api.deployments as deployments
+
+from api.client import ApiClient
 from common import User, Device, Authset, Tenant, \
-        create_tenant, create_tenant_user, \
-        clean_mongo, mongo
+        create_user, create_tenant, create_tenant_user, \
+        create_authset, change_authset_status, clean_mongo, mongo
 
+def rand_id_data():
+    mac = ":".join(["{:02x}".format(random.randint(0x00, 0xFF), 'x') for i in range(6)])
+    sn = "".join(["{}".format(random.randint(0x00, 0xFF)) for i in range(6)])
 
-# Global APIClients
-deploymentclient = ApiClient(
-    api.client.GATEWAY_URL + "/api/management/v1/deployments")
-useradmm = ApiClient(useradm.URL_MGMT)
+    return {'mac': mac, 'sn': sn}
+
+def make_pending_device(utoken, tenant_token=''):
+    id_data = rand_id_data()
+
+    priv, pub = util.crypto.rsa_get_keypair()
+    new_set = create_authset(id_data, pub, priv, utoken, tenant_token=tenant_token)
+
+    dev = Device(new_set.did, new_set.id_data, utoken, tenant_token)
+
+    dev.authsets.append(new_set)
+
+    dev.status = 'pending'
+
+    return dev
+
+def make_accepted_device(utoken, devauthd, tenant_token=''):
+    dev = make_pending_device(utoken, tenant_token=tenant_token)
+    aset_id = dev.authsets[0].id
+    change_authset_status(dev.id, aset_id, 'accepted', utoken)
+
+    aset = dev.authsets[0]
+    aset.status = 'accepted'
+
+    # obtain auth token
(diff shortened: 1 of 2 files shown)


backend-tests/tests/test_deployments.py
```diff
@@ -0,0 +1,387 @@
...
+                # First phase don't need to specify `start_ts`
+                pass
+            elif "start_ts" in deployment["phases"][i]:
+                # Sleep until next phase starts
+                start_ts = datetime.strptime(
+                    deployment["phases"][i]["start_ts"],
+                    "%Y-%m-%dT%H:%M:%SZ"
+                )
+                now = datetime.utcnow()
+                wait_time = start_ts - now
```
mchalski: does this mean we're waiting for each phase to start before ever pinging 'next'?

that's 'good', and must work, but note that one of the most useful checks for phased rollout is:
to make sure that in between phases, devices will not get an update.

I'm not sure we're testing this, it appears not.

remember this is the top business value here - a user can inspect a deployment after phase 1(say 5%), and has a time window to abort the whole thing if it went sideways. so this is the primary thing to test.

so it would be good to test:
- in between phases - that nothing can get through
- while a phase is in progress - ping `GET /deployment/id` to see if the device count is incremented correctly

backend-tests/tests/test_deployments.py
```diff
@@ -502,3 +583,315 @@ def compare_phases_json(expected, response):
...
+            ]
+        }
+        self.try_phased_updates(deployment_req, devs, utoken)
+
+
+    def test_three_phased_deployments(self, setup_devices_and_management):
+        """
+        Three phases; with no start_ts in first and no batch_size in third.
+        """
+        user, tenant, utoken, devs = setup_devices_and_management(nr_devices=101)
```
mchalski: I almost said you're calling the fixture twice - once in the argument, and once here.

I can see though that you're doing a clever trick, and the fixture simply returns a func. that's clever, and I'm not sure what to do with it. so do I understand correctly that:
- by having the fixture as arg, we get clean mongo + a func we can call - but no init data yet
- this call is what actually does the init?

I assume it's a way of parametrizing a fixture. pytest does provide fixture parametrization though - it's a slightly advanced pattern with the `indirect` keyword. by advanced I mean - unintuitive and confusing.

my suggestion would be to simply get rid of the fixture, and call it just as a func. so:
- `setup_devices_and_management` remains an ordinary func - returns result, not another func
- the fixture decorator is removed
- each test accepts the `clean_mongo` fixture instead.

as a result stuff is simple. you call setup_ as usual in every test, we don't fiddle with fixture parametrization, no second-order functions fly around the codebase, each test gets a clean db as it should.

backend-tests/tests/test_deployments.py
```diff
@@ -502,3 +583,315 @@ def compare_phases_json(expected, response):
...
+
+        deployment_req = {
+            "name"    : "three-phased-deployments",
+            "artifact_name" : "deployments-phase-testing",
+            "devices" : [dev.id for dev in devs],
+            "phases"        : [
+                {
+                    "batch_size" : 13
```
mchalski: batches got ugly too, nice ;)

backend-tests/tests/test_deployments.py
mchalski: cool, I think you caught everything here:
- no device can get in before a phase starts - check
- devs from prev phase get 204 already-installed, and don't update the phase cnt - check
- phase count incrementation during the phase - check, actually device-by-device
- phase count is correct after the full phase - check
- after a phase is saturated - next devices in line can't get in - check
- deployment is 'finished' after all phases

whew, I can't think of anything more, so that's +1.

backend-tests/tests/test_deployments.py
```diff
@@ -502,3 +583,315 @@ def compare_phases_json(expected, response):
...
+            # Check phase count has not been incremented by the above requests
+            resp = api_mgmt_deploy.with_auth(
+                user_token).call(
+                    'GET',
+                    deployments.URL_DEPLOYMENT.format(id=deployment_id)
+                )
+            phase = resp.json()["phases"][i]
+            assert phase["device_count"] == 0
+
+            if batch_sizes[i] > 0:
```
mchalski: nitpick - won't this always be true? we fill up batch size with known batch size, and also - with an extra one if the last phase doesn't have it.

backend-tests/tests/test_deployments.py
alfrunes: I agree. My mind was fixed on using this setup function as a fixture that I came up with this workaround. Explicitly calling the function is much cleaner.

backend-tests/tests/test_deployments.py
alfrunes: Yes it is, but it is legal to use `batch_size` of zero? Then perhaps we should include a test case for this special case aswell?

PR #685, opened 2019-09-06
Fix excessive builds

kacf:
```
commit 242f774c0049f80b32fb7c8a97549d3ffbf34bf4
Date:   Fri Sep 6 11:04:24 2019

    Mark deployments-enterprise and useradm-enterprise as release components
    
    Changelog: None
    
commit ad34995da63e3e53be79fbbb6ec27b47201a3cc7
Date:   Fri Sep 6 11:01:20 2019

    release_tool: Get rid of yaml warnings.
    
    Changelog: None
    
commit 0a9512b0a2f421f312d8791392288464c5a9a0e9
Date:   Fri Sep 6 10:54:38 2019

    release_tool: Improve `-f` to take `release_component` into account.
    
    We only want `-f` to give results for components that have been part
    part of a release. Those that haven't can have any arbitrary version
    listed in the docker-compose files, and are not really *part* of a
    release.
    
    For this we have to do quite some restructuring: We need to move the
    COMPONENT_MAPS out of the tool and into its own data file, so that we
    can query it from other branches. Welcome component-maps.yml! We use
    this to query for whether the component was releaseable in the version
    that is being requested, with some special logic to handle cases where
    the file is not present (old branches).
    
    Changelog: None
    
```

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -53,234 +53,7 @@ DRY_RUN = False
 USE_GITLAB = False
 
 class Component:
-    # A map that lists all our git repositories, docker images, and docker
-    # container names, and how they are associated to one another. When you add
-    # something to this list, make sure to add it to all three sections: "git",
-    # "docker_image" and "docker_container". Binary tools that don't have Docker
-    # components, will only have the "git" part.
-    COMPONENT_MAPS = {
-        "git": {
-            "deployments": {
-                "docker_image": ["deployments"],
-                "docker_container": ["mender-deployments"],
-                "release_component": True,
-            },
-            "deployments-enterprise": {
-                "docker_image": ["deployments-enterprise"],
-                "docker_container": ["mender-deployments"],
-                "release_component": False,
-            },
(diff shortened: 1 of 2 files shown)


extra/release_tool.py
```diff
@@ -370,6 +155,9 @@ def yml_components(self):
             comp.type = "yml"
         return comps
 
+    def is_release_component(self):
```
lluiscampos: It works because so far it is only used once in which case we have a custom map.

But by design this method should call `_initialize_component_maps()`

kacf: Oh, nice catch! Indeed, I will fix that!

kacf: One more function was affected as well. Fixed both.

And the test runner cancelled my old build beautifully!

lluiscampos: Nice! Thanks for reporting.

kacf:
There are some failures, but all are related to either infrastructure of phased rollout. I will merge this, since we really need it in master.

PR #687, opened 2019-09-09
Add Enterprise production template and restructure a little.

kacf:
The restructuring is mainly to address a weakness in the original
template, which was that since the user was expected to copy all the
files in the template directory, it's not possible for us to provide
updates to the `run` script.

Addressed this by instead providing an OOTB production directory, with
tracked `run` script and templates named with `.template` suffix. This
should be completely compatible with existing installations except
that the user need to delete the `run` script from his repository.
This also means that we can cut the step in the production
guide to replace path names, because the files are already in the
correct directory.

Changelog: Introduced `enterprise.yml` template in production
directory to install an Enterprise backend server.

Changelog: The old `template` directory has been replaced with a
dedicated `production` directory, and templates are now provided as
single files with the `.template` suffix instead. These should be
copied to their non-`.template` location before being used. The `run`
script should no longer be copied, and if it already exists in the
`production` directory before merging this change, it should be
removed before attempting to merge or rebase.

### DIFF
--- a/production/config/enterprise.yml.template
+++ b/production/config/enterprise.yml.template
@@ -1,0 +1,16 @@
+# This is a template file for running Mender Enterprise. It builds on the
+# prod.yml template file, so everything described there applies to this template
+# as well.
+#
+# Notes:
+# - integration/docker-compose.enterprise.yml file is assumed to be included
+
+version: '2'
+services:
+
+    mender-device-auth:
+        environment:
+            # Set this to a tenant token that should be used for clients that
+            # don't supply a tenant token. If empty, clients without a tenant
+            # token will be rejected.
+            DEVICEAUTH_DEFAULT_TENANT_TOKEN: ''
--- a/production/config/prod.yml.template
+++ b/production/config/prod.yml.template
@@ -10,8 +10,6 @@
 # - certificates and key are mounted into containers using volumes
 # - minio artifacts are stored in a named volume `mender-artifacts`; volume
 #   needs to be created manually using `docker volume create mender-artifacts`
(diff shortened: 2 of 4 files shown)


kacf:
@mchalski, @kjaskiewiczz, @merlin-northern, @drewmoseley, @mirzak: Would really like some eyes on this one, especially those of you who have set up a production environment. This is a very user facing change.

This will obviously come with some changes to [the docs](https://docs.mender.io/2.1/administration/production-installation).

merlin-northern:
> @mchalski, @kjaskiewiczz, @merlin-northern, @drewmoseley, @mirzak: Would really like some eyes on this one, especially those of you who have set up a production environment. This is a very user facing change.
> 
> This will obviously come with some changes to [the docs](https://docs.mender.io/2.1/administration/production-installation).

and note that it comes together with this one https://github.com/mendersoftware/mender-docs/pull/748

drewmoseley:
I like this a lot.  I think it will make user upgrades easier too since they should actually be able to "git merge" to bring changes in rather than having to manually figure out what they are and apply them to their newly created files.

merlin-northern:
good direction in my opinion, we saw with Drew some problems with the fact that users copied the whole directory. I think you addressed some of those! My only concern is the documentation and explaining to the users the change. update in docs of course is needed, but also in [here](https://github.com/mendersoftware/sre-tools/pull/34) (still on my todo list to merge, now will include eventual changes)

kacf:
Ok, good, that's what I was aiming for! The user still cannot "git merge" changes to the yml files themselves, but these consist almost entirely of user settings anyway, so I don't think it's a big problem.

PR #688, opened 2019-09-10
MEN-2707: Integration tests for default tenant token

oleorhagen:
Changelog: None


oleorhagen:
@kjaskiewiczz Here is the initial implementation. Nothing is yet refactored. But his passes with kacf's PR: https://github.com/mendersoftware/deviceauth/pull/294. And sometimes I think the pytest fixtures decreases readability. Thoughts?

backend-tests/tests/test_devauth_v2.py
```diff
@@ -1224,3 +1224,134 @@ def compare_aset(authset, api_authset):
...
+
+        # Device must not show up in the default tenant's account
+        r = self.devauthm.with_auth(default_utoken).call('GET',
+                                            deviceauth_v2.URL_DEVICES)
+        assert r.status_code == 200
+        api_devs = r.json()
+        assert len(api_devs) == 0
+
+
+    def test_invalid_tenant_token_never_added(self, clean_mongo):
```
kacf: This was discussed heavily on the backend channel yesterday. We concluded that we will have to support invalid tokens after all, and they will be redirected to the default tenant. Can you change the test case condition?

For now it's going to fail, since I haven't implemented it yet. If you want you can add a `@pytest.mark.xfail(strict=True)` to invert the pass condition while keeping the code correct.

oleorhagen:
Do u want it retargeted now, or wait till we see the result of the build?

kacf:
You can retarget now. The commit should still be the same if you use the Edit button at the top.

backend-tests/tests/test_devauth_v2.py
```diff
@@ -1224,3 +1224,135 @@ def compare_aset(authset, api_authset):
...
+        # Create a user for the default tenant
+        default_tenant_user = create_user('root@admin.com', 'foobarbaz', default_tenant.id)
+        r = self.uc.call('POST',
+                    useradm.URL_LOGIN,
+                    auth=(default_tenant_user.name, default_tenant_user.pwd))
+        assert r.status_code == 200
+        default_utoken = r.text
+
+        # Create a device, and try to authorize
+        create_random_authset(default_utoken, default_tenant.tenant_token)
```
kjaskiewiczz: if I understand correctly you are trying to test if the device with empty tenant token will appear in the default tenant account, but you've created device with tenant token (not empty, but regular, valid one)

oleorhagen: Yes! good catch.

PR #689, opened 2019-09-11
exposed backend-running-ip to gui container to allow onboarding

mzedel:
Changelog: None

### DIFF
--- a/demo
+++ b/demo
@@ -67,8 +67,10 @@ fi
 
 if [[ "$OSTYPE" == "darwin"* ]]; then
     ARTIFACT_SIZE_BYTES=$(stat -f %z ${DEMO_ARTIFACT_NAME}) # BSD is not GNU -_-
+    export GATEWAY_IP=$(ifconfig | sed -En 's/127.0.0.1//;s/.*inet (addr:)?(([0-9]*\.){3}[0-9]*).*/\2/p' | head -1)
 else
     ARTIFACT_SIZE_BYTES=$(stat -c %s ${DEMO_ARTIFACT_NAME})
+    export GATEWAY_IP=$(ip route get 1 | awk '{print $7;exit}')
 fi
 
 # speed up first start by pulling containers in parallel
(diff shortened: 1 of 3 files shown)


demo
```diff
@@ -96,6 +96,7 @@ export INTEGRATION_VERSION=$(git describe --tags --abbrev=0)
 export MENDER_ARTIFACT_VERSION=$(extra/release_tool.py -g mender-artifact)
 export MENDER_VERSION=$(extra/release_tool.py -g mender)
 export MENDER_DEB_PACKAGE_VERSION=$MENDER_VERSION
+export GATEWAY_IP=$(ifconfig | sed -En 's/127.0.0.1//;s/.*inet (addr:)?(([0-9]*\.){3}[0-9]*).*/\2/p')
```
kacf: We should use `ip addr` here, not `ifconfig`. `ifconfig` is deprecated, and no longer present in RHEL7, nor in minimal Debian installs.

Also, use `sed -r`, not `sed -E`. The latter is BSD-only, and although GNU introduced `-r`, I believe it's compatible with both (at least FreeBSD's man page claims it is).

template/run
```diff
@@ -7,6 +7,7 @@ export INTEGRATION_VERSION=$(git describe --tags --abbrev=0)
 export MENDER_ARTIFACT_VERSION=$(../extra/release_tool.py -g mender-artifact)
 export MENDER_VERSION=$(../extra/release_tool.py -g mender)
 export MENDER_DEB_PACKAGE_VERSION=$MENDER_VERSION
+export GATEWAY_IP=$(ifconfig | sed -En 's/127.0.0.1//;s/.*inet (addr:)?(([0-9]*\.){3}[0-9]*).*/\2/p')
```
kacf: Same.

demo
mzedel: unfortunately `ip` is not available on mac same as `sed -r` so now that had to get a bit messier 🙁

PR #693, opened 2019-09-18
scaffolding for OS->ENT migration

mchalski:
https://tracker.mender.io/browse/MEN-2743

there's no 'meat' here yet, it's a POC of:
* where to stick the test
* how to organize moving through open to ent setup with a migration in between

so I piggybacked the test to plain 'integration'. the main reason is that it works on the host and routinely runs/kills multiple docker setups in its course. I'd happily put it in backend-integration, but there's no way to shoehorn it like this.

as for running os->migrate->ent, a couple class level fixtures are provided (with some being placeholders).

I'm not including a client - we might use it or we might not. with a mock client (plain api calls) we can simulate any scenarios (as well as - large numbers of devs), we can always add some tests with an actual client. 

if we can proceed with this, I wouldn't mind merging this poc - then we can see it in action alongside other tests (I'm running just this one obviously - works on my machine...).

questions:
- I struggled a bit with keeping the mongo container around - now I'm thinking that a production user will actually have a volume, so there's no need for it. would you prefer to emulate such a volume instead of keeping the container around? imo we can do it either way go either way, we're always testing just an approximation of the user's setup (god knows what's there after some time in operation)


tests/common_docker.py
```diff
@@ -101,16 +101,31 @@ def docker_compose_cmd(arg_list, use_common_files=True, env=None):
         raise Exception("failed to start docker-compose (called: %s): exit code: %d, output: %s" % (e.cmd, e.returncode, e.output))
 
 
-def stop_docker_compose():
+def stop_docker_compose(exclude=[]):
```
lluiscampos: I would split these into two separate functions. They are not sharing any code whatsoever.

tests/tests/test_os_ent_migration.py
```diff
@@ -0,0 +1,59 @@
...
+    """ Stop the running OS setup, restart as ENT. Keep mongo alive. """
+    stop_docker_compose(exclude=['mongo'])
+
+    docker_compose_cmd("-f " + COMPOSE_FILES_PATH + "/docker-compose.yml \
+                        -f " + COMPOSE_FILES_PATH + "/docker-compose.enterprise.yml \
+                        -f " + COMPOSE_FILES_PATH + "/docker-compose.testing.yml \
+                        -f " + COMPOSE_FILES_PATH + "/docker-compose.storage.minio.yml up -d --no-recreate",
+                       use_common_files=False)
+
+@pytest.mark.usefixtures("migrated_ent_setup")
```
lluiscampos: Let's mark it as skip before merging

tests/common_docker.py
kacf: +1

lluiscampos:
> I think the PR should be targeted at the MEN-2648 branch, not master, since that is where the default tenant lives ATM, which you will need.

In that case, we don't need to skip the test. I don't mind having a dummy test in a feature branch while we work on it.

tests/common_docker.py
mchalski: you mean like - `stop_docker_compose` and `stop_docker_compose_exclude`?

lluiscampos: Something like that.

My suggestions:
- `stop_docker_compose_full` + stop_docker_compose_partial`
- `stop_docker_compose_all` + stop_docker_compose_exclude`

I don't know, something like that

kacf:
I wasn't really planning on merging anything back until the whole ENT upgrade epic is done, hence the MEN-2648 branch. Among other things, this is because the tests you are working on is a big part of what validates that the whole thing works, and there are more repositores involved (at least deviceauth and tenantadm). I'd rather not put that in master until it is at least semi-stable.

kacf:
Can we close this? It's essentially a duplicate of #698.

kacf:
Superseded by #698.

PR #696, opened 2019-09-20
Backend integration test: tenantadm `create-org` CLI command

alfrunes:
I also made the `mongo_cleanup` fixture a little more graceful by deleting all documents instead of completely dropping all collections.

### DIFF
--- a/backend-tests/tests/test_tenantadm.py
+++ b/backend-tests/tests/test_tenantadm.py
@@ -1,0 +1,131 @@
+import logging
+import pytest
+import subprocess
+import time
+
+import api
+import pymongo
+import infra
+
+from common import mongo, mongo_cleanup
+from common import User, Device, Tenant, create_user, create_tenant, \
+        create_tenant_user, create_org
+
+logger = logging.getLogger(__name__)
+
+@pytest.yield_fixture(scope="function")
+def clean_mongo_tenant_migration(mongo):
+    mongo_cleanup(mongo)
+    tenant_cli = infra.cli.CliTenantadm()
+    tenant_cli.migrate()
+    yield mongo
+    mongo_cleanup(mongo)
+
+
+class TestCreateOrganizationCLIEnterprise:
+    api_mgmt_useradm = api.client.ApiClient(api.useradm.URL_MGMT)
+    logger = logger.getChild("TestCreateOrganizationCLIEnterprise")
+
+    def test_success(self, clean_mongo_tenant_migration):
+        """
+        Create a single organization and verify that the created user
+        is able to log in.
+        """
+        self.logger.info("Starting `test_success`")
+
+        out = create_org("fooCorp", "user@example.com", "pwd")
+        self.logger.debug("tenant id: %s" % out)
+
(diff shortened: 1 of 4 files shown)


backend-tests/tests/test_tenantadm.py
```diff
@@ -0,0 +1,119 @@
...
+        Create a single organization and verify that the created user
+        is able to log in.
+        """
+        self.logger.info("Starting `test_success`")
+
+        out = create_organization("fooCorp", "user@example.com", "pwd")
+        self.logger.debug("tenant id: %s" % out)
+
+        # Retry login for 5 min
+        for i in range(60 * 5):
```
kjaskiewiczz: you are retrying for 5 minutes here and for 3 minutes in other places;

alfrunes: I think 5 minutes is a little overkill, made all login attempts retry for 3 minutes.

alfrunes:
> Please rebase and retarget this PR towards the `MEN-2648` branch. Then it will run with the right branches.
> 
> The code looks good to me, but it should be ack'ed by one of the senior backend devs.

@kacf; Done.

kacf:
There are a couple of reasons why I think this failed:

* https://github.com/mendersoftware/tenantadm/pull/90 isn't merged yet
* There is a bug in integration-test-runner, which makes it choose the wrong branch from tenantadm. I will fix this shortly

You can run it manually though, but make sure you get the branches right. The best thing is to use the release tool:

```
integration/extra/release_tool.py -b MEN-2648 --pr integration/696 --pr tenantadm/90
```

Once https://github.com/mendersoftware/tenantadm/pull/90 is merged you can drop the final two arguments.

kacf:
Both of the above points have been dealt with now.

alfrunes:
Just started the build: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/84006385

kacf:
Ok, figured out why it started the wrong build. It's because we haven't merged the Enterprise feature yet, which means that tenantadm is not yet considered a part of the build, hence it receives a default parameter instead of the correct one, MEN-2648. Kind of a chicken and egg problem...

I started [a manual build](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/84139122) for now.

kacf:
Now I don't know why it fails, maybe it's related to the `mongo_cleanup` change? It passes in the MEN-2648 branch, so it's definitely something in this PR.

kacf:
Started https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/84980284

backend-tests/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
```
kacf: @alfrunes: Well well well. It's *this* change that causes the test to fail!

Can we leave it out, or do you need it?

kacf:
Passed!

kacf:
There was a typo, I resubmitted in #704.

PR #698, opened 2019-09-26
tests for OS->ENT migration

mchalski:
https://tracker.mender.io/browse/MEN-2743

the core of the test is in the last few commits - https://github.com/mendersoftware/integration/commit/cd6617d60ca7f1f5b6c71c1a986801f834764524 and up. it's the simplest 'no migration' scenario:
- have some functioning os setup
- restart as enterprise
- create organization, check old users/devices won't get in
- restart yet again with the default tenant token
- see that the new org/user are functioning, and old devices do appear in deviceauth for acceptance

a lot of the work went into factoring out some helpers form backend-tests, as `integration` is extremely frustrating to navigate. it already paid off, but will pay off more the more we expand this test (or introduce migration tests finally).

---
https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/84950652


kacf:
All the statuses with 693 in them can be ignored, it's because the same branch was used in #693. The rest is starting to look good. Only one job left.

tests/tests/test_os_ent_migration.py
```diff
@@ -0,0 +1,240 @@
...
+    ent_data = migrate_ent_setup()
+
+    # preserve the user/tenant created before restart
+    stop_docker_compose_exclude(['mender-mongo'])
+
+    docker_compose_cmd("-f " + COMPOSE_FILES_PATH + "/docker-compose.yml \
+                        -f " + COMPOSE_FILES_PATH + "/docker-compose.enterprise.yml \
+                        -f " + COMPOSE_FILES_PATH + "/docker-compose.testing.yml \
+                        -f " + COMPOSE_FILES_PATH + "/docker-compose.testing.enterprise.yml \
+                        -f " + COMPOSE_FILES_PATH + "/docker-compose.storage.minio.yml up -d --no-recreate",
```
kacf: Does `--no-recreate` matter? Isn't all persistent data in the volumes? If it does, this needs to be included in the production documentation as well.

tests/tests/test_os_ent_migration.py
```diff
@@ -0,0 +1,240 @@
...
+def initialize_os_setup():
+    """ Seed the OS setup with all operational data - users and devices. 
+        Return {"os_devs": [...], "os_users": [...]}
+    """
+    uadmm = ApiClient('https://{}/api/management/v1/useradm'.format(get_mender_gateway()))
+    dauthd = ApiClient('https://{}/api/devices/v1/authentication'.format(get_mender_gateway()))
+    dauthm = ApiClient('https://{}/api/management/v2/devauth'.format(get_mender_gateway()))
+
+    users = [create_user("foo@tenant.com", "correcthorse"),
+             create_user("bar@tenant.com", "correcthorse")]
```
kacf: staple battery....

tests/tests/test_os_ent_migration.py
```diff
@@ -0,0 +1,240 @@
...
+    def test_devs_ok(self, migrated_enterprise_setup):
+        uadmm = ApiClient('https://{}/api/management/v1/useradm'.format(get_mender_gateway()))
+        dauthd = ApiClient('https://{}/api/devices/v1/authentication'.format(get_mender_gateway()))
+        dauthm = ApiClient('https://{}/api/management/v2/devauth'.format(get_mender_gateway()))
+        depld = ApiClient('https://{}/api/devices/v1/deployments'.format(get_mender_gateway()))
+
+        # current dev tokens don't work right off the bat
+        # the deviceauth db is empty
+        for d in migrated_enterprise_setup["os_devs"]:
+            resp = depld.with_auth(d.token).call(
```
kacf: Is this supposed to work afterwards, if we retry it at the end? Maybe we should?

tests/tests/test_os_ent_migration.py
mchalski: it matters in tests only. it was just easier to preserve data via no recreate, than fiddling with env in dockerfiles.

if a volume persists in prod install - that's an equivalent.

kacf: Ok, thanks for confirming.

tests/tests/test_os_ent_migration.py
mchalski: no, old tokens are dead forever.

we could accept devices at the end of the test, and then ask for updates - that should succeed.

but I stop at making sure devices are ready for acceptance - because the (later) accept + make sure devices can get auth pattern is tested all over.

kacf: Yes, but is it tested after an upgrade? I reckon it's worth testing this here, because the consequences of this not working would be very bad (lose ability to deploy to devices after upgrading the server). And something could go wrong with the deployments database, for example.

If you still think it's not the proper place, I'm fine with that though. I'm not an expert in this area. :-)

mchalski: I'd leave it alone tbh. We have to stop somewhere, and there's a myriad of other things we could also test here

kacf: Ok!

mchalski:
the pipeline is green https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/84950652
(when built with correct dev branches at last)

kacf:
> the pipeline is green https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/84950652
(when built with correct dev branches at last)

It's built with the correct MEN-2648 branch, but it doesn't include this PR. I started this one: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/84974972

(The parameters of a job are not so easy to see in Gitlab, but they appear in the init_workspace log after a minute or two)

mchalski:
oh my goodness, right

mchalski:
@kacf @kjaskiewiczz 
the latest job unfortunately failed on integration. given the runtimes, I won't be able to fully wrap this up before my vacation. someone has to pick this up, below are some notes on what could be wrong.

the failure (TestEntMigration):
- after create-org, the new user can't log in
- happens only on gitlab
- never failed locally*

suspects:
- create_org needs some time - undetermined
- it starts an async job, and user creation is actually the last step
- there's a time.sleep after create org - maybe it needs upping?

as a last resort, I'm pushing an increased timeout as a separate commit, and running a new pipeline:
https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/85012713

kjaskiewiczz:
https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/85012713
tests are passing
@kacf can we merge this one?

kacf:
> suspects:
> 
> * create_org needs some time - undetermined
> * it starts an async job, and user creation is actually the last step
> * there's a time.sleep after create org - maybe it needs upping?
> 
> as a last resort, I'm pushing an increased timeout as a separate commit, and running a new pipeline:

I had to do the same thing for the production template test in mender-docs, so seems reasonable.

> can we merge this one?

Yep, looks good!

PR #702, opened 2019-09-30
QA-99: Add xUnit/HTML reports to backend-tests execution

lluiscampos:
It goes together with https://github.com/mendersoftware/mender-qa/pull/310

Modifying the pytest command line args could have been done in three places:
- Appending to PYTEST_ARGS in mender-qa script (around [here](https://github.com/mendersoftware/mender-qa/blob/master/.gitlab-ci.yml#L725)).
- In `backend-tests/tests/run.sh`
- In ENTRYPOINT in Dockerfile

I chose the latest one as to keep it closer to where the requirements are installed. But other suggestions are welcomed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@ tests/large_image.dat
 tests/results.xml
 tests/downloaded-tools
 
+backend-tests/results.xml
+backend-tests/report.html
+
 __pycache__
 .cache
 artifacts
--- a/backend-tests/docker/Dockerfile
+++ b/backend-tests/docker/Dockerfile
@@ -2,10 +2,12 @@ FROM ubuntu:16.04
 
 RUN apt-get -y -qq update && apt-get -qq -y install \
     python3-pip \
-    python3-pytest \
     docker.io \
     python3-crypto
 
 RUN pip3 install --quiet requests==2.19 pymongo==3.6.1
+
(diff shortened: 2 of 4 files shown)


lluiscampos:
It doesn't work correctly: we have to find a way to merge the reports of the two pytest sessions (open + enterprise tests)

lluiscampos:
> It goes together with [mendersoftware/mender-qa#310](https://github.com/mendersoftware/mender-qa/pull/310)
> 
> Modifying the pytest command line args could have been done in three places:
> 
>     * Appending to PYTEST_ARGS in mender-qa script (around [here](https://github.com/mendersoftware/mender-qa/blob/master/.gitlab-ci.yml#L725)).
> 
>     * In `backend-tests/tests/run.sh`
> 
>     * In ENTRYPOINT in Dockerfile
> 
> 
> I chose the latest one as to keep it closer to where the requirements are installed. But other suggestions are welcomed.

Okey, I ended up doing what I wanted to avoid: to spread the logic among different layers (i.e. adding the report args at top level mender-qa script). As the full test suite is called twice from the top-level script, I cannot see how to avoid this.

See https://github.com/mendersoftware/mender-qa/pull/310

Anyway, the logic should be okey now.

lluiscampos:
@kacf Thanks!

I am rerunning the Pipeline after a minor fix in the other PR (in artifact recollection path). I will merge after I verify it works correctly.

PR #707, opened 2019-10-03
Set pytest to v2.x to fix strange error in backend integ tests

lluiscampos:
There seems to be a modification in the way the Exceptions are expected
and reported in pytest from v3 onward. The correct fix is to modify the
test, but meanwhile...

The failure was introduced in d8c4c6b, when pytest was indirectly upgraded
to latest, but has been hidden due to the development of the test
happening in an isolated feature branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,9 +19,6 @@ tests/large_image.dat
 tests/results.xml
 tests/downloaded-tools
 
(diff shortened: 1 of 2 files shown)


lluiscampos:
Tests passed. Merging to get the fix into next nightly run.

PR #708, opened 2019-10-04
Backend integration tests: update to latest pytest

lluiscampos:
```
commit 6fa7d1d28b84357438b2b4684c13ca222314ce75
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Fri Oct 4 09:48:51 2019 +0200

    Set Assertion Rewriting for our common.py file
    
    Some tets, namely test_invalid_tenant_token_added_to_default_account,
    expect and verify asserts happening in common.py. This function tells
    pytest to treat the file as test code and rewrite the asserts.
    
    See https://docs.pytest.org/en/latest/writing_plugins.html#assertion-rewriting
    
    It turned out to be that this is the correct behaviour, while v2.x of
    pytest was having a bug (which we were taking advantage of).
    
    Signed-off-by: Lluis Campos <lluis.campos@northern.tech>

commit 9bbeb42b6955b1dd6a3b2585b2d6a7761aab10b9
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Fri Oct 4 09:53:04 2019 +0200

    Update to latest pytest and fix pycrypto version as well
    
    Now that the test framework is ready for latest pytest, let's update.
    
    Fix pycrypto also, so that we don't rely on Ubuntu packaging for it.
    
    Changelog: None
    
    Signed-off-by: Lluis Campos <lluis.campos@northern.tech>
```

New commit:
```
commit 87a1d56aa74d4c58826086846aeb435bf7d64e2b
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Thu Oct 17 16:58:28 2019 +0200

    Update backend integ tests image base to Python 3
    
    Arrange also the dependencies in an external file and upgrade now the
    components that were not compatible with Python 3.7 (pytest-html).
    
    Changelog: None
    
    Signed-off-by: Lluis Campos <lluis.campos@northern.tech>
```

PR #709, opened 2019-10-04
FIX added assertion rewrite to fix the pytest.raises issue

alfrunes:
changelog: none


lluiscampos:
I launched a similar PR, see #708 

Following what we discussed in last QA meeting, we want to keep the dependencies controlled from pip (that is, ideally in a requirements.txt or else in pip install) and avoid the upstream provided ones (like Ubuntu python-pytest for instance).

PR #711, opened 2019-10-07
Fix docker-compose logs call specifying the project name

lluiscampos:
Otherwise, the call will exit abruptly when the project name is not the
same as the folder name.

Changelog: Fix issue when demo script exists abruptly on user request
for logs. The issue only showed up when the folder name contained "-"
or "." characters.

### DIFF
--- a/demo
+++ b/demo
@@ -294,4 +294,4 @@ echo "Press Enter to show the logs."
 echo "Press Ctrl-C to stop the backend and quit."
 read -se
 
-docker-compose logs --follow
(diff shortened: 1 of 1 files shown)


lluiscampos:
Stopping the Pipeline. There is no automated test that would verify this change.

lluiscampos:
You are right!

I missed it because `changelog-generator (...) 2.1.0..2.2.x` shows me that change. Strange... Maybe something wrong locally? Could you double check this?

PR #712, opened 2019-10-07
Fix docker-compose logs call specifying the project name

lluiscampos:
Otherwise, the call will exit abruptly when the project name is not the
same as the folder name.

Changelog: Fix issue when demo script exists abruptly on user request
for logs. The issue only showed up when the folder name contained "-"
or "." characters.

(cherry picked from commit e2165fbb69e61c7ce1cf5776e3a658af2f035ab3)


lluiscampos:
Cherry-pick of #711 into 2.2.x

PR #717, opened 2019-10-08
MEN-2836: Fix test_demo_artifact test failures

oleorhagen:
The demo script, which was started from pytest did not have a open
stdin file-descriptor, which in turn caused the interactive read
call in the shell to never block, like it is supposed to. The fix
was giving the stdin of the subprocess a filedescriptor to block on.

Changelog: None

### DIFF
--- a/tests/tests/test_demo_artifact.py
+++ b/tests/tests/test_demo_artifact.py
@@ -55,6 +55,7 @@ class TestDemoArtifact(MenderTesting):
                     conftest.docker_compose_instance, 'up'
                 ],
                 cwd="..",
+                stdin=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 env=test_env)
             logging.info('Started the demo script')
--- a/tests/tests/test_demo_artifact.py
+++ b/tests/tests/test_demo_artifact.py
@@ -38,14 +38,23 @@ class TestDemoArtifact(MenderTesting):
     deploy = deployments.Deployments(auth, authv2)
 
     @pytest.fixture(scope="function")
-    def run_demo_script(self, request):
+    def run_demo_script(self, request, exit_cond="Login password:"):
         """Simple fixture which returns a function which runs 'demo up'.
         Afterwards the fixture brings down the docker-compose environment,
-        so that each invocation run keeps the environment clean."""
+        so that each invocation run keeps the environment clean
+
+        :param exit_cond
+
+               Is the string which needs to be present in the output of the demo
+               script in order for it to return the process handle to the test
+               asking for it. If the string does not exist, the process will block
+               indefinitely, and hence the tests employing this fixture will need
+               to have a timeout.
+        """
 
         request.addfinalizer(stop_docker_compose)
 
-        def run_demo_script_up():
+        def run_demo_script_up(exit_cond=exit_cond):
             test_env = os.environ.copy()
             test_env[
                 'DOCKER_COMPOSE_PROJECT_NAME'] = conftest.docker_compose_instance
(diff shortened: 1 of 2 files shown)


tests/tests/test_demo_artifact.py
```diff
@@ -148,5 +149,5 @@ def demo_artifact_installation(self, run_demo_script):
     def demo_up_down_up(self, run_demo_script):
         """Test that bringing the demo environment up, then down, then up succeeds"""
         self.demo_artifact_upload(run_demo_script)
-        self.demo_artifact_upload(run_demo_script)
+        self.demo_artifact_upload(run_demo_script, exit_cond="The user already exists")
```
lluiscampos: To me, this seems to change the purpose of the test, and I don't see this mentioned in the commit message.

The test documentation does not help to figure out what are we actually verifying here: "Test that bringing the demo environment up, then down, then up succeeds".

So what is this test trying to do? Is the purpose of the test to see that a rerun of the demo script no user is created but the containers are anyway up and running?

tests/tests/test_demo_artifact.py
```diff
@@ -55,14 +55,15 @@ def run_demo_script_up():
...
+                stdin=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 env=test_env)
             logging.info('Started the demo script')
             password = ""
             time.sleep(60)
             for line in iter(proc.stdout.readline, ''):
                 logging.info(line)
-                if "Login password:" in line.strip():
+                if exit_cond in line.strip():
```
lluiscampos: If the `exit_cond` is anything but "Login password:", the code block below will fail, right?

tests/tests/test_demo_artifact.py
oleorhagen: Hmm, ur right. It is a poor name. I will split it up into two commits, and rename the test.

tests/tests/test_demo_artifact.py
oleorhagen: It will block on readline, until the test times out.

lluiscampos: So you are saying that for test `demo_up_down_up` there is no really `exit_cond` because it will block in readline * before* getting here?

The code is not very descriptive, if you pass an exit_cond that will never reach...

oleorhagen: It has two different exit conditions. One for the first case. Bringing it up the first time, in which it has to provide a password. 
Then in the second case, the exit condition is the message that the user does not exist.

oleorhagen:
@lluiscampos  split the commit in two, and added some comments, see if it is easier to unserstand now

tests/tests/test_demo_artifact.py
lluiscampos: You mean that the user already exists, right?

So back to my original point: in the case of `exit_cond` being anything but "Login password": how will the following code work?
```
password = line[-13:-1]
...
```

lluiscampos: I mean, I know "how" it works, because you only check for the len, so no matter if the password is a sane one or just some random substring from "The user already exists"...

My point is that we are using this `exit_cond` for two things, as a proper exit condition and as an expectation for where to grab the password from.

oleorhagen: Ahh, yeh. ur right! Sry, did not see this. Fixing!

oleorhagen: How about now?

lluiscampos: Good! :+1:

lluiscampos:
Remember to port again the changes to the other PRs. (you can skip the tests if they passed already)

oleorhagen:
Alright, let's pray it passes. Yeh, on it :persevere:

PR #715, opened 2019-10-08
added enterprise enabling flag to enterprise composition gui container

mzedel:
Changelog: None
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

PR #720, opened 2019-10-09
Add changelog tag for previous commit 2e51ab0

lluiscampos:
We forgot to add a Changelog in PR #715


lluiscampos:
What do you think @mzedel ?

lluiscampos:
Okey. I don't have an strong opinion...

My criteria is that this is something facing the user which is broken in the beta and we have fixed for the final, hence a Changelog is appropriate.

What do you think @kacf ?

kacf:
Definitely this is a bugfix (old version requires a workaround, new one doesn't), so I think there should be a changelog.

mzedel:
Having talked to @kacf I understand the point now, so merging...

kacf:
@mzedel: Do you mind making a note about this in [the Enterprise installation guide](https://docs.mender.io/2.2/administration/production-installation#configuring-enterprise)? And create a task targeted for 2.2.0 to revert that again.

mzedel:
This should be https://github.com/mendersoftware/mender-docs/pull/778 and https://tracker.mender.io/projects/MEN/issues/MEN-2846

lluiscampos:
Thanks.

Cherry-picked to 2.2.x on #721

PR #721, opened 2019-10-10
Remove old kubernetes templates.

estenberg:
Do not longer work and can be confusing. Different templates will now be used for larger scale instead.


merlin-northern:
@0lmi just FYI the old unused and unmaintained definitions are being removed from integration

merlin-northern:
I have no problem here, @kjaskiewiczz do you use those?

0lmi:
I never looked on it and never used it, so, it's ok for me.

kacf:
Can you rebase your PR @estenberg? The failures should go away, in theory.

estenberg:
Thanks, rebased.

PR #729, opened 2019-10-21
Migrate from python-pycrypto to python-cryptography

alfrunes:
Python pycrypto uses deprecated functions not supported in python 3.8.

changelog: none

### DIFF
--- a/backend-tests/docker/Dockerfile
+++ b/backend-tests/docker/Dockerfile
@@ -2,12 +2,11 @@ FROM ubuntu:16.04
 
 RUN apt-get -y -qq update && apt-get -qq -y install \
     python3-pip \
-    docker.io \
-    python3-crypto
+    docker.io
 
-RUN pip3 install --quiet requests==2.19 pymongo==3.6.1
+RUN pip3 install --quiet requests==2.19 pymongo==3.6.1 pycrypto==2.6
 
 # NOTE: pytest-html 1.13 is the latest one compatible with Python 3.5 (Ubuntu 16.04 latest)
-RUN pip3 install "pytest<3.0" pytest-html==1.13
+RUN pip3 install --quiet pytest==5.2 pytest-html==1.13
 
 ENTRYPOINT ["bash", "/tests/run.sh"]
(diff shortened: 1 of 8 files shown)


backend-tests/docker/Dockerfile
```diff
@@ -1,13 +1,8 @@
...
-    docker.io \
-    python3-crypto
+RUN apt-get -y -qq update && apt-get -y -qq install \
+    docker.io
 
-RUN pip3 install --quiet requests==2.19 pymongo==3.6.1
-
-# NOTE: pytest-html 1.13 is the latest one compatible with Python 3.5 (Ubuntu 16.04 latest)
-RUN pip3 install "pytest<3.0" pytest-html==1.13
+RUN pip3 install requests==2.19 pymongo==3.6.1 pytest==5.2 pytest-html==2.0 cryptography==2.8
```
lluiscampos: You should get the commits from PR #708, specifically the one moving these to an external requirements file. This was agreed by the full team is the last QA meeting.

testutils/common.py
```diff
@@ -125,14 +125,14 @@ def create_authset(dauthd1, dauthm, id_data, pubkey, privkey, utoken, tenant_tok
...
-    assert r.status_code == 401
+    assert r.status_code == 401, r.text
 
     # dev must exist and have *this* aset
     api_dev = get_device_by_id_data(dauthm, id_data, utoken)
     assert api_dev is not None
 
     aset = [a for a in api_dev['auth_sets'] if testutils.util.crypto.rsa_compare_keys(a['pubkey'], pubkey)]
-    assert len(aset) == 1
+    assert len(aset) == 1, str(aset)
```
lluiscampos: I don't think these will be enough to make all tests pass. During last release, I had problems with certain tests expecting literal "assert 2 == 1". Hence my alternative suggestion for this problem which is to let pytest rewrite asserts on test files.

lluiscampos:
The commit message also needs amending: `pycrypto` is the one using deprecated functions

backend-tests/docker/Dockerfile
```diff
@@ -1,13 +1,10 @@
-FROM ubuntu:16.04
+FROM python:3.7
```
lluiscampos: Missing to update this line to 3.8

PR #731, opened 2019-10-22
Update version of requests package

lluiscampos:
The version that was being used has a reported vulnerability.


lluiscampos:
Backend integ tests passed.

PR #732, opened 2019-10-23
Fix a flag typo in integration tests

alfrunes:
The typo was previously automatically corrected by the flag package in
mender, but becomes a problem for the upgraded CLI package.

changelog: none

### DIFF
--- a/tests/tests/test_basic_integration.py
+++ b/tests/tests/test_basic_integration.py
@@ -161,7 +161,7 @@ class TestBasicIntegration(MenderTesting):
                 wait_count = 0
 
         def deployment_triggered_callback():
-            output = run("mender --check-update")
+            output = run("mender -check-update")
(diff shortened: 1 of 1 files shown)


kacf:
But I thought that upgrading the cli package was exactly to make double `--` flags work?

alfrunes:
Yeah, but `--check-update` is a command (not a flag) and should be illegal. The new CLI package discerns commands and flags - the single hyphen version is supported only through manual translation of "deprecated flags".

kacf:
Ok, I see. I was not aware that the double flags would actually work though. This is a bit unfortunate, but we will need to make sure that they keep working with the new package if they worked with the old one. Otherwise people, especially those using standalone mode, could suddenly brick their devices because `--install` is not accepted anymore.

PR #733, opened 2019-10-23
release_tool: Fix typo in Docker pushing logic.

kacf:
Since it is only used during releases it was not caught during the
original fix.

Changelog: None


kacf:
Cancelling build. This is not going to affect anything outside of manual releases.

kacf:
Cherry-picked to 2.1.x and 2.2.x as well.

PR #734, opened 2019-10-31
[WIP] Men 1940

oleorhagen:
Initial draft for the one-liner integration test runner.

This is a good time to influence the direction of this PR :)

- O1


tests/README.md
```diff
@@ -1,25 +1,239 @@
...
+        - [Example -- Running with a Custom Client](#example----running-with-a-custom-client)
+    - [Running the Tests in a VM](#running-the-tests-in-a-vm)
+        - [Vagrant](#vagrant)
+- [Legacy Stuff (TODO -- Remove)](#legacy-stuff-todo----remove)
+    - [-](#--1)
+    - [OS X](#os-x)
+    - [Known issues](#known-issues)
+    - [Tips and Tricks](#tips-and-tricks)
+
+<!-- markdown-toc end -->
```
lluiscampos: Nice!

tests/README.md
```diff
@@ -1,25 +1,239 @@
...
+
+
+-------------------------------------------------------------------------------
+
+
+## Getting Started
+
+The dependencies for the integration tests are collected and organized in
+dependency files in the `./requirements` folder, and separated into
```
lluiscampos: Actually they are in `requirements/integration`, and I don't think we need the second level.

tests/README.md
```diff
@@ -1,25 +1,239 @@
...
+
+```bash
+$ apt install -yyq $(cat requirements/integration/deb-requirements.txt)
+```
+
+### Alpine Linux
+
+```bash
+# TODO -- What is the correct command for the Alpine package manager
+$ apk add -yyq $(cat requirements/integration/apk-requirements.txt)
```
lluiscampos: I think is just `apk --update add`. Optionally, pass the `--no-cache` flag: `apk --update --no-cache add`

tests/README.md
```diff
@@ -1,25 +1,239 @@
...
+```
+
+This will install the requirements needed to run the tests.
+
+### Running the Tests
+
+Next, run the tests with the `run.sh` script.
+
+```bash
+$ ./run.sh
```
lluiscampos: We can fix independently, but there is at least one dependency missing in `run.sh`: the update module generators. To fix this we should donwload them there.

tests/README.md
```diff
@@ -1,25 +1,239 @@
...
+
+#### Example -- Running with a Custom Backend Service
+
+For the backend services there exists a `Dockerfile` in the root repository, and
+as such a new image can be built and tagged by
+
+```bash
+$ cd /path/to/<service>/
+$ docker build . -t mendersoftware/<service>:master
+```
```
lluiscampos: :clap:

tests/requirements/integration/python-requirements.txt
```diff
@@ -0,0 +1,23 @@
...
+fabric==1.14.0             ; python_version < "3.0"
+pycrypto                   ; python_version < "3.0"
+# test_full_integration
+docker-compose==1.24.0     ; python_version < "3.0"
+
+
+####################
+#  Pip3 specifics  #
+####################
+pyyaml                      ; python_version >= "3.0"
```
lluiscampos: I didn't know about this! Thanks

tests/vagrant/Vagrantfile
```diff
@@ -0,0 +1,82 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
```
lluiscampos: I cannot comment on this one.

tests/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+# Ignore Vagrant tmp files
+vagrant/.vagrant
```
lluiscampos: This should be in the root .gitignore

tests/README.md
lluiscampos: Mmm. Maybe I am wrong. Have you managed to run all the tests with just the dependencies installed from the run.sh wrapper?

tests/.gitignore
kacf: I disagree. I think it should be in the most local file. We are not expecting vagrant files elsewhere, are we?

tests/README.md
```diff
@@ -1,25 +1,239 @@
-Mender Integration testing
-===================
+# Mender Integration Testing
 
+<!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-refresh-toc -->
```
kacf: I got it, but maybe for those not involved, mention that this is using Emacs/Spacemacs?

tests/requirements/integration/deb-requirements.txt
```diff
@@ -0,0 +1,3 @@
+docker
+docker-compose
+python-pip
```
kacf: Should this be either `python2-pip` or `python3-pip`? Or both?

tests/requirements/integration/python-requirements.txt
kacf: Is this `pip` recognized syntax?

tests/README.md
```diff
@@ -1,25 +1,239 @@
...
+
+In order to avoid dependency mismanagement due to Python packages differing from
+one test environment to the other, it is recommended to use a python virtual
+environment. The "de-facto" standard is
+[virtualenv](https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/).
+
+#### Install
+
+```bash
+$ pip install virtualenv
```
kacf: I didn't know about `virtualenv`!

tests/README.md
```diff
@@ -1,25 +1,239 @@
...
+#### Example -- Running with a Custom Client
+
+For building a custom client the approach is a little different, due to the fact
+that the client comes bundles with a Yocto image. Therefore, in order to build
+and run a custom client with the integration test setup, first build a Yocto
+image, containing the custom client. Then, build a Docker image containing this
+client by
+
+```bash
+$ cd /path/to/meta-mender
```
kacf: Just to make it crystal clear, I would add the `source oe-init-build-env` as an explicit step here. It doesn't hurt to do it more than once, and might save some confusion (people like to skim read!).

tests/README.md
```diff
@@ -1,25 +1,239 @@
...
+image, and install all the *Debian* and *Python* requirements in the
+*requirements/(deb|python)requirements.txt* file.
+
+This requires that Vagrant is installed, as well as having the
+
+| Environment variable      | Reason |
+| ---                       | ---    |
+|GOPATH                     | Used for mounting your local GO tree in the VM as a synced folder |
+|DOCKER_USER                | Needed for logging in to the Mender docker registry  |
+|MENDER_DOCKER_REGISTRY_PWD |  Needed for logging in to the Mender docker registry  |
```
kacf: Maybe sync the naming for these two: `MENDER_DOCKER_REGISTRY_USER` and `MENDER_DOCKER_REGISTRY_PWD`

tests/README.md
```diff
@@ -1,25 +1,239 @@
...
 
-`Fatal error: Needed to prompt for a connection or sudo password (host: 172.18.0.6:8822), but abort-on-prompts was set to True Aborting.`
+`Fatal error: Needed to prompt for a connection or sudo password (host:
+172.18.0.6:8822), but abort-on-prompts was set to True Aborting.`
 
 These can simply be ignored.
+
+## Tips and Tricks
+
+Remove the leftover `pycache` and `pyc` files, before testing.
```
kacf: Why?

tests/.gitignore
lluiscampos: Okey, but then we should move the following ones from the root `.gitignore` to here:

```
tests/.cache/
tests/broken_update.ext4
tests/core-image-full-cmdline-vexpress-qemu-broken-network.ext4
tests/core-image-full-cmdline-vexpress-qemu.ext4
tests/core-image-full-cmdline-qemux86-64-broken-network.ext4
tests/core-image-full-cmdline-qemux86-64.ext4
tests/mender-image-full-cmdline-rofs-qemux86-64.ext4
tests/mender-image-full-cmdline-rofs-vexpress-qemu.ext4
tests/large_image.dat
tests/results.xml
tests/downloaded-tools
```

tests/README.md
oleorhagen: Damn, thought I had gotten them all :p

oleorhagen: On the second level one, I initially started out with just one level, until I read the `.gitlab-ci.yml` file. The dependencies there are -- as far as I could understand -- sort of ordered this way, where say, the acceptance tests have different dependencies, and hence I thought it natural to split it this way, and change most of them here. Then again, maybe we should just keep this for the integration tests?

tests/README.md
oleorhagen: Just locally, and I might have all the dependencies already installed. I'll do a full run on the VM to verify.

tests/requirements/integration/python-requirements.txt
oleorhagen: Yes. I'm still not sure if we should use this over splitting into two files though...

Ref: https://www.python.org/dev/peps/pep-0508/

tests/README.md
lluiscampos: Well, mind the different repos here.

It is true that `.gitlab-ci.yml` in `mender-qa` differentiate that way, but actually the requirements are coming from different repos, right? So only integration tests reqs come from `integration` repo (the one we are now reviewing in this PR), while the acceptance tests requirements come from `meta-mender` repo. Hence to make here this second level I don't see it useful.

tests/README.md
oleorhagen: I'm just gonna remove it actually (Not the TOC, but the marker)

tests/README.md
lluiscampos: Or from a simpler point of view: this `tests` folder already implies "integration tests". Other test suites (backend-tests or acceptance tests) live in different folders and repos, and only there shall they have their requirements listed (if necessary).

tests/README.md
oleorhagen: It's pretty neat!

tests/README.md
oleorhagen: Because they cache things like paths, so when running on Vagrant, you will get the wrong paths.

tests/README.md
oleorhagen: Yes, you are right! Thx. Removing.

tests/requirements/integration/deb-requirements.txt
oleorhagen: The package is actually `python-pip`, the third version, as you spotted, has the `3`. It has not errored when running in the VM still though, so where is pyyaml actually needed? I guess it is for the release-tool?

tests/README.md
oleorhagen: Perhaps add this as Vagrant specific though. Yes. That I will.

lluiscampos: FWIW, I think you can set `PYTHONDONTWRITEBYTECODE` to avoid Python creating these files. ref  https://docs.python.org/3.8/using/cmdline.html#envvar-PYTHONDONTWRITEBYTECODE

tests/requirements/integration/deb-requirements.txt
lluiscampos: Yes, it is for release_tool

tests/README.md
oleorhagen: Nice, TIL.

tests/README.md
kacf: Oh, but the comment is useful though! If we wanted to update the TOC in an automated way, I would not know how without this comment.

oleorhagen:
Feel free to have a closer look now :)

lluiscampos:
> Feel free to have a closer look now :)

@oleorhagen but is this still WIP?

oleorhagen:
No, but I didn't wanna start a new build, not sure if it would cancel the current running pipeline or not

tests/README.md
oleorhagen: I think I'm gonna leave this one out actually, as it is emacs specific (to a specific package), plus the comment is bullocks, I generate a new one every time.

.gitignore
```diff
@@ -8,17 +8,6 @@ tags
...
-tests/broken_update.ext4
-tests/core-image-full-cmdline-vexpress-qemu-broken-network.ext4
-tests/core-image-full-cmdline-vexpress-qemu.ext4
-tests/core-image-full-cmdline-qemux86-64-broken-network.ext4
-tests/core-image-full-cmdline-qemux86-64.ext4
-tests/mender-image-full-cmdline-rofs-qemux86-64.ext4
-tests/mender-image-full-cmdline-rofs-vexpress-qemu.ext4
-tests/large_image.dat
-tests/results.xml
-tests/downloaded-tools
```
lluiscampos: :+1:

tests/README.md
```diff
@@ -1,25 +1,242 @@
...
+> managing dependencies, especially with Python can be a hassle. Therefore it is
+> recommended to add a virtual Python environment to isolate the dependencies
+> needed for running the integration tests.
+
+### Isolating Python Dependencies
+
+In order to avoid dependency mismanagement due to Python packages differing from
+one test environment to the other, it is recommended to use a python virtual
+environment. The "de-facto" standard is
+[virtualenv](https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/).
```
lluiscampos: I think this should be before the `Installing Dependencies` section. If I am a new reader, I will most likely have installed already all pip stuff when I get to this point. Another reason is that below in this section we have duplicated the `pip install -r ...` command.

What do you think? We could have this first and still pointing out that it is optional and the reader might decide to go directly to dependencies installation to get python stuff natively installed.

tests/README.md
```diff
@@ -1,25 +1,242 @@
...
+#### Install
+
+```bash
+$ pip install virtualenv
+```
+
+#### Initialize the Virtual Environment
+
+```bash
+$ cd integration-dir/tests
```
lluiscampos: What is `integration-dir`? I would assume we are already here, so that we can do `cd tests`. Else you can keep it but mark it in a way that is user dependent, like `cd <integration-dir>/tests`

tests/README.md
```diff
@@ -1,25 +1,242 @@
...
+$ virtualenv -p /usr/bin/python2.7 <name-of-virtualenv-folder>
+```
+
+#### Activate the Virtual Environment
+
+```bash
+$ source <name-of-virtualenv-folder>/bin/activate
+```
+
+This now means that you have a clean python environment, and no packages you
```
lluiscampos: Python

tests/README.md
lluiscampos: Has this been verified, then?

tests/README.md
```diff
@@ -1,25 +1,242 @@
...
+Next, run the tests with the `run.sh` script.
 
-####Known issues
+```bash
+$ ./run.sh
+```
 
-Since we attempting to SSH into the virtual mender device, before the OS is up and running, you may see errors such as:
+**NOTE**: This is dependent upon having a functioning Docker environment, and being
+logged in to `registry.mender.io`.
```
lluiscampos: Mmmm. You should be able to run the Open Source tests without logging in. So I would rephrase it a bit to show that it is not a requirement

tests/README.md
```diff
@@ -1,25 +1,242 @@
...
+`docker-compose.yml` file in the root directory, under the `image:` key.
+
+#### Example -- Running with a Custom Backend Service
+
+For the backend services there exists a `Dockerfile` in the root repository, and
+as such a new image can be built and tagged by
+
+```bash
+$ cd /path/to/<service>/
+$ docker build . -t mendersoftware/<service>:master
```
lluiscampos: (related to the above). For an enterprise repo, you would need to tag it as `registry.mender.io/mendersoftware/<service>:master`

tests/README.md
lluiscampos: Following Kristian comment, I would merge the two bunch of steps into a single one:
```
cd /path/to/yocto/dir
source oe-init-build-env
bitbake ... <- The full image, I don't remember now the exact name
cd ../meta-mender
cd meta-mender-qemu/docker
./build-docker qemux86-64 -t mendersoftware/mender-client-qemu:master
```

tests/README.md
```diff
@@ -1,25 +1,242 @@
...
+
+
+## Running the Tests in a VM
+
+### Vagrant
+
+The setup provides a
+[Vagrantfile](https://www.vagrantup.com/intro/getting-started/) in order to
+provide a common ground for running the integration tests on, based on an
+Ubuntu18.04 image. This can be found in `tests/vagrant`, and provisions a VM
```
lluiscampos: Ubuntu 18.04

tests/README.md
```diff
@@ -1,25 +1,242 @@
...
+
+This requires that Vagrant is installed, as well as having the
+
+| Environment variable        | Reason                                                            |
+| :---                        | :---                                                              |
+| GOPATH                      | Used for mounting your local GO tree in the VM as a synced folder |
+| MENDER_DOCKER_REGISTRY_USER | Needed for logging in to the Mender docker registry               |
+| MENDER_DOCKER_REGISTRY_PWD  | Needed for logging in to the Mender docker registry               |
+
+set.
```
lluiscampos: The [TABLE] set -> The following variables set: [TABLE]

lluiscampos:
I had a look at the CI run also :+1:

tests/README.md
oleorhagen: In general the run-script has a fair few options, that are not really documented. Originally I was thinking about adding it to the readme, but a usage flag to run.sh would probably be better. Thoughts?

oleorhagen: And, also, it seems not all Enterprise tests are collected in the backend tests yet, which means that now `run.sh` will run (amongst others), `TestMultiTenancyEnterprise`. Which means it is required I think

tests/README.md
```diff
@@ -1,25 +1,296 @@
...
+### Debian
+
+```bash
+$ apt install -yyq $(cat requirements/deb-requirements.txt)
+```
+
+### Alpine Linux
+
+```bash
+$ apk --update add -yyq $(cat requirements/apk-requirements.txt)
```
lluiscampos: This gives me a warning: `apk: unrecognized option: y`

tests/README.md
```diff
@@ -1,25 +1,296 @@
...
+```
+
+### Python2 and Python3
+
+```bash
+$ pip2  install -r requirements/python-requirements.txt
+$ pip3 install  -r requirements/python-requirements.txt
```
lluiscampos: Something strange with the spacing here...

tests/README.md
lluiscampos: Yes it is better, but now it is inconsistent. For the open source you used the `SPECIFIC_INTEGRATION_TEST` method while for enterprise you use the `-k` method. We should align and IMO the least hacky one is the `-k`

tests/run.sh
```diff
@@ -5,9 +5,33 @@ DEFAULT_TESTS=tests/
...
+    echo "    --no-download                    Do not download the external dependencies"
+    echo "    <pytest-args>                    Passes these arguments along to pytest"
+    echo "    tests/<testfile.py>              Name the test-file to run"
+    echo
+    echo "Recognized Environment Variables:"
+    echo
+    echo "XDIST_PARALLEL_ARG                 The number of parallell jobs for pytest-xdist"
+    echo "SPECIFIC_INTEGRATION_TEST          The ability to pass <testname-regexp> to pytest -k"
+    exit 0
+}
```
lluiscampos: Nice! :+1: 

You could add that any other parameter will be passed to pytest

tests/run.sh
```diff
@@ -152,7 +176,6 @@ XDIST_ARGS="${XDIST_ARGS:--n ${XDIST_PARALLEL_ARG:-auto}}"
 MAX_FAIL_ARG="--maxfail=1"
 HTML_REPORT="--html=report.html --self-contained-html"
 UPGRADE_TEST_ARG=""
-SPECIFIC_INTEGRATION_TEST_ARG=""
```
lluiscampos: Nice. The previous one, `UPGRADE_TEST_ARG`, can also be removed

tests/run.sh
```diff
@@ -5,9 +5,33 @@ DEFAULT_TESTS=tests/
...
+}
+
 check_tests_arguments() {
     while [ -n "$1" ]; do
         case "$1" in
+            -h)
+                set +x
+                usage
+                ;;
+            --help)
```
lluiscampos: You can unify in a single case `-h|--help`

tests/README.md
oleorhagen: Yes, I agree. The reason though, is that `./run.sh -k "not Enterprise" does not work :p

So if we are to align them, then the env var should be used in both cases.

oleorhagen:
> The missing dependency in the run.sh it is not very important, no need to spend a lot of time on that.

So I should not add the `host-tools` as a test-dependency?

lluiscampos:
> > The missing dependency in the run.sh it is not very important, no need to spend a lot of time on that.
> 
> So I should not add the `host-tools` as a test-dependency?

It is only `mender-artifact` and the `directory-artifact-gen`, right? It should be quite simple to add the missing one to the run.sh. I am just saying that it was already missing, so not really high priority to fix it in this PR.

oleorhagen:
> > > The missing dependency in the run.sh it is not very important, no need to spend a lot of time on that.
> > 
> > 
> > So I should not add the `host-tools` as a test-dependency?
> 
> It is only `mender-artifact` and the `directory-artifact-gen`, right?

Yes, as far as I've been able to discover.

 It should be quite simple to add the missing one to the run.sh. I am just saying that it was already missing, so not really high priority to fix it in this PR.

Alright. I'll call the dependencies freezed now then.

tests/README.md
lluiscampos: I see. Let's use the env var then

oleorhagen:
@lluiscampos I cannot get a green pipeline for this. The integration tests fail with the usual culprits, which means that this PR (along with mender-qa) works I guess. Also some of the acceptance tests fail, but they should be unrelated.

Can we merge?

lluiscampos:
(I don't see the comment that I got by e-mail).

But I agree with merging this, and the related `mender-qa` PR, if the test errors are the ones we are seeing in master anyway.

oleorhagen:
Hi @lluiscampos. 

Sry, I deleted it. Rebased both, and re-ran one last time. Let's see how it goes this time around.

lluiscampos:
> Hi @lluiscampos.
> 
> Sry, I deleted it. Rebased both, and re-ran one last time. Let's see how it goes this time around.

:+1:

oleorhagen:
Everything passed, except one infamous integration test, which should be completely unrelated to this task. Merging along with https://github.com/mendersoftware/mender-qa/pull/323

PR #735, opened 2019-11-01
MEN-2817: Remove the Python dependency in the demo script

oleorhagen:
Remove the Python dependency in the demo script, to decrease the dependency
surface of the demo script.

Now the Mender-Artifact and Mender versions are parsed from the
docker-compose.client.yml and other-components.yml files through a simple AWK
script instead.

Changelog: Commit

### DIFF
--- a/demo
+++ b/demo
@@ -95,8 +95,10 @@ fi
 
 # Pass this value on to the GUI container as an env variable
 export INTEGRATION_VERSION=$(git describe --tags --abbrev=0)
-export MENDER_ARTIFACT_VERSION=$(extra/release_tool.py -g mender-artifact)
-export MENDER_VERSION=$(extra/release_tool.py -g mender)
+# Parse the Mender-Artifact version used from the other-components.yml file's image tag
+export MENDER_ARTIFACT_VERSION=$(awk -F':' '/mendersoftware\/mender-artifact/ {print $3}' other-components.yml)
+# Parse the mender version from docker-compose.yml mender image's tag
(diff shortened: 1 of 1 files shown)


oleorhagen:
https://tracker.mender.io/browse/MEN-2817

oleorhagen:
This will not handle the case where a version is overridden by another docker-compose file though.

demo
```diff
@@ -95,8 +95,10 @@ fi
 
 # Pass this value on to the GUI container as an env variable
 export INTEGRATION_VERSION=$(git describe --tags --abbrev=0)
-export MENDER_ARTIFACT_VERSION=$(extra/release_tool.py -g mender-artifact)
-export MENDER_VERSION=$(extra/release_tool.py -g mender)
+# Parse the Mender-Artifact version used from the other-components.yml file's image tag
+export MENDER_ARTIFACT_VERSION=$(awk -F':' '/mendersoftware\/mender-artifact/ {print $3}' other-components.yml)
+# Parse the mender version from docker-compose.yml mender image's tag
+export MENDER_VERSION=$(awk -F':' '/mendersoftware\/mender-client/ {print $3}' docker-compose.client.yml)
```
lluiscampos: Okey, I think this is equivalent (as we are not using the `-i` option) and it solves the problem.

lluiscampos:
I don't love the change, tough, as a change in the syntax of these files will break the script without us realizing in an automated manner. But as long as we keep using this wrapper on the release checklist, we'll be fine.

PR #738, opened 2019-11-12
Re-include the skipped pre-authorization tests

oleorhagen:
These tests have been skipped for eternity, due to spurious errors. These errors I have not been able to reproduce locally, as the tests have been behaving nicely all morning.

Therefore re-adding them, to see how we fare.


oleorhagen:
@lluiscampos I have run the tests four times now, and the re-included tests have passed every single time. Is this sufficient you think?

PR #739, opened 2019-11-13
[DON'T MERGE/DEBUG] backend-integration: wait for workflows instead of sleeping

mchalski:
the sleep is causing us issues. if the workflow doesn't complete, it's
a silent failure.

this is an attempt to actually monitor the status of the started
workflow. it's POC quality.

unfortunately, the create_organization workflow never completes. even
when it 'completes', and all the workflow steps have been executed, and
tests pass... the wf status is still 'running', despite very generous
timeouts (2 x 10s).

changelog: none

https://tracker.mender.io/browse/MEN-2914


tranchitella:
@mchalski can we close this PR?

PR #740, opened 2019-11-13
[DON'T MERGE/DEBUG] backend-integration: wait for docker setup before running workflows

mchalski:
_**this actually worked and the build is: https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/350512985.**_

_**do we want to merge it or do sth more fancy, like - add the extra sleep in docker setup fixtures, so that any test has it? the downside is that not all tests need it, and all will take extra time...**_

---

trying this out as a possible solution to:
https://tracker.mender.io/browse/MEN-2914

maybe conductor timing out on various services is simply because
they're not fully up yet?

changelog: none


oleorhagen:
My 2 cents are: If this passes somewhat consistently now, merge it, then we are gonna have to do some more planning around possibly adding some more fancy monitoring to our services during the tests.

PR #742, opened 2019-11-14
Infra encapsulation: backwards compatible changes

lluiscampos:
Collecting in this PR some changes I am making in the test framework that are backwards compatible. This will ease the final review.

The main change is the imports clean-up

### DIFF
--- a/tests/pytest.ini
+++ b/tests/pytest.ini
@@ -1,0 +1,4 @@
+# Disable stdout capture, as it conflicts with Fabric
+#  see https://github.com/pytest-dev/pytest/issues/1585
+[pytest]
+addopts = --capture=no
--- a/tests/tests/common_update.py
+++ b/tests/tests/common_update.py
@@ -13,15 +13,17 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
-from ..common_docker import *
+import tempfile
+import random
+
+from fabric.api import *
+import pytest
+
+from .. import conftest
 from ..common import *
 from ..helpers import Helpers
 from ..MenderAPI import auth_v2, deploy, image, logger
-import random
-from fabric.api import *
-import tempfile
 from . import artifact_lock
-import pytest
 
 
 def common_update_procedure(install_image=None,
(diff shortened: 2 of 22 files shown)


lluiscampos:
This can be reviewed already. The tests passed (once) and that is all the evidence we need, as any error in the import would have make a test to constantly fail.

FYI @oleorhagen ^

tests/pytest.ini
```diff
@@ -0,0 +1,4 @@
+# Disable stout capture, as it conflicts with Fabric
```
oleorhagen: s/stout/stdout/

lluiscampos:
> How about running this on it also? https://pypi.org/project/isort/
> Just to verify.

I didn't know about this tool. Thanks! At the moment it is a half way reorder and I will tackle it again in the refactor.

tests/tests/common_update.py
```diff
@@ -13,15 +13,17 @@
...
 
-from ..common_docker import *
+import tempfile
+import random
+
+from fabric.api import *
+import pytest
+
+from .. import conftest
 from ..common import *
```
mchalski: left on purpose?

mchalski: I mean I see the commit message, but why everything else and not this?

lluiscampos: Yes, see commit comment: https://github.com/mendersoftware/integration/pull/742/commits/9b6aba7acf92911208a0d4a7ae9fc5f3e8f22af1

Basically the common is overriding some fabric functions, so it must be this way (for now, at least).

tests/pytest.ini
lluiscampos: :eyes:

lluiscampos:
Tests passed here: [Pipeline 95988319](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/95988319). The last force-push is a typo + rebase from master.

Merging

tests/tests/common_update.py
lluiscampos: I should have been explicit in the commit message, you are right!

PR #747, opened 2019-11-20
Runsh

oleorhagen:


tests/run.sh
```diff
@@ -179,12 +169,11 @@ if [[ -n $SPECIFIC_INTEGRATION_TEST ]]; then
...
-    $SPECIFIC_INTEGRATION_TEST_FLAG "$SPECIFIC_INTEGRATION_TEST" \
-    $DEFAULT_TESTS
+        $XDIST_ARGS \
+        $MAX_FAIL_ARG \
+        -s \
+        --verbose \
+        --junitxml=results.xml \
+        $HTML_REPORT \
+        "$@" \
+        $SPECIFIC_INTEGRATION_TEST_FLAG "$SPECIFIC_INTEGRATION_TEST"
```
lluiscampos: Why this indent change?

tests/run.sh
```diff
@@ -179,12 +169,11 @@ if [[ -n $SPECIFIC_INTEGRATION_TEST ]]; then
...
-    -s \
-    --verbose \
-    --junitxml=results.xml \
-    $HTML_REPORT \
-    $pass_args \
-    $SPECIFIC_INTEGRATION_TEST_FLAG "$SPECIFIC_INTEGRATION_TEST" \
-    $DEFAULT_TESTS
+        $XDIST_ARGS \
+        $MAX_FAIL_ARG \
+        -s \
```
lluiscampos: The `-s` can be dropped now that we have  `--capture=no` in the ini file. (I should have done this myself in the other PR)

oleorhagen:
@lluiscampos Some simple additions to the `run.sh` script, in order to simplify the usage. It is separated into two commits (which could really be their own PR's). Chronologically they are:

1) Remove the `DEFAULT_TESTS` variable from the script, and replace it with the `testpaths` variable in the `pytest.ini` script.

2) Enable separating arguments passed to `run.sh` from arguments that should be passed on to `pytest`.

Ideally, if you like these lines, I might even like to remove the `SPECIFIC_INTEGRATION_TEST` variable, and move on to either, having `mender-qa` run the script with `run.sh -- -k 'some regexp'`, or move them to use `PYTEST_ADDOPTS` to pass the arguments to pytest, so that the `run.sh` script will not have to handle this logic.

tests/run.sh
```diff
@@ -179,12 +169,11 @@ if [[ -n $SPECIFIC_INTEGRATION_TEST ]]; then
...
-    $MAX_FAIL_ARG \
-    -s \
-    --verbose \
-    --junitxml=results.xml \
-    $HTML_REPORT \
-    $pass_args \
-    $SPECIFIC_INTEGRATION_TEST_FLAG "$SPECIFIC_INTEGRATION_TEST" \
-    $DEFAULT_TESTS
+        $XDIST_ARGS \
+        $MAX_FAIL_ARG \
```
lluiscampos: More room for improvement: these two are complementary, we will have either one or the other. I suggest merging them. The ugly but backwards compatible way would be to just set `XDIST_ARGS="--maxfail=1"` when pytest is not installed. The nicer way would be with a new variable, but then you have to fix mender-qa as well.

tests/run.sh
oleorhagen: Sry, I'll fix it up.

lluiscampos:
> Ideally, if you like these lines, I might even like to remove the `SPECIFIC_INTEGRATION_TEST` variable, and move on to either, having `mender-qa` run the script with `run.sh -- -k 'some regexp'`, or move them to use `PYTEST_ADDOPTS` to pass the arguments to pytest, so that the `run.sh` script will not have to handle this logic.

It might be a nice cleanup. But right now it is not the best moment: I will be reworking quite a bit mender-qa for the next release and other colleagues are debugging integ tests. Let's do it some other time :)

tests/run.sh
oleorhagen: Alright! I like it. I think I'll create a seperate PR for it though, as then I am able to separate this PR from running it alongside `mender-qa` with a separate PR :)

oleorhagen:
Also removed the quiet option, and updated the `README` file.

oleorhagen:
@lluiscampos I'm 99% sure the failures are unrelated. I'll let you look it over before merging though.

tests/run.sh
oleorhagen: Follow-up task: https://tracker.mender.io/browse/QA-121

PR #748, opened 2019-11-20
smoke test for minio

mchalski:
minio is the most probable culprit in our integration issues.

add a smoke test to every fixture which verifies that minio is indeed up and serving.

right now a single check is performed on 2 health-check endpoints, exposed by minio itself.
an alternative would be to curl some api endpoint, but this is cleaner (if the healthcheck endpoints actually work).

TODO:
- build timeout/retry logic into smoke tests?
- port some existing smoke tests for conductor to this module (long term)

### DIFF
--- a/tests/common_setup.py
+++ b/tests/common_setup.py
@@ -22,6 +22,7 @@ from . import conftest
 from .MenderAPI import auth, auth_v2, reset_mender_api
 from .common import *
 from .common_docker import *
+import testutils.util.smoke as smoke
 
 def wait_for_containers(expected_containers, defined_in):
     for _ in range(60 * 5):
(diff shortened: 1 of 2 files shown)


mchalski:
unfortunately, all smoke tests pass, even though tests fail...

so it looks like minio is serving for some time at least.

it's TBD if we even want these checks. I say we just close this.

merlin-northern:
then just close it. good try though.

merlin-northern:
no wait. let me try to do something in here, I can use it for a second.

merlin-northern:
can we do the same but also through the storage proxy? i.e.: use https://s3 ?

lluiscampos:
@mchalski should we close this one?

tranchitella:
> @mchalski should we close this one?

Yes, I think so.

PR #750, opened 2019-11-20
docker compose: wait for minio container to be healthy on startup

kjaskiewiczz:


merlin-northern:
@kjaskiewiczz definitely good direction IMHO taking into account the timeline of what we see in logs when the tests crash, I think "container healthy" is not enough. we should check that:
* minio works, by running mc utility and adding a host, pushing something simple
* deployment service works by sending a trial GET to the /deployments endpoint
and then start the test.

oleorhagen:
Heart-warming to see a green pipe :1st_place_medal: 

Should we retry it a few times?

kjaskiewiczz:
@oleorhagen I was restarting it a couple of times; this one is the first that passed...

PR #752, opened 2019-11-21
MEN-2922: fix integration test failures

tranchitella:
Fix integration test failures:

* demo script must generate 12 chars length passwords, it occasionally generated 11 chars length passwords after commit 1cf6b17d1a06815d73d1a603d0dda02031b8d380. In the meantime, we improved the extraction of the randomly-generated password to be more resilient.
* integration tests uploading an artifact are just too fast and perform the API call to the deployments micro-service before the minio service is up and running. While these calls have a retry mechanism in place, the deployments micro-service doesn't handle the error (minio not available yet) and restarts, causing side effects that lead to 502 errors (connection refused from API gateway because deployments micro-service is not yet up and running). While effective, this fix is just a workaround: we wait 60 seconds before uploading an artifact to the deployments service to be sure minio has completed its startup process.

As the integration tests already support a retry logic, the best way to solve this issue is to introduce a `wait-for` in the deployments service, either in the micro-service itself or in the docker image entry point using the `wait-for` shell script, in order to delay the start-up of the deployments service until the S3/minio is available. In this case, the deployments service would fail (connection refused from the API gateway) without causing a restart of the docker container.

ChangeLog:title

### DIFF
--- a/demo
+++ b/demo
@@ -155,7 +155,7 @@ if [[ $retval -ne 0 ]]; then
 fi
 
 USER='mender-demo@example.com'
-PASSWORD=$(hexdump -n 6 -e '"%X"' < /dev/urandom)
+PASSWORD=$(hexdump -n 8 -e '"%X"' < /dev/urandom | cut -c1-12)
 
 
 RETRY_LIMIT=5
--- a/tests/MenderAPI/deployments.py
+++ b/tests/MenderAPI/deployments.py
@@ -44,6 +44,9 @@ class Deployments:
     def upload_image(self, filename, description="abc"):
         image_path_url = self.get_deployments_base_path() + "artifacts"
 
+        logger.info("Sleep 15 seconds to be sure minio is up and running...")
+        time.sleep(15)
+
         r = requests_retry().post(image_path_url,
                                   verify=False,
                                   headers=self.auth.get_auth_token(),
--- a/tests/common_docker.py
+++ b/tests/common_docker.py
@@ -81,6 +81,8 @@ def docker_compose_cmd(arg_list, use_common_files=True, env=None):
 
                 if "up -d" in arg_list:
                     store_logs()
+                    logger.info("sleep 15 seconds to wait for all containers to properly startup...")
+                    time.sleep(15)
 
                 # Return as string (Python 2/3 compatible)
                 if isinstance(output, bytes):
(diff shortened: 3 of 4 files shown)


demo
```diff
@@ -155,7 +155,7 @@ if [[ $retval -ne 0 ]]; then
 fi
 
 USER='mender-demo@example.com'
-PASSWORD=$(hexdump -n 6 -e '"%X"' < /dev/urandom)
+PASSWORD=$(hexdump -n 8 -e '"%X"' < /dev/urandom | cut -c1-12)
```
lluiscampos: :+1:

tests/MenderAPI/deployments.py
```diff
@@ -44,6 +44,9 @@ def get_deployments_base_path(self):
     def upload_image(self, filename, description="abc"):
         image_path_url = self.get_deployments_base_path() + "artifacts"
 
+        logger.info("Sleep 60 seconds to be sure minio is up and running...")
+        time.sleep(60)
```
lluiscampos: In the tests that use the client (most of them) we will be already waiting for a minute to get SSH before arriving here.

I suggest we move this sleep to the fixtures in `common_setup`, adding it only to the fixtures that do not wait for the client (i.e. the ones who do not call to the esoteric function `ssh_is_opened()`)

lluiscampos: Also, there are tests which will call upload_image several times, making the workaround here even more aggressive,

lluiscampos:
> Also, we are very keen on keeping a clean Git history. Please rebase your changes into the minimum commits. Up to you, my suggestion would be one commit for the demo/test_demo_artifact and one more for the sleep workarounds (this one will be a good candidate to revert once we have a proper fix in deployments).

Basically follow your very detailed and helpful PR description :)

tests/MenderAPI/deployments.py
tranchitella: I reduced the sleeping to 15 seconds (which are enough); in my tests moving it to the common_setup did not work as expected and still triggered the issue.

lluiscampos: That is very strange. But if you have tried it I will not argue. Let's merge this workaround as-is.

I still hope to be able to have a proper fix soon and revert it

lluiscampos:
The timing commit is not needed anymore after https://github.com/mendersoftware/integration/pull/750 and https://github.com/mendersoftware/deployments/pull/374

The demo script commit has been cherry-picked here: https://github.com/mendersoftware/integration/pull/755

Closing this PR.

PR #755, opened 2019-11-29
Fix integration test failures: demo script

lluiscampos:
Fix integration test failures: demo script

demo script must generate 12 chars length passwords, it occasionally generated
11 chars length passwords after commit 1cf6b17. In the meantime, we improved
the extraction of the randomly-generated password to be more resilient.

Changelog: None

### DIFF
--- a/demo
+++ b/demo
@@ -155,7 +155,7 @@ if [[ $retval -ne 0 ]]; then
 fi
 
 USER='mender-demo@example.com'
-PASSWORD=$(hexdump -n 6 -e '"%X"' < /dev/urandom)
+PASSWORD=$(hexdump -n 8 -e '"%X"' < /dev/urandom | cut -c1-12)
 
 
 RETRY_LIMIT=5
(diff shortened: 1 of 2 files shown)


lluiscampos:
Already reviewed here: https://github.com/mendersoftware/integration/pull/752

demo
```diff
@@ -155,7 +155,7 @@ if [[ $retval -ne 0 ]]; then
 fi
 
 USER='mender-demo@example.com'
-PASSWORD=$(hexdump -n 6 -e '"%X"' < /dev/urandom)
+PASSWORD=$(hexdump -n 8 -e '"%X"' < /dev/urandom | cut -c1-12)
```
oleorhagen: Free lunch to whom explains to me why the old one sometimes did 11 instead of 12 characters o_O

kacf: Maybe `\0` or `\n` is dropped by hexdump if it is the last character..? Otherwise I have no idea.

oleorhagen: Interesting theory. Will explore

PR #757, opened 2019-12-03
Add option to specify a custom release state file

lluiscampos:
Add option to specify a custom release state file

Default is same file as before, but with this option things get easier
when trying to do two releases in parallel.


lluiscampos:
Stopping automated tests.

PR #761, opened 2019-12-09
Migrate CI script to GitLab

tranchitella:
Changelog: None


tranchitella:
recheck

.gitlab-ci.yml
```diff
@@ -16,12 +16,9 @@ before_script:
...
     # Get last remaining tags, if any.
     - git fetch --tags origin
+    # Set up git identity
+    - git config --global user.name "Northern.tech" && git config --global user.email "info@northern.tech"
 
-    - git clone git://github.com/mendersoftware/mendertesting
-
-script:
-    # Check commit compliance.
-    - mendertesting/check_commits.sh
```
lluiscampos: This check we shouldn't remove.

The modern way of having it is just importing it from `mendertesting`. See [here](https://github.com/mendersoftware/mender-convert/blob/next/.gitlab-ci.yml#L7-L8) for an example.

.gitlab-ci.yml
```diff
@@ -16,12 +16,9 @@ before_script:
     - git fetch origin 'refs/heads/*:refs/heads/*'
     # Get last remaining tags, if any.
     - git fetch --tags origin
```
lluiscampos: I don't think this is necessary anymore. It was there for the check_commits which has been removed

.gitlab-ci.yml
```diff
@@ -1,12 +1,12 @@
...
-    depth: false
 
-env:
-    global:
-        TEST_BRANCH: $TRAVIS_BRANCH
+stages:
+  - test
 
-before_script:
+test:unit:
```
lluiscampos: There are no unit tests as such here, I suggest we rename the job to `test:changelog-generator` or something generic like `test:extra-tools`

tranchitella: Done, using `test:extra-tools`

.gitlab-ci.yml
tranchitella: Removed.

PR #764, opened 2019-12-11
Container Manager prototype

lluiscampos:
It started just as a PoC but it is rather a prototype at this point.

All `common_docker` logic has been moved away into a `container_manager` module. And along with it changes in the fixture setup/teardown, helper functions, MenderAPI module, and of course the tests themselves.

However, the docker-compose usage and logic has not been changed at all. The API of the new module is little expressive.

The main goals in this review are:
- Validate/discuss the encapsulation
- Bring up ideas for a cleaner API in `container_manager`

Then, depending on the discussions I will follow-up with new JIRA tickets (getting this into a feature branch) or just fix ad-hoc.

Also, there are two tests explicitly skipped at the moment: `test_demo_artifact` and `test_os_ent_migration`. I will work on them while this review is ongoing.

As usual on my PRs, **I recommend reviewing it commit by commit**.


lluiscampos:
Only one test failed, one of the cases in test_state_scripts.

However, the full run took **5 hours** instead of the usual 1h30min... I don't know why. I will investigate.

lluiscampos:
I just pushed a costemic change to force a re-build of the tests. I want to confirm that the long running times are due to the code and not just some GCP performance issue.

testutils/infra/container_manager/base.py
```diff
@@ -12,3 +15,16 @@ def teardown(self):
         """Stops the running containers
         """
         pass
+
+    def execute(self, container_id, cmd):
+        """Executes the given cmd on an specific container
+        """
+        pass
```
tranchitella: Minor note, I would replace `pass` with `raise NotImplementedError` in order to catch, well, not-implemented-errors in classes providing the actual implementation based on this abstract class.

tests/common_setup.py
```diff
@@ -36,69 +43,85 @@ def wait_for_containers(expected_containers, defined_in):
...
-    docker_compose_cmd("scale mender-client=%d" % clients)
     ssh_is_opened()
+    reset_mender_api()
 
-    auth.reset_auth_token()
-    auth_v2.accept_devices(clients)
+    def fin():
+        env.teardown()
 
+    request.addfinalizer(fin)
```
tranchitella: Here you can write it as: `request.addfinalizer(env.teardown)`

testutils/infra/container_manager/base.py
```diff
@@ -0,0 +1,14 @@
+
+class BaseContainerManagerNamespace:
+    """Base class to define a containers namespace
+    """
+
+    def setup(self):
```
tranchitella: I suppose at least setup shall be implemented by child classes, I would `raise NotImplementedError`

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -0,0 +1,136 @@
...
+# Temporary hack: assume the caller sets the variable to whatever is
+# used globally (ie.e conftest.docker_compose_instance)
+# This will eventually be the id of the namespace
+docker_compose_instance = "invalid"
+
+# Global lock to sycronize calls to docker-compose
+docker_lock = filelock.FileLock("docker_lock")
+
+def get_host_ip():
+    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
```
tranchitella: Not sure if you have strict requirements here, but I would use:
`return socket.gethostbyname(socket.gethostname())`

tests/common_setup.py
```diff
@@ -120,142 +106,106 @@ def fin():
...
+    env.setup()
 
     ssh_is_opened()
+    reset_mender_api()
     auth_v2.accept_devices(1)
 
+    def fin():
+        env.teardown()
+
+    request.addfinalizer(fin)
```
tranchitella: `request.addfinalizer(env.teardown)`

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -137,3 +172,58 @@ def setup(self):
...
+    def teardown(self):
+        self.stop_docker_compose()
+
+class DockerComposeEnterpriseSetup(DockerComposeNamespace):
+    def __init__(self, name, num_clients=0):
+        self.num_clients = num_clients
+        DockerComposeNamespace.__init__(self, name)
+    def setup(self):
+        if self.num_clients > 0:
+            raise RuntimeError("Clients not implemented for Enterprise setup")
```
tranchitella: `NotImplementedError`

tests/MenderAPI/__init__.py
```diff
@@ -10,6 +10,11 @@
 # don't complain about non-verified ssl connections
 requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
 
+# Define get funtion before the below imports to avoid circular dependency
```
tranchitella: The "correct" way to avoid circular dependency is to move the global and the getter/setter method to a dedicated file which can be imported multiple times by the consumers.

tests/MenderAPI/__init__.py
```diff
@@ -24,12 +29,14 @@
 # -- When adding something here, also add a reset method and add it below --
 
 
-def reset_mender_api():
+def reset_mender_api(manager):
```
tranchitella: `dev reset_mender_api(manager=None):`

tests/MenderAPI/__init__.py
```diff
@@ -24,12 +29,14 @@
...
     auth_v2.reset()
     deploy.reset()
     image.reset()
     inv.reset()
+    global container_manager
+    container_manager = manager
 
 
-reset_mender_api()
+reset_mender_api(None)
```
tranchitella: ... then you don't need anymore the `None` there.

tests/MenderAPI/__init__.py
lluiscampos: Thanks.

In here I tried to do the minimum changes as the full `MenderAPI` module is the next thing to refactor. The current design is very troublesome (mainly because of the global scope).

tranchitella:
@lluiscampos 

I've reviewed the PR from the point of view of a "Pythonista" and suggested a couple of stylistic changes. My general impression from the diff is that this is a huge improvement to our tests, I like the idea of the ContainerManager and its factory and I believe it would be easy (but requires a bit of time) to implement a k8s (kind based, as we do in Helm chart tests) container manager. I'm willing to help on this if/when time comes.

Well done, high quality and great job!

tests/common_setup.py
lluiscampos: Nicely spotted! Thanks. I will add it in all fixtures.

testutils/infra/container_manager/base.py
lluiscampos: Good idea. I will do it in all methods.

testutils/infra/container_manager/docker_compose_manager.py
lluiscampos: :+1:

tests/MenderAPI/__init__.py
lluiscampos: :+1:

testutils/infra/container_manager/docker_compose_manager.py
lluiscampos: I tried, but `socket.gethostbyname(socket.gethostname())` returns the loopback address, 127.0.1.1 in my machine, while the original implementation returns the LAN address. The test does not seem to work if setting the docker environment with the loopback address.

lluiscampos:
> I've reviewed the PR from the point of view of a "Pythonista" and suggested a couple of stylistic changes.

Thank you for your comments. I fixed all the stylistic suggestions.

> My general impression from the diff is that this is a huge improvement to our tests

Thanks! I am very glad we see value on doing this. The idea is to be better ready to do extensions or changes for the test framework.

lluiscampos:
The execution time is now back to normal: last test run took 90 minutes.

I cannot explain exactly why. All I could see is that pytest was failing in import time (see previous run [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/377461356) and that made the framework not to run tests in parallel.

Commit 6b50b27 (force-pushed) now comments out the import statements together with disabling the tests. These need rework anyway.

tests/tests/test_state_scripts.py
```diff
@@ -426,63 +425,63 @@
...
+    #             "ArtifactInstall_Leave_01",
+    #             "ArtifactReboot_Enter_01",
+    #             "ArtifactReboot_Leave_01",
+    #             "ArtifactCommit_Enter_89",
+    #             "ArtifactCommit_Leave_01",  # kill!
+    #             "ArtifactCommit_Leave_01",  # rerun
+    #             "ArtifactCommit_Leave_02",
+    #         ],
+    #     },
+    # ),
```
kacf: > Also, there are two tests explicitly skipped at the moment: `test_demo_artifact` and `test_os_ent_migration`. I will work on them while this review is ongoing.

Don't forget that these are also disabled.

lluiscampos: Good catch! That was committed by mistake.

testutils/infra/container_manager/docker_compose_manager.py
lluiscampos: Disregard my previous comment: it was failing locally due to a different bug in EnterpriseSetup(s). Indeed your suggested change works and it is way simpler, see commit https://github.com/mendersoftware/integration/pull/764/commits/5b1b24693f37cd264905a11f8f49ac33567f6d90

lluiscampos:
> This is quite difficult to review, since the code moving makes it difficult to understand what is new and what has just moved.

Yes, I am aware of that.

Basically all the functions from the old common_docker are now methods of DockerComposeNamespace, and the different child classes implement the setup/teardown that the fixutres used to do themselves. But the logic inside the methods, and the use that the setup/teardown do of them, have not been changed at all (at least not intentionally! ;) )

testutils/infra/container_manager/factory.py
```diff
@@ -0,0 +1,49 @@
...
+from .docker_compose_manager import DockerComposeLegacyClientSetup
+from .docker_compose_manager import DockerComposeSignedArtifactClientSetup
+from .docker_compose_manager import DockerComposeShortLivedTokenSetup
+from .docker_compose_manager import DockerComposeFailoverServerSetup
+from .docker_compose_manager import DockerComposeEnterpriseSetup
+from .docker_compose_manager import DockerComposeEnterpriseSmtpSetup
+
+class ContainerManagerFactory:
+    def getStandardSetup(self, name, num_clients=1): pass
+    def getDockerClientSetup(self, name): pass
```
oleorhagen: Question: 
It sort of stands out to me that the generic container manager has docker in its API.
Is this done on purpose?

testutils/infra/container_manager/factory.py
```diff
@@ -0,0 +1,49 @@
...
+        return DockerComposeLegacyClientSetup(name)
+    def getSignedArtifactClientSetup(self, name):
+        return DockerComposeSignedArtifactClientSetup(name)
+    def getShortLivedTokenSetup(self, name):
+        return DockerComposeShortLivedTokenSetup(name)
+    def getFailoverServerSetup(self, name):
+        return DockerComposeFailoverServerSetup(name)
+    def getEnterpriseSetup(self, name, num_clients=0):
+        return DockerComposeEnterpriseSetup(name, num_clients)
+    def getEnterpriseSmtpSetup(self, name):
```
oleorhagen: Nitpick: 
The convention is to uppercase names like `SMTP`

lluiscampos: Thanks, I will fix.

testutils/infra/container_manager/factory.py
lluiscampos: Good eyes! :eyes: I see your point and my first reaction was to agree with you in that it is wrong.

However, looking a bit closer, the point is that the container image used in this setup is called `mender-client-docker` (as per being different than `mender-client-qemu` or `mender-client-qemu-rofs`). And I think that even in the upcoming migration to a different container orchestration, the container image itself is unlikely to change. So unless we rename the image, then I think that "Docker" belongs here.

@kacf @tranchitella Do you guys think that we should rename the image itself?

kacf: Indeed, the "docker" here is a reference to the specific client image which is a pure docker image, not the bastardized VM-in-Docker solution which is the other image. We could consider renaming it maybe, but every name I can think of just makes it less clear.

lluiscampos:
As agreed, I am merging it into a feature branch `refactor-container-manager`.

I will work on follow-up PR and we will still have a last full review before getting it into production.

PR #766, opened 2019-12-16
QA-106: Cosmetics updates to remove Jenkins references

lluiscampos:
Cosmetics updates to remove Jenkins references

Leave Jenkins naming only in jenkins specific logic, and replace
everywhere else with "CI" or other generic terms.

Modified also the default value for USE_GITLAB, even though it has no
real effect as 'gitlab' is the default value for -c parameter.


lluiscampos:
Stopping integration pipeline for this PR as it will not verify the code changes.

PR #767, opened 2019-12-17
MEN-2985 ES6 support

merlin-northern:
WIP: testing image

ChangeLog:title


docker-compose.yml
```diff
@@ -105,19 +105,22 @@ services:
                     - mongo-inventory
                     - mongo-useradm
 
-    mender-conductor:
```
lluiscampos: As this PR is not really modifying mender-conductor, it would be nicer without this re-ordering of the services (so that we don't disturb Git history).

merlin-northern: > As this PR is not really modifying mender-conductor, it would be nicer without this re-ordering of the services (so that we don't disturb Git history).

this is part of the WIP, I will clean it before review. also: I think we need better health checks, like that:
```yaml
healthcheck:
  test: curl -sS http://mender-elasticsearch:9200 || exit 1
  interval: 5s
  timeout: 4s
  retries: 8
```

merlin-northern: since the health of all services is defined by the ability to answer to GET on certain endpoint. this should fix more startup issues.

merlin-northern: redis and mongo can use `nc -z -v ip port`

merlin-northern:
@kacf @lluiscampos this is work in progress for now do not pay attention. it is about integration tests failing -- not related to es6 support, I only use this PR for debugging.

tranchitella:
Superseded by https://github.com/mendersoftware/integration/pull/803.

PR #768, opened 2019-12-18
[DO NOT MERGE] 2.2.1 mongodb36

tranchitella:
This PR is functional to running CI and pipelines using mongo 3.6, it is not intended to be merged back in branch `2.2.x`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -91,7 +91,7 @@ services:
             - mender-mongo
 
     mender-mongo:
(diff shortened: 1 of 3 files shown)


tranchitella:
(Almost) all the integration tests pass with mongodb 3.6, but the typical ones which sometimes fail (e.g., state scripts and the famous too many 502 errors from deployments). We can assume everything is just fine here, so let's close the PR.

PR #769, opened 2019-12-18
QA-121: Unify the pytest environment variables passed to run.sh

oleorhagen:
Previously two environment variables were used, although the functionality was
complementary of eachother. The functionality was set, depending on whether or
not xdists was installed. If it was installed then either the XDISTS_ARGS
variables was passed to pytest. It xdist was not installed then pytest would be
given a --maxfail=1 parameter.

This commit unifies these options into one variable, the EXTRA_TEST_ARGS
variable, which is set according to the functionality described above.

Changelog: Commit

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -148,16 +148,16 @@ debugfs -w -R "rm /lib/systemd/systemd-networkd" core-image-full-cmdline-$MACHIN
 
 dd if=/dev/urandom of=broken_update.ext4 bs=10M count=5
 
-XDIST_ARGS="${XDIST_ARGS:--n ${XDIST_PARALLEL_ARG:-auto}}"
-MAX_FAIL_ARG="--maxfail=1"
+# Contains either the arguments to xdists, or '--maxfail=1', if xdist not found.
+EXTRA_TEST_ARGS=
 HTML_REPORT="--html=report.html --self-contained-html"
 
 if ! pip2 list |grep -e pytest-xdist >/dev/null 2>&1; then
-    XDIST_ARGS=""
+    EXTRA_TEST_ARGS="--maxfail=1"
     echo "WARNING: install pytest-xdist for running tests in parallel"
 else
     # run all tests when running in parallel
-    MAX_FAIL_ARG=""
+    EXTRA_TEST_ARGS="${XDIST_ARGS:--n ${XDIST_PARALLEL_ARG:-auto}}"
 fi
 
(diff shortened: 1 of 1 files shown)


tests/run.sh
```diff
@@ -148,16 +148,16 @@ debugfs -w -R "rm /lib/systemd/systemd-networkd" core-image-full-cmdline-$MACHIN
...
 HTML_REPORT="--html=report.html --self-contained-html"
 
 if ! pip2 list |grep -e pytest-xdist >/dev/null 2>&1; then
-    XDIST_ARGS=""
+    EXTRA_TEST_ARGS="--maxfail=1"
     echo "WARNING: install pytest-xdist for running tests in parallel"
 else
     # run all tests when running in parallel
-    MAX_FAIL_ARG=""
+    EXTRA_TEST_ARGS="${XDIST_ARGS:--n ${XDIST_PARALLEL_ARG:-auto}}"
```
lluiscampos: I think we could further optimize by unifying `XDIST_ARGS` and  `XDIST_PARALLEL_ARG` (i.e. get the rid of one). See mender-qa comment [here](https://github.com/mendersoftware/mender-qa/blob/master/.gitlab-ci.yml#L46).

oleorhagen:
https://github.com/mendersoftware/integration/pull/769/commits/4140965733d777cd0db76ad9631d72c2c9db16a4 Relies on: https://github.com/mendersoftware/mender-qa/pull/344

tests/run.sh
```diff
@@ -148,16 +148,16 @@ debugfs -w -R "rm /lib/systemd/systemd-networkd" core-image-full-cmdline-$MACHIN
...
 HTML_REPORT="--html=report.html --self-contained-html"
 
 if ! pip2 list |grep -e pytest-xdist >/dev/null 2>&1; then
-    XDIST_ARGS=""
+    EXTRA_TEST_ARGS="--maxfail=1"
     echo "WARNING: install pytest-xdist for running tests in parallel"
 else
     # run all tests when running in parallel
-    MAX_FAIL_ARG=""
+    EXTRA_TEST_ARGS="${XDIST_ARGS:--n ${TESTS_IN_PARALLEL:-auto}}"
```
lluiscampos: After your rework and your related pr in mender-qa, we still can remove `XDIST_ARGS` from here, right?

oleorhagen: We can. Depends on whether you want the possibility to override the xdists args in general or not?

lluiscampos: Ok, good point. Let's leave it

PR #771, opened 2019-12-19
QA-117: Finalize ContainerManager

lluiscampos:
Follow-up PR from #764. As the previous one, I recommend a commit-by-commit review.

In a nutshell: brings the `container_manager` module to production ready with considerable internal clean-ups and beautification and one interface change. The interface change is commit ______, which lets the manager(s) create individual namespaces

Also, two details to pay attention to:
- test_demo_artifact (@oleorhagen) There is a rather ugly fixture dependency of `run_demo_script` and `running_custom_production_setup`, but I couldn't figure out a better way without modifying the original intention.
- test_os_ent_migration (@mchalski) I took the simplification of using the old template file in all the enterprise setups even when the default token is not yet created. For these cases docker-compose will print a warning but the test will still pass. `WARNING: The DEFAULT_TENANT_TOKEN variable is not set. Defaulting to a blank string.`

Thanks for your time!


tests/conftest.py
```diff
@@ -41,9 +40,6 @@
 logger = log.setup_custom_logger("root", "master")
 logging.getLogger().setLevel(logging.INFO)
 
-docker_compose_instance = "mender" + str(random.randint(0, 9999999))
-
-docker_lock = filelock.FileLock("docker_lock")
```
kacf: Are you sure you want to remove the lock? This is still used in several places.

It was added a long time ago to fight spurious Docker errors when docker-compose was run in parallel. I'm not sure if it's relevant anymore, so I'm ok with removing it, but I think it should be done in a separate commit in that case.

kacf:
Nice work otherwise!

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -232,94 +231,99 @@ def get_mender_conductor(self):
...
         else:
-            self.docker_compose_cmd("up -d", use_common_files=False, file_list=self.ENTERPRISE_FILES)
-            self.wait_for_containers(15, defined_in=self.ENTERPRISE_FILES)
+            DockerComposeNamespace.__init__(self, name, self.ENTERPRISE_FILES)
+    def setup(self, recreate=True, env=None):
+        cmd = "up -d"
+        if not recreate:
+            cmd += " --no-recreate"
+        self._docker_compose_cmd(cmd, env=env)
+        self.wait_for_containers(15)
```
tranchitella: Can we factor out this magic number (15) to a global variable?
When I've added the workflows service I had to increment it and this wasn't something obvious.

tests/conftest.py
lluiscampos: The lock is still [there](https://github.com/mendersoftware/integration/pull/771/files#diff-d577a734c45f719bc351d949a423323fR27) since the previous PR. This is only a clean-up of the unused one in `conftest.py`.

There is an change, though, and that is that each xdist thread have a different lock object. But because they all use the same file, it should work exactly as before.

testutils/infra/container_manager/docker_compose_manager.py
lluiscampos: Thanks for the suggestion.

tests/common_setup.py
```diff
@@ -168,14 +168,17 @@ def setup_failover(request):
 def running_custom_production_setup(request):
     conftest.production_setup_lock.acquire()
 
-    reset_mender_api(None)
+    env = container_factory.getNoneSetup(conftest.docker_compose_instance)
```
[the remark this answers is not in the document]
lluiscampos: I agree in that the name is bad, but TBH from the ones suggested none of them resonate here...

For an start I am already poorly using the "setup" term for both the environment or namespace, and the method to bring it up. Which makes this naming discussion just weird: the factory must return a "Setup" to be consistent with the others, but it is a "Setup" with no setup method.

Sorry, I cannot explain this better...

What about `DockerComposeCustomSetup` and `getCustomSetup`?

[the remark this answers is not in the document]
oleorhagen: Yeh, I like the latter.

PR #772, opened 2019-12-21
MEN-2958: Add workflows to docker-compose

tranchitella:
We introduce a new Mender component, named workflows, that requires
a new container in the docker composition:

- mender-workflows-server: the workflow orchestration API server

Changelog: None

### DIFF
--- a/component-maps.yml
+++ b/component-maps.yml
@@ -114,6 +114,13 @@ git:
     - mender-useradm
     release_component: true
 
+  workflows:
+    docker_image:
+    - workflows
+    docker_container:
+    - mender-workflows-server
+    release_component: true
+
 docker_image:
   deployments:
     git:
@@ -227,6 +234,13 @@ docker_image:
     - mender-useradm
     release_component: true
 
+  workflows:
+    git:
+    - workflows
+    docker_container:
+    - mender-workflows-server
+    release_component: true
+
 docker_container:
   mender-deployments:
     git:
@@ -311,4 +325,11 @@ docker_container:
     docker_image:
     - useradm
     - useradm-enterprise
+    release_component: true
+
+  mender-workflows-server:
+    git:
+    - workflows
+    docker_image:
+    - workflows
     release_component: true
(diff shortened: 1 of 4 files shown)


component-maps.yml
```diff
@@ -312,3 +328,17 @@ docker_container:
...
+    docker_image:
+    - workflows
+    release_component: true
+
+  mender-workflows-worker:
+    git:
+    - workflows
+    docker_image:
+    - workflows
+    release_component: true
```
lluiscampos: Our release tool (also used for integ testing) does not support two different containers from the same image. Adding support for this will make the tool quite complex.

What was the reason to decide upon a single image?

tranchitella: @lluiscampos my fault, I wasn't aware of this limitation, fixing it.

lluiscampos: I didn't think about it in advance neither.

Giving the circumstances (most of the team in holidays), to avoid blocking the epic I think the best is to stick to what we have: workflows repo should build/publish two different Docker images and release_tool can treat them the same way we do for conductor-enterprise (for example).

I will be available Monday morning (8-9am) so we can sync then if you need help!

tranchitella: @lluiscampos TBH, what I need right now is to include the server. As we are not migrating other workflows, and the epic we are working requires its own specialised worker, I'm reworking this PR (and the Helm chart one) to include workflows server only.

I've launched a PR on mender-qa to include workflows there as well, and I hope it will unblock me here as well. Can you have a look at it?

It's perfect to sync tomorrow morning (Monday) at 8am.

component-maps.yml
```diff
@@ -312,3 +326,10 @@ docker_container:
     - useradm
     - useradm-enterprise
     release_component: true
+
+  mender-workflows-server:
+    git:
+    - workflows
+    docker_image:
+    - workflows
+    release_component: true
```
lluiscampos: Please add end line

tranchitella:
> Binary file `mender-demo-artifact.mender` added by mistake.
> The rest looks good!

Ops, you are right, force-pushed to remove it.

PR #778, opened 2020-01-09
QA-139 workflows server fixing demo command

merlin-northern:
ChangeLog:title


merlin-northern:
goes with goes with https://github.com/mendersoftware/workflows/pull/13

PR #781, opened 2020-01-14
MEN-2985 use docker.elastic.co/elasticsearch/elasticsearch:6.8.5

merlin-northern:
* es6 image updated
 * mx updated
 * timeout in backend-tests/tests/test_devauth_v2.py increased
 * order of files with -f and -f options added for logs follow in demo

ChangeLog:title

### DIFF
--- a/demo
+++ b/demo
@@ -141,8 +141,8 @@ trap exitfunc SIGTERM
 echo "Starting the Mender demo environment..."
 
 docker-compose \
-    -f docker-compose.yml \
     -f docker-compose.storage.minio.yml \
+    -f docker-compose.yml \
     -f docker-compose.demo.yml \
     -p ${DOCKER_COMPOSE_PROJECT_NAME} \
     $CLIENT \
@@ -297,4 +297,7 @@ echo "Press Enter to show the logs."
 echo "Press Ctrl-C to stop the backend and quit."
 read -se
 
-docker-compose -p ${DOCKER_COMPOSE_PROJECT_NAME} logs --follow
+docker-compose -f docker-compose.storage.minio.yml \
+    -f docker-compose.yml \
+    -f docker-compose.demo.yml \
+    -p ${DOCKER_COMPOSE_PROJECT_NAME} logs --follow
--- a/docker-compose.client.demo.yml
+++ b/docker-compose.client.demo.yml
@@ -7,3 +7,7 @@ services:
     mender-client:
         ports:
             - "85:85"
+        tmpfs:
+            - /data
+            - /tmp
+            - /mnt
--- a/docker-compose.client.yml
+++ b/docker-compose.client.yml
@@ -11,3 +11,7 @@ services:
         stdin_open: true
         tty: true
         privileged: true
+        tmpfs:
+            - /data
+            - /tmp
+            - /mnt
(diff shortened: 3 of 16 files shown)


demo
```diff
@@ -141,8 +141,8 @@ trap exitfunc SIGTERM
 echo "Starting the Mender demo environment..."
 
 docker-compose \
-    -f docker-compose.yml \
     -f docker-compose.storage.minio.yml \
+    -f docker-compose.yml \
```
lluiscampos: It is okey, but I think that the order was not a problem. What fixes the issue is your later fix (missing storage.minio the docker-compose files in the logs call).

docker-compose.client.demo.yml
```diff
@@ -7,3 +7,7 @@ services:
     mender-client:
         ports:
             - "85:85"
+        tmpfs:
+            - /data
+            - /tmp
+            - /mnt
```
lluiscampos: As little as I understand docker-compose, as this is already in `docker-compose.client.yml` we don't need to copy it here. Let's get advantage of inheritance design of the tool.

docker-compose.client.yml
```diff
@@ -11,3 +11,7 @@ services:
         stdin_open: true
         tty: true
         privileged: true
+        tmpfs:
+            - /data
+            - /tmp
+            - /mnt
```
lluiscampos: I am curious: why do we need these? Does it considerably improve the client performance?

tests/MenderAPI/authentication.py
```diff
@@ -101,12 +101,15 @@ def get_auth_token(self, create_new_user=True):
                     self.create_user(self.username, self.password)
 
             # It might take some time for create_org to propagate the new user.
-            # Retry login for a minute.
-            for i in range(60):
+            # Retry login for some time.
+            try_number=0
+            for i in range(256):
+                time.sleep(8)
```
lluiscampos: Don't you think that 34 minutes is an overkill? I know you love these numbers that I don't see they make much sense... I would say that if I can't log in in couple of minutes is enough to abort and fail the test.

tests/mender-conductor-enterprise/server/workflows/create_organization.json
```diff
@@ -0,0 +1,152 @@
+{
```
lluiscampos: Where are these workflows files coming from?

I don't think that any of these workflows files belong here in the first place, and on top of that I am very skeptical into committing "enterprise" files into an OS repository.

lluiscampos: Also, I cannot review the content, someone else should do that.

tests/prod.yml.template.patch
```diff
@@ -0,0 +1,32 @@
+--- /Users/merlin/master/integration/production/config/prod.yml.template	2020-01-14 16:45:53.000000000 +0100
++++ production/config/prod.yml.template	2020-01-16 07:46:53.000000000 +0100
```
lluiscampos: So what is the problem in leaving these volumes mounted in the production environment?

In principle, we shouldn't apply this patch and try always to test production envs as close as possible to real production envs.

tests/MenderAPI/authentication.py
lluiscampos: My point is that retrying so much just for the sake of having a green test might hide other robustness or performance issues. From the QA point of view: we are not scared about failures, we test to **find** problems before our users do ;)

tranchitella:
Superseded by https://github.com/mendersoftware/integration/pull/803

PR #782, opened 2020-01-14
Enable back two state script tests

lluiscampos:
That were disabled by mistake in commit e7947f3e.


lluiscampos:
The tests passed. Merging.

PR #783, opened 2020-01-15
aligned failover composition with the rest of the project

mzedel:
minio, storage proxy, mongo & redis versions were off & caused unneeded pulls
besides this the deprecated deviceadm service was still referenced -> no more

Changelog: None

### DIFF
--- a/extra/failover-testing/docker-compose.failover-server.yml
+++ b/extra/failover-testing/docker-compose.failover-server.yml
@@ -57,7 +57,6 @@ services:
         # these servers and exits with 'upstream server not found'
         depends_on:
             - mender-device-auth-2
-            - mender-device-adm-2
             - mender-deployments-2
             - mender-gui-2
             - mender-useradm-2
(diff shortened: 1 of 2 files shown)


mzedel:
This is to check if it is ok to remove the custom minio & openresty forks we have (and the Travis pipelines connected to them) and also to save some time by reusing images that are part of the overall project setup. Unfortunately I couldn't find a nice way to reference the image used in another service (e.g. the main redis or the main minio) thus the still hardcoded version strings

mzedel:
though the pipeline failed, the test in question ran successfully, so the https://github.com/mendersoftware/minio and https://github.com/mendersoftware/openresty forks could be obsolete... (along with their Travis pipelines)

extra/failover-testing/docker-compose.failover-server.yml
```diff
@@ -162,15 +146,15 @@ services:
             - CONFIG_PROP=config.properties
 
     mender-elasticsearch-2:
-        image: mendersoftware/elasticsearch:2.4
+        image: elasticsearch:5-alpine
```
lluiscampos: What about `elasticsearch`, is our fork only used in this test?

mzedel: as far as I can tell... and 2.4 is definitely out of date...

mzedel: and the https://hub.docker.com/r/mendersoftware/elasticsearch/tags looks interesting

PR #785, opened 2020-01-16
Improve robustness of docker_compose_manager and related fixes

lluiscampos:
This PR brigs a fix for [QA-140](https://tracker.mender.io/browse/QA-140) and other minor fixes and cleanups.

Mainly:

- Always add fixture finalizer before calling setup (fixes QA-140).
This is the way that we will ensure that the finalizer (teardown) will
happen even when an error appears during setup.

- Fix the retry logic of the docker-compose calls
On the one hand, starting the range from 0 made the framework not
sleep between the first and second attempt.
On the other hand, some containers might still be up after a failed
docker-compose up call so the second call to docker-compose up with
the same files will immediately return success. Now we terminate the
containers between attempts.

- Unify and clean logging usage across tests to help a bit humans when reading the tests logs.

- Move have_token function from common to the only test using it

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -95,13 +95,25 @@ def pytest_configure(config):
     MenderTesting.set_test_conditions(config)
 
 
-def pytest_runtest_setup(item):
-    logger = log.setup_custom_logger("root", item.name)
-    logger.info("%s is starting.... " % item.name)
+# If we have xdist installed, the testlogger fixture will include the thread id
+try:
+    import xdist
+    @pytest.fixture(scope="function", autouse=True)
+    def testlogger(request, worker_id):
+        test_name = request.node.name
+        logger = log.setup_custom_logger("root", test_name, worker_id)
+        logger.info("%s is starting.... " % test_name)
+except ImportError:
+    @pytest.fixture(scope="function", autouse=True)
+    def testlogger(request):
+        test_name = request.node.name
+        logger = log.setup_custom_logger("root", test_name, None)
+        logger.info("%s is starting.... " % test_name)
+
 
 def pytest_exception_interact(node, call, report):
     if report.failed:
-        logging.error("Test %s failed with exception:\n%s" % (str(node), call.excinfo.getrepr()))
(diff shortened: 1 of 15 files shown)


lluiscampos:
The PR is now ready for final review. I edited the PR description to better describe the scope.

[Here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/408156421) or [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/409448934) you can see test runs with multiple failures at setup (because they were run with parallel tests to 4) but where the framework managed to clean up properly and avoid this hung issue. So I consider [QA-140](https://tracker.mender.io/browse/QA-140) fixed.

Note that the changes I commented wrt the logs are not included in this PR. I will probably launch a new PR in the upcoming days for that.

@kacf please have another look. Your original review only contained the first commit (and actually it still had bugs at the time and has been reworked).

tests/conftest.py
```diff
@@ -95,13 +95,25 @@ def pytest_configure(config):
...
+        test_name = request.function.__name__
+        logger = log.setup_custom_logger("root", test_name, worker_id)
+        logger.info("%s is starting.... " % test_name)
+except ImportError:
+    @pytest.fixture(scope="function", autouse=True)
+    def testlogger(request):
+        test_name = request.function.__name__
+        logger = log.setup_custom_logger("root", test_name, None)
+        logger.info("%s is starting.... " % test_name)
+
```
lluiscampos: I think I am going to revert this.

I transformed it into a fixture to be able to have the xdist worker id, but the drawback is that we don't capture correctly the test name. For parametrized tests with plugin implementation `item.name` we would get `test_reboot_recovery[simulate_powerloss_in_artifact_commit_leave-test_set2]` but with fixture implementation `request.function.__name__` we get only `test_reboot_recovery`.

So unless someone illuminates me on how to get the full qualified test name in a fixture, I am reverting it.

lluiscampos: I found it! We can use `request.node.name`. Reworking...

PR #786, opened 2020-01-20
backend-tests: remove volumes and orphans when doing cleanup

kjaskiewiczz:


kjaskiewiczz:
test_full_integration failed, but the change affects only test_backend_integration which passed
that's why I'm merging the PR without restarting test_full_integration

PR #787, opened 2020-01-20
Add create-artifact-worker service to docker-compose

tranchitella:
We introduce a new Mender component, named create-artifact-worker, that requires
a new container in the docker composition:

- mender-create-artifact-worker: workflows worker creating artifacts

Changelog: None

### DIFF
--- a/component-maps.yml
+++ b/component-maps.yml
@@ -121,6 +121,13 @@ git:
     - mender-workflows-server
     release_component: true
 
+  create-artifact-worker:
+    docker_image:
+    - create-artifact-worker
+    docker_container:
+    - mender-create-artifact-worker
+    release_component: true
+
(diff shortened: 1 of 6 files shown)


tranchitella:
> * Is this really a release component?
Can you please define what is a release component?
This is something we are going to assign a version number, if it answers your question.

> * We are missing an update in `./production/config/prod.yml.template`
I believe I've updated this file, isn't it enough?

> * We are also missing a PR on `mender-qa` (but you are probably on it already?)
Almost ready to open it :)

lluiscampos:
> > * We are missing an update in `./production/config/prod.yml.template`
> 
> I believe I've updated this file, isn't it enough?

Sorry, you are right. I missed it!

PR #789, opened 2020-01-22
Test logging to files

lluiscampos:
```
Create log files per test and append all containers logs at teardown

Redeisgn the way the logging is being done around the test framework.

The idea is for all tests and helpers to use the same default logger (in
effect this was the case before also) but now adding a per test file
handler that can log at debug level.

Then, instead of having a background process dumping docker logs to /tmp
files, just append all containers logs to the logger (i.e. to the file)
at teardown time.

This will avoid 1) printing enormous logs into console on test failures
and 2) not having control on which docker log corresponds to which test.

All test logs will be saved at <integration>/tests/mender_test_logs/
```

### DIFF
--- a/tests/log.py
+++ b/tests/log.py
@@ -1,22 +1,64 @@
+import os
 import logging
+import unicodedata
+import re
 
-def setup_custom_logger(name, testname, worker_id=None):
-    log_format = "%(asctime)s [%(levelname)s]: >> %(message)s"
+TEST_LOGS_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), "mender_test_logs")
 
-    logging.basicConfig(format=log_format, level=logging.INFO)
-    logger = logging.getLogger(name)
+# Default logger deafult level to DEBUG
+logging.getLogger().setLevel(logging.DEBUG)
 
-    for h in list(logger.handlers):
-        logger.removeHandler(h)
+def setup_test_logger(test_name, worker_id=None):
+    """Sets the default test logger
 
-    consoleHandler = logging.StreamHandler()
-    logFormatter = logging.Formatter(log_format)
+    The  logger contains two handlers:
+    - An INFO level console handler
+    - A DEBUG level file handler, with a filename being an slug of the test name
+    """
+
+    # Get the default logger and remove previous handlers
+    logger = logging.getLogger()
+    for handler in list(logger.handlers):
+        logger.removeHandler(handler)
+
+    # Define format. For console logging, prepend the test name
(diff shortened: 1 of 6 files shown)


lluiscampos:
This is now ready for review.

You can find how the logs will look like [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/411545720/artifacts/browse/mender_test_logs/) (from a previous run missing last commit).

lluiscampos:
FYI @merlin-northern

lluiscampos:
Test logs for the latest run can be found [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/412678765/artifacts/browse/mender_test_logs/).

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -153,10 +143,16 @@ def _stop_docker_compose(self):
             logger.info("running %s" % cmd)
             subprocess.check_call(cmd, shell=True)
 
+    def _debug_log_containers_logs(self):
+        logs = self._docker_compose_cmd("logs --no-color")
+        for line in logs.split('\n'):
+            logger.debug(re.sub(r'[\r\n]*', '', line))
+
```
tranchitella: As you are using the regular expression N times, one per row, it is more efficient to compile it.
You can define, at class level:

> re_newlines_sub = re.compile(r'[\r\n]*').sub

And then call it:

> logger.debug(self.re_newlines_sub('', line))

tests/log.py
```diff
@@ -1,22 +1,60 @@
...
+    and converts spaces to hyphens.
+
+    Inspired by Django framework:
+    https://github.com/django/django/blob/3.0.2/django/utils/text.py#L393
+    """
+
+    value = unicode(value)
+    value = unicodedata.normalize('NFKD', value).encode('ascii', 'ignore').decode('ascii')
+    value = re.sub(r'[^\w\s-]', '', value).strip().lower()
+    return re.sub(r'[-\s]+', '-', value)
```
tranchitella: Same here, you can compile regexp and reuse them.

testutils/infra/container_manager/docker_compose_manager.py
lluiscampos: Makes sense. Thanks for the tip.

lluiscampos:
@tranchitella Thanks for the review! I just force-pushed with your latest suggested optimization.

lluiscampos:
No need to run again the integration pipeline. Merging.

PR #791, opened 2020-01-23
MEN-3041: create_artifact back-end integration tests

tranchitella:
Changelog: none

### DIFF
--- a/backend-tests/tests/test_create_artifact.py
+++ b/backend-tests/tests/test_create_artifact.py
@@ -1,0 +1,101 @@
+# Copyright 2020 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        https://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import os
+import requests
+import tempfile
+import time
+
+from json import dumps
+
+from testutils.api.client import ApiClient
+from testutils.api import deployments, useradm
+from testutils.common import create_org, mongo, clean_mongo
+
+
+class TestCreateArtifactEnterprise:
+    def test_create_artifact(self, mongo, clean_mongo):
+        tenant, username, password = "test.mender.io", "user@test.mender.io", "secret"
+        tenant = create_org(tenant, username, password)
+
+        r = ApiClient(useradm.URL_MGMT).call(
+            "POST", useradm.URL_LOGIN, auth=(username, password)
+        )
+        assert r.status_code == 200
+        assert r.text is not None
+        assert r.text != ""
+
+        auth_token = r.text
+
+        api_client = ApiClient(deployments.URL_MGMT)
+        api_client.headers = {}  # avoid default Content-Type: application/json
+        api_client.with_auth(auth_token)
+
+        f = tempfile.NamedTemporaryFile(delete=False)
+        try:
+            f.write(b"#!/bin/bash\ntrue\n")
+            f.close()
+
+            filename = f.name
+
+            r = api_client.call(
+                "POST",
+                deployments.URL_DEPLOYMENTS_ARTIFACTS_GENERATE,
+                files={
+                    "name": (None, "artifact"),
+                    "description": (None, "description"),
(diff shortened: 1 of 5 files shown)


tranchitella:
Ready for review, it is in the draft state because I don't want the tests to run until we have all the updated docker images we need to be uploaded to the registry.

backend-tests/tests/test_create_artifact.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2018 Northern.tech AS
```
lluiscampos: Update to 2020

backend-tests/tests/test_create_artifact.py
```diff
@@ -0,0 +1,82 @@
...
+        files={
+            "name": (None, "artifact"),
+            "description": (None, "description"),
+            "device_types_compatible": (None, "beaglebone"),
+            "type": (None, "single_file"),
+            "args": (None, dumps({"filename": "run.sh", "dest_dir": "/tests",})),
+            "file": (filename, open(filename, "rb"), "application/octet-stream", {}),
+        },
+        verify=False,
+    )
```
lluiscampos: As it is now, `api_client` is unused.

In other bakend-tests I have seen the requests expressed in the following manner:
```
api_client.with_auth(auth_token).call("POST", deployments.URL_DEPLOYMENTS_ARTIFACTS_GENERATE, files=...)
```

backend-tests/tests/test_create_artifact.py
```diff
@@ -0,0 +1,82 @@
...
+
+    artifact = None
+    for i in range(15):
+        time.sleep(1)
+        r = requests.get(
+            artifact_url,
+            headers={"Authorization": "Bearer " + auth_token},
+            verify=False,
+        )
```
lluiscampos: Also here could be standardized.

backend-tests/tests/test_create_artifact.py
```diff
@@ -0,0 +1,82 @@
...
+    tenant, username, password = "test.mender.io", "user@test.mender.io", "secret"
+    tenant = create_org(tenant, username, password)
+    r = ApiClient(useradm.URL_MGMT).call(
+        "POST", useradm.URL_LOGIN, auth=(username, password)
+    )
+    assert r.status_code == 200
+    assert r.text is not None
+    assert r.text != ""
+
+    filename = "/tests/run.sh"
```
lluiscampos: Should we write some data into the file? It would still be valid behavior if mender-artifact refused to create an update module with an empty script.

docker-compose.yml
```diff
@@ -116,6 +116,9 @@ services:
             service: mender-base
         environment:
             - WORKFLOWS_MONGO_URL=mongodb://mongo-workflows:27017
+            - CREATE_ARTIFACT_GATEWAY_URL=https://mender-api-gateway
+            - CREATE_ARTIFACT_DEPLOYMENTS_URL=http://mender-deployments:8080
+            - CREATE_ARTIFACT_SKIPVERIFY=1
```
lluiscampos: What does this `SKIPVERIFY` do? It looks to me like something that should be in the `docker-compose-demo.yml` instead of here.

backend-tests/tests/test_create_artifact.py
tranchitella: Well spotted, thanks!

docker-compose.yml
tranchitella: @lluiscampos it is needed to skip verification of tls (aka, HTTPS) certificates when running the tests; you are right, it makes sense to move them to demo.

backend-tests/tests/test_create_artifact.py
tranchitella: @lluiscampos 

I'm using `/tests/run.sh` in the container as file to install; I just need a random file to upload; if you believe it is a better solution, I can write a temporary file and use that instead, it doesn't change anything from test's point of view.

backend-tests/tests/test_create_artifact.py
tranchitella: Right, I'm changing the code.

backend-tests/tests/test_create_artifact.py
tranchitella: Right, I'm changing the code.

backend-tests/tests/test_create_artifact.py
lluiscampos: I misunderstood then! Now I see: that file actually exists and it is the tests wrapper :)

So for another reason now: but I still think we should change it. The test should be framework agnostic. What if I run the test locally (not from a container with a volume mounted in a very specific way)? Of what if we rename the script or move it somewhere else?

I think a temp file with just dummy content will do:
```
#!/bin/bash
true
```

tranchitella: Ok, done!

PR #794, opened 2020-01-24
Fix backend-tests warnings

lluiscampos:
```
commit 0e15ad87cae407f490ed329f387622e4ebf6d64e
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Fri Jan 24 13:13:53 2020 +0100

    Fix wrong parentheses assert in test_deployments
    
    Which was producing an always true assertion. Fixes warning:
    PytestAssertRewriteWarning: assertion is always true, perhaps remove
    parentheses?

commit a18daf665253c251bb7eccf98ceacc805558601a
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Fri Jan 24 13:16:19 2020 +0100

    Upgrade all pip dependencies in backend-tests
    
    Update all deps to latest stable today. The intention is to remove a
    warning from pymongo being printed 1722 times.
```

### DIFF
--- a/backend-tests/requirements-py3.txt
+++ b/backend-tests/requirements-py3.txt
@@ -1,8 +1,8 @@
-requests==2.22
-pymongo==3.6.1
+requests==2.22.0
+pymongo==3.10.1
 cryptography==2.8
-pytest==5.2
-pytest-html==2.0
+pytest==5.3.4
+pytest-html==2.0.1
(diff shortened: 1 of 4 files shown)


lluiscampos:
All backend tests passed.

This PR is only about updating deps and clearing warnings, so merging without formal revision.

PR #799, opened 2020-01-30
[MEN-3090] Fix setup for running without SSL termination

breuerfelix:
How could this setup be optimized ?

- Create a new GitHub Repo for the storage proxy and bake the nginx.conf into a docker image
  - disable / enable SSL via entrypoint.sh just like api-gateway
  - idea: do the same for elasticsearch and redis, this will clean the integration repo alot
- Use non-SSL setup for the demo setup
  - We would get rid of nginx.conf.demo
  - Get rid of generated Certificates
  - Get rid of "Do you want to trust this website ?" for demo setup

There is some duplicated code in `docker-compose.non-ssl.yml` and `docker-compose.demo.yml`.  
I could create a new file called `docker-compose.mounts.yml` for example. Non-ssl and demo will inherit from this file also.

These are all ideas in my head, I would love to know what you think about that !

P.S. this PR requires [this](https://github.com/mendersoftware/mender-api-gateway-docker/pull/109) to be merged.

### DIFF
--- a/docker-compose.no-ssl.yml
+++ b/docker-compose.no-ssl.yml
@@ -3,13 +3,14 @@ services:
 
     mender-api-gateway:
         ports:
-            - "8090:80"
+            - "80:80"
+        environment:
+            SSL: "false"
+
+    storage-proxy:
+        volumes:
+            - ./storage-proxy/no-ssl.nginx.conf:/usr/local/openresty/nginx/conf/nginx.conf
 
     mender-deployments:
         environment:
-            DEPLOYMENTS_AWS_URI: http://minio.s3.docker.mender.io:9001
-
-    minio:
-        ports:
-            - "9001:9001"
-        command: server --address ":9001" /export
+            DEPLOYMENTS_AWS_URI: http://s3.docker.mender.io:9000
--- a/README.md
+++ b/README.md
@@ -201,19 +201,17 @@ how to modify it.
 
 ### Enabling non-SSL access
 
-For debugging purposes, it may be useful to temporarily enable non-SSL access.
-API Gateway configuration enables plain HTTP on port 80, however the port is not
-published by default, thus it remains inaccessible from the outside. For
-convenience, an overlay compose file is provided that publishes port 80 of API
-Gateway to port 8090 on current host. The overlay file has to be explicitly
-included when setting up the environment like this:
+For debugging purposes or when using third party SSL reverse proxy, it may be useful to enable non-SSL access.  
+API Gateway configuration enables plain HTTP on port 80 when setting the `SSL` environment variable to `'false'`.  
+The nginx configuration will only be changed on container creation. If you previously ran with SSL, delete and re-create the container.  
+An example compose file can be included like this:
 
 ```
-docker-compose ... -f docker-compose.no-ssl.yml ...
+./demo -f docker-compose.no-ssl.yml up
 ```
 
 **NOTE** make sure that plain HTTP port is not published in production
-deployment.
+deployment. Use a reverse proxy for example.
 
 ## Demo client
 
--- a/demo
+++ b/demo
@@ -123,6 +123,12 @@ fi
 
 MENDER_SERVER_URI="https://localhost"
 
+# use http when providing no-ssl config
+if [[ "$@" == *"-f docker-compose.no-ssl.yml"* ]]
+then
+    MENDER_SERVER_URI="http://localhost"
+fi
+
 # Make sure that the demo environment is brought down on SIGINT
 exitfunc() {
     retval=$(docker-compose \
--- a/storage-proxy/no-ssl.nginx.conf
+++ b/storage-proxy/no-ssl.nginx.conf
@@ -1,0 +1,101 @@
+worker_processes  auto;
+
+error_log  /dev/stdout warn;
+pid        /var/run/nginx.pid;
+
+events {
+    worker_connections  1024;
+}
+
+env DOWNLOAD_SPEED;
+env MAX_CONNECTIONS;
+
+http {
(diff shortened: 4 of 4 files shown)


tranchitella:
@breuerfelix, thanks for your contributions to the Mender project, much appreciated.
Now that your PR on the api-gateway got merged (https://github.com/mendersoftware/mender-api-gateway-docker/pull/109), we can review this one. While supporting HTTP is a useful feature for our users, we strongly disagree with allowing them to run the demo environment in HTTP mode unless they know what they are doing.

May I ask you to rework your PR, limiting the changes to allow the users to run Mender in HTTP mode?

As we discussed in the other PR, we are exploring different solutions in the reverse proxy area, and we are not sure we'll keep using openresty in the next releases. Let's try to minimize the impact of this PR to allow users to run Mender using their own SSL termination proxy without changing the default settings.

Thank you in advance!

breuerfelix:
@tranchitella Thanks for the review !

I just removed the section on how to run mender with a reverse proxy :)  
I only provided the correct docker-compose.no-ssl.yml and the correct no-ssl nginx conf for the storae proxy :)

docker-compose.no-ssl.yml
```diff
@@ -1,15 +1,72 @@
...
+            - ./keys/useradm/private.key:/etc/useradm/rsa/private.pem
+
+    mender-device-auth:
+        command: server --automigrate
+        volumes:
+            - ./keys/deviceauth/private.key:/etc/deviceauth/rsa/private.pem
+
+    mender-inventory:
+        command: server --automigrate
+
```
tranchitella: I don't get why we need these settings here, can you please explain?

docker-compose.no-ssl.yml
```diff
@@ -1,15 +1,72 @@
...
+        environment:
+            # use nginx syntax for rate limiting, see
+            # https://nginx.org/en/docs/http/ngx_http_core_module.html#limit_rate
+            # Examples:
+            #   1m - 1MB/s
+            #   512k - 512kB/s
+            DOWNLOAD_SPEED: 3m
+            MAX_CONNECTIONS: 30
+        volumes:
+            - ./storage-proxy/no-ssl.nginx.conf:/usr/local/openresty/nginx/conf/nginx.conf
```
tranchitella: Same here.

.gitignore
```diff
@@ -1,9 +1,9 @@
 *.log
 *.pyc
 
-
 tags
 
+/keys-generated
```
tranchitella: I suppose this is a left-over from previous patches.

docker-compose.no-ssl.yml
breuerfelix: I copied over these settings from `docker-compose.demo.yml`.

The problem here is that you cant run it like this:
```
docker-compose -f docker-compose.yml -f docker-compose.demo.yml -f docker-compose.no-ssl.yml up
```
because the demo file mounts the certificates for api gateway and storage proxy. Sadly in compose you cannot revoke this in a compose file which got merged later.

So I copied everything from the demo file and removed all the lines according to certificates.

With these changes you can run the command I mentioned in the README file and it works. Otherwise, the user needs to do changes on his own.

tranchitella:
@breuerfelix can you have a look at my PR to see if it would be enough, from your point of view?
https://github.com/mendersoftware/integration/pull/804

docker-compose.no-ssl.yml
tranchitella: The right way to run the demo is using the demo script, indeed.
`./demo -f docker-compose.no-ssl.yml up`
This brings a full working demo with HTTP-only support, see my other PR for details.

breuerfelix: Now I understand what you are trying to do :)  
That way the certificates still get mounted but they will be ignored.

Problem: the storage proxy still got HTTPS. I will try to adjust my PR furthermore.

breuerfelix:
@tranchitella just updated my PR with the minimum for running without HTTPS and updated the readme aswell. Certificates will be created but ignored.

tranchitella:
@breuerfelix something strange happens, I cannot run the demo from your branch:
```
integration (breuerfelix/master) $ ./demo -f docker-compose.no-ssl.yml up
Starting the Mender demo environment...
Creating network "integration_mender" with the default driver
Creating integration_mender-mongo_1         ... done
Creating integration_mender-gui_1           ... done
Creating integration_mender-redis_1                  ... done
Creating integration_mender-elasticsearch_1 ... done
Creating integration_minio_1                         ... done
Creating integration_mender-create-artifact-worker_1 ... done
Creating integration_mender-useradm_1                ... done
Creating integration_mender-workflows-server_1       ... done
Creating integration_mender-inventory_1              ... done
Creating integration_mender-conductor_1              ... done
Creating integration_mender-device-auth_1            ... done
Creating integration_storage-proxy_1                 ... done
Creating integration_mender-deployments_1            ... done
Creating integration_mender-api-gateway_1            ... done
It does not seem the useradm service is up and running yet. Retrying...
It does not seem the useradm service is up and running yet. Retrying...
It does not seem the useradm service is up and running yet. Retrying...
It does not seem the useradm service is up and running yet. Retrying...
It does not seem the useradm service is up and running yet. Retrying...
It does not seem the useradm service is up and running yet. Retrying...
Retried 5 times without success. Giving up.
```
Any idea why this is happening?

breuerfelix:
@tranchitella fixed that.

sorry that I missed that ... It failed cause the demo script was still testing on https:// if the server is up and running.

tranchitella:
Test failures are unrelated to changes in this PR, let's merge it anyway.

PR #801, opened 2020-01-31
backend-tests: retry user login after creating organization

kjaskiewiczz:
Create organization is an async job.
It's possible that the email has been sent but the user could not log in yet.
The only thing we can do is to retry user login.

### DIFF
--- a/backend-tests/tests/test_create_org.py
+++ b/backend-tests/tests/test_create_org.py
@@ -67,10 +67,23 @@ class TestCreateOrganizationEnterprise:
         smtp_mock.assert_called()
         logging.info("TestCreateOrganizationEnterprise: Assert ok.")
 
-        r = uc.call(
-            "POST", useradm.URL_LOGIN, auth=("some.user@example.com", "asdfqwer1234")
-        )
-        assert r.status_code == 200
+        # Try log in every second for 3 minutes.
+        # - There usually is a slight delay (in order of ms) for propagating
+        #   the created user to the db.
+        for i in range(3 * 60):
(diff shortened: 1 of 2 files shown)


backend-tests/tests/test_create_org.py
```diff
@@ -67,10 +67,21 @@ def test_success(self, clean_migrated_mongo):
...
+        #   the created user to the db.
+        for i in range(3*60):
+            rsp = api.call(
+                "POST",
+                useradm.URL_LOGIN,
+                auth=("some.user@example.com", "asdfqwer1234")
+            )
+            if rsp.status_code == 200:
+                break
+            time.sleep(1)
```
kacf: If this is a common pattern, we might think about wrapping it in some prebaked function.

kjaskiewiczz: we are moving tests from integration to backend integration so there is quite a lot happening in the backend integration tests recently;
we're trying to be careful when moving things to a common part of the test, but next time I'll need this pattern I'll move it there (to testutils/common)

kjaskiewiczz:
@lluiscampos done

tranchitella:
@kjaskiewiczz in the recently added create_artifact test we create an organization (and a user):
https://github.com/mendersoftware/integration/blob/master/backend-tests/tests/test_create_artifact.py#L108

Do you believe this thing can be related to the the test_create_organization failures?

kjaskiewiczz:
@tranchitella I don't think it's related
in our test setup, different tests shouldn't affect each other results

PR #803, opened 2020-02-01
[MEN-2985] Upgrade elasticsearch to version 6

tranchitella:
Changelog: upgrade Elasticsearch to version 6

### DIFF
--- a/docker-compose.demo.yml
+++ b/docker-compose.demo.yml
@@ -79,13 +79,6 @@ services:
         ports:
             - "8080:8080"
 
-    mender-elasticsearch:
-        environment:
-            - ES_NETWORK_HOST=0.0.0.0
-            - ES_JAVA_OPTS=-Xms0m -Xmx100m
-        volumes:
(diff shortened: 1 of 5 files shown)


docker-compose.demo.yml
```diff
@@ -79,13 +79,6 @@ services:
         ports:
             - "8080:8080"
 
-    mender-elasticsearch:
-        environment:
-            - ES_NETWORK_HOST=0.0.0.0
-            - ES_JAVA_OPTS=-Xms0m -Xmx100m
-        volumes:
-            - ./conductor/elasticsearch/elasticsearch.yml:/usr/share/elasticsearch/config/elasticsearch.yml
```
lluiscampos: I think we need to remove this same volume from the `prod.yml.template`, [here](https://github.com/mendersoftware/integration/blob/master/production/config/prod.yml.template#L132).

tranchitella: Right, force-pushed to remove this volume as well.

PR #805, opened 2020-02-06
Misc fixes during release

lluiscampos:
```
[release_tool] Update instructions to build/extract gui disclaimer
The gui docker image does not longer contain the disclaimer, it is built
separately.

[statistics-generator] Add companies to Company Map
Note that:
- tranchitella.eu is the personal domain of Fabio Tranchitella, but he
has been used this signature for NT work.
- Alex Miliukov has authored commits with users.noreply.github.com, so
we assume from now on that all these GitHub aliases are NT.
```

For the first commit, see also https://github.com/mendersoftware/gui/pull/624

### DIFF
--- a/extra/gitdm/company-map
+++ b/extra/gitdm/company-map
@@ -1,6 +1,8 @@
 northern.tech Northern.tech
 mender.io Northern.tech
 cfengine.com Northern.tech
+tranchitella.eu Northern.tech
+users.noreply.github.com Northern.tech
 rndity.com RnDity
 amarulasolutions.com Amarula Solutions
 essvote.com Election Systems & Software
@@ -24,3 +26,4 @@ reachtest.com Reach Technologies Inc
 trellis-logic.com Trellis-Logic
 walkbase.com walkbase
 ni.com National Instruments
+baylibre.com BayLibre
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -1280,7 +1280,7 @@ def do_license_generation(state, tag_avail):
             gui_tag = "mendersoftware/gui:tmp"
             for tmpdir in tmpdirs:
                 if os.path.basename(tmpdir) == "gui":
-                    query_execute_list([["docker", "build", "-t", gui_tag, tmpdir]])
+                    query_execute_list([["docker", "build", "-t", gui_tag, "-f", os.path.join(tmpdir, "Dockerfile.disclaimer"), tmpdir]])
                     break
         elif reply == "2":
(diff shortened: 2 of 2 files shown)


lluiscampos:
FYI @tranchitella and @0lmi

lluiscampos:
Also FYI @mzedel, if you are curious about how the release tool generates the licenses file.

extra/gitdm/company-map
```diff
@@ -1,6 +1,8 @@
 northern.tech Northern.tech
 mender.io Northern.tech
 cfengine.com Northern.tech
+tranchitella.eu Northern.tech
+users.noreply.github.com Northern.tech
```
kacf: Github uses this domain for all people who enabled private email settings. We cannot assume that such emails belong to Northern.tech, particularly when using the Github UI to make changes. There are already [several](https://github.com/mendersoftware/mender/commit/1198c9f34c98a7f16ea020276e7cad8726dabc00) [examples](https://github.com/mendersoftware/mender-docs/commit/23eab8336f6f4d3fa4074ee5d9046d945c936f46) [of this](https://github.com/mendersoftware/mender-docs/commit/5560cafbb6a4dc544e500ba772be3750ade6fd29).

We will have to add his dedicated name.

extra/release_tool.py
```diff
@@ -1289,9 +1289,8 @@ def tag_or_followed_branch(repo_git):
             return
 
         executed = query_execute_list([
-            ["docker", "run", "-d", "--name", "release_tool_gui_licenses", gui_tag],
-            ["docker", "cp", "release_tool_gui_licenses:/var/www/mender-gui/dist/disclaimer.txt", "gui-licenses.txt"],
-            ["docker", "rm", "-f", "release_tool_gui_licenses"],
+            ["docker", "run", "--rm", "--entrypoint", "/bin/sh", "-v", os.getcwd() + ":/extract", gui_tag,
+            "-c", "cp disclaimer.txt /extract/gui-licenses.txt"]
```
kacf: This solution has the root-problem, that the file is owned by root after this operation.

kacf: I reverted back to the original for now, as changing the file location seemed a simpler change than adding `chown`.

kacf:
New PR in #806.

extra/gitdm/company-map
lluiscampos: Thanks for resubmitting!

I tried with adding it to `mailmap` but missed the `groups-map` part, so it had no effect... But now I know for next time :)

extra/release_tool.py
lluiscampos: :+1:

PR #807, opened 2020-02-08
Fix back-end tests to specify enterprise plan by default

tranchitella:
Changelog: none


tranchitella:
Superseded by https://github.com/mendersoftware/integration/pull/808

PR #808, opened 2020-02-10
MEN-3145: Fix tests setting plan when creating organizations

tranchitella:


tranchitella:
Pipeline running with obac feature branches:
https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/116880702

PR #814, opened 2020-02-18
[changelog-generator] Fix bogus changelog entries with "None" text

lluiscampos:
Make the tool ignore message following a "Changelog: None" type of tag.
The previous logic was creating a bogus None entry if a text line was
found directly after the "Changelog: None" one.

Added a test case to test-changelog-entry to show the problem

### DIFF
--- a/extra/changelog-generator/changelog-generator
+++ b/extra/changelog-generator/changelog-generator
@@ -164,6 +164,8 @@ for repo in repos:
                     log_entry = title
                 elif re.match("^Commit[ .]*$", match.group(1), re.IGNORECASE):
                     log_entry_commit = True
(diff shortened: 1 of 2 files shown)


lluiscampos:
See this commit in `deployments` for some context: https://github.com/mendersoftware/deployments/pull/389/commits/e54fd0b1ec00432174ada742dc6f4ff600c27583

lluiscampos:
> Code looks good, but please create a test for this [here](https://github.com/mendersoftware/integration/blob/master/extra/changelog-generator/test-changelog-generator). Basically the test is a long list of git commits, with the resulting changelog near the bottom, so just add yours with a new number (it should then not change the output).

Good suggestion, thanks! Done.

PR #815, opened 2020-02-18
Password storage subdirectory

lluiscampos:
[release-tool] Allow for password storage in subdirectories
After the find command, parse and concatenate the tree output so that we
can retrieve the secret in cases where the storage is in a subdirectory.

[release-tool] Fix parsing of pass output
At least on my Ubuntu, "Search Terms" is capitalized, so the check was
actually failing to detect this and always assuming the key was found.

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -234,7 +234,7 @@ def get_value_from_password_storage(server, key):
         output = subprocess.check_output(["pass", "find", server]).decode()
         count = 0
         for line in output.split('\n'):
-            if line.startswith("Search terms: "):
+            if line.lower().startswith("search terms: ") or line == "":
                 continue
             count += 1
         if count == 0:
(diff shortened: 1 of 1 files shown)


lluiscampos:
> This will break if there is more than one entry, but I cannot think of a way to fix that which isn't annoyingly complicated. There are no additional entries for gitlab.com in mystiko ATM, so I'm not affected by this now. If you can think of a clever way to fix it, it would be a nice fix to have for the future, otherwise I'm ok with sweeping this under the rug for now.

Exactly. I also thought about that but didn't find an easy solution. 

The thing is that I need this PR because I moved all my passwords to a "lluistiko" subdirectory. As I already spend a considerable time here (matching these bash colors wasn't that clear to me!), I will merge as-is and fix if that day in the future arrives :)

PR #816, opened 2020-02-18
backend-integration: device deployment status update tests

kjaskiewiczz:
https://tracker.mender.io/browse/QA-34


backend-tests/tests/test_deployments.py
```diff
@@ -1105,3 +1137,299 @@ def test_two_phases_concurrent_devices(self, clean_mongo):
...
+        )
+        assert resp.status_code == 204
+        status_verify(
+            deploymentsm,
+            deployment_id,
+            user_token,
+            device_id=devs[1].id,
+            device_deployment_status="already-installed",
+            deployment_status="inprogress",
+        )
```
tranchitella: I suggest to refactor status_verify and status_update_and_verify as methods exposed by a helper class, something among the lines of...

```
class StatusVerifier:
    def __init__(self, deploymentsm, deploymentsd):
        self.deploymentsm = deploymentsm
        self.deploymentsd = deploymentsd

    def status_verify(self, deployment_id, user_token, ...):
        ...
````

In this way we can instantiate the helper once and use it multiple times:

```
status_verifier = StatusVerifier(deploymentsm, deploymentsd)
...
status_verifier.status_verify(
    user_token,
    device_id=devs[1].id,
    device_deployment_status="already-installed",
    deployment_status="inprogress",
)
```
To make it more readable, I'd always use keyword arguments `deployment_status="inprogress"` instead of positional arguments.

Feel free to ignore this suggestion, it doesn't really matter and the test is just fine.

Comment (kjaskiewiczz) [reply backend-tests/tests/test_deployments.py]: @tranchitella good suggestions; thanks!
I've applied them in a separate commit - https://github.com/mendersoftware/integration/pull/816/commits/e02ab157205971c366442795d4b008302d1d8921

Comment (tranchitella) [reply backend-tests/tests/test_deployments.py]: Great!

File: backend-tests/tests/test_deployments.py [hunk 2]
```diff
@@ -1105,3 +1137,299 @@ def test_two_phases_concurrent_devices(self, clean_mongo):
...
+        # wrong status
+        status_update_and_verify(
+            deploymentsm,
+            deploymentsd,
+            devs[0].id,
+            devs[0].token,
+            deployment_id,
+            user_token,
+            "foo",
+            "pending",
```
merlin-northern: just a nitpick: I would consider making "pending", "inprogress", "installing", "downloading" into consts

backend-tests/tests/test_deployments.py
```diff
@@ -1105,3 +1137,289 @@ def test_two_phases_concurrent_devices(self, clean_mongo):
...
+            status_update="pending",
+            device_deployment_status="success",
+            deployment_status="inprogress",
+            status_update_error_code=400,
+            substate="bar",
+        )
+
+        # second device
+
+        # device deployment: pending -> success
```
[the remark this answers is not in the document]
kjaskiewiczz: right, fixed
thanks

kjaskiewiczz:
@lluiscampos about backend-integration vs acceptance - I'm thinking of backend integration tests like of API tests.
Thus, it doesn't really matter to me if the service is calling other services while handling the request or not (it may change in the future; in this particular case it's possible that deployments will start to propagate data to inventory on status update at some point). I just want to test the endpoint.

IMO, the main difference between our acceptance and backend integration tests is that in the acceptance tests I can mock other services and test error handling by mocking errors which I cannot produce normally, by using the API (like 500).

kjaskiewiczz:
@lluiscampos one more thing - test_state_scripts.py from integration looks to me like mostly related to mender client and if I'm not wrong it's also time and resource consuming.
Now we'll have a backend integration test for status reporting endpoint.
Would it be possible and reasonable to disable this test for backend PRs?

lluiscampos:
> @lluiscampos about backend-integration vs acceptance - I'm thinking of backend integration tests like of API tests.

Your point of view makes sense. Thanks for clarifying.

lluiscampos:
> Would it be possible and reasonable to disable this test for backend PRs?

This is a good suggestion, indeed: to have tests "only" for client changes. We don't have such filtering functionality ATM (actually, we run backend-tests on client PRs also) but it sounds like a good idea so I will think about it.

kacf:
I was curious about this, so I [ran the numbers](https://gist.github.com/kacf/df978b2d7c8af2597284002fe159875d). `test_state_scripts` is responsible for less than 20% of the total running time, and is not the longest running one, despite having the most sub tests. It seems to be pretty evenly distributed among many different tests.

So the question is if it's worth it then. I think we have a lot of complexity in the tests already, so I'm a bit reluctant to have even more.

PR #819, opened 2020-02-24
Merge OBAC feature-branch into master

tranchitella:


lluiscampos:
> Looking at the tests change set, I cannot see where are we actually testing the OBAC itself. I was expecting to find tests like "given this os tenant, such and such API calls shall be declined". Are these only in acceptance testing? Can you please point me to them?

I guess [this](https://github.com/mendersoftware/tenantadm/pull/122/commits/857b83ca815c0ea356356936ecaafd10f3cd5152) is the commit I was looking for :)

PR #820, opened 2020-02-24
[integ-tests] Modify test_double_update_rofs to use lzma

lluiscampos:
There is no test in the full suite that verifies support for lzma
compressed Artifacts. Extend this test rather than create a new one just
for this.


tests/tests/test_basic_integration.py
```diff
@@ -35,7 +35,7 @@ class TestBasicIntegration(MenderTesting):
     @pytest.mark.skipif(not os.path.exists("mender-image-full-cmdline-rofs-%s.ext4" % conftest.machine_name),
                         reason="Thud branch and older from Yocto does not have R/O rootfs support")
     def test_double_update_rofs(self, standard_setup_one_rofs_client_bootstrapped):
-        """Upgrade a device with two consecutive R/O images"""
+        """Upgrade a device with two consecutive R/O images using different compression algorithm"""
```
oleorhagen: Nitpick: s/algorithm/algorithms

lluiscampos: Thank you my dear spell checker!

oleorhagen: s/spell/grammar

PR #831, opened 2020-03-05
Add inventory-enterprise to the GIT_TO_BUILDPARAM map

oleorhagen:
Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -172,6 +172,7 @@ GIT_TO_BUILDPARAM_MAP = {
     "deviceauth": "DEVICEAUTH_REV",
     "gui": "GUI_REV",
     "inventory": "INVENTORY_REV",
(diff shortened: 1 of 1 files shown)


oleorhagen:
Should be the last missing piece...

extra/release_tool.py
```diff
@@ -172,6 +172,7 @@ def is_release_component(self):
     "deviceauth": "DEVICEAUTH_REV",
     "gui": "GUI_REV",
     "inventory": "INVENTORY_REV",
+    "inventory-enterprise": "INVENTORY_ENTERPRISE_REV",
```
lluiscampos: ping @tranchitella: to be added to the list of things to do on new repos.

We could think about a better way to do this, though. :thinking:

oleorhagen: There has been quite a few pieces needed for this one simple thing...

PR #833, opened 2020-03-05
all: hide stripe key under env var

mchalski:
this is a very recent followup:
https://tracker.mender.io/browse/MEN-3274


merlin-northern:
@mchalski merged as we agreed.

PR #834, opened 2020-03-05
dump-gitlab-env: pull and display gitlab envvars

mchalski:
some of the env is out there in compose files, but
some is hidden in gitlab (mender-qa).

to get reliable local test runs it's good to have
them pulled from the single source of truth.

this util pulls the env vars from the gitlab project and
dumps them in a format suitable for sourcing to your shell.

changelog: none

### DIFF
--- a/extra/dump-gitlab-env.py
+++ b/extra/dump-gitlab-env.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+import os
+import requests
+import argparse
+
+GITLAB_TOKEN = os.environ.get("GITLAB_TOKEN")
+if GITLAB_TOKEN is None:
+    raise RuntimeError("provide the GITLAB_TOKEN variable!")
+
+# not every var is interesting for local dev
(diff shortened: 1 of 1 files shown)


extra/dump-gitlab-env.py
```diff
@@ -0,0 +1,66 @@
...
+import os
+import requests
+import argparse
+
+GITLAB_TOKEN = os.environ.get("GITLAB_TOKEN")
+if GITLAB_TOKEN is None:
+    raise RuntimeError("provide the GITLAB_TOKEN variable!")
+
+# not every var is interesting for local dev
+DEV_VARS = ["STRIPE_API_KEY", "AWS_SECRET_ACCESS_KEY", "AWS_ACCESS_KEY_ID"]
```
lluiscampos: The `AWS_SECRET_ACCESS_KEY` and `AWS_ACCESS_KEY_ID` are not testing keys: they are the ones used to publish deliverables into S3. So why do we need them for local testing? We should then have two pairs of keys as we do in other Pipelines, and this tool should only retrieve the one with lower privileges.

PR #838, opened 2020-03-07
QA-119: Encapsulate all SSH operations in dedicated classes

lluiscampos:
```
    This commit moves absolutely all SSH operations (run/put) dispersed
    across single tests and helpers into classes MenderDevice and
    MenderDeviceGroup. The main goal is to have a single place where to
    manage Fabric API calls.

    For now, the different setups attach a .device/.device_gruop obj to the
    already shared containers_manager instance that the individual tests can
    use for SSH operations or Helper functions. This could be even nicer
    with a higher scoped class "TestEnv" or similar which can then serve the
    test for accessing containers, devices (and MenderAPI in the future!).
    
    The old run/put methods in common.py have been moved for internal use in
    device.py; at the moment using Fabric 1 (Python 2) but paving the road
    for an smooth migration to Fabric 2 (Python 3).
    
    Note that tests using the Fabric 1 context managers (hide/warn_only)
    are now just passing these into the new run implementation while
    keeping the functionality as it was (i.e. tests that decided to ignore
    errors they continue doing it so). Tests that were having their own
    retry mechanism now use the timeout from run (which makes it a bit more
    noisy in runtime but cleaner code wise).
    
    Other minor changes:
    - common.py: Removed unused sudo method
    - common.py: Moved MENDER_STORE to the only test using it
    - misc tests: replace print with logger
    - misc tests: remove unused imports
    - misc tests: remove ignored variables
```


lluiscampos:
Ready for review :tada: 

The tests passed in the previous run, the last force-push added docstrings and clean-ups, so I don't expect big surprises.

Thank you for your time!

tests/tests/test_basic_integration.py
```diff
@@ -37,43 +36,39 @@ class TestBasicIntegration(MenderTesting):
...
-        if not env.host_string:
-            execute(self.test_double_update_rofs,
-                    standard_setup_one_rofs_client_bootstrapped,
-                    hosts=mender_clients)
-            return
+        mender_device = standard_setup_one_rofs_client_bootstrapped.device
 
         # Verify that partition is read-only as expected
-        run("mount | fgrep 'on / ' | fgrep '(ro,'")
+        mender_device.run("mount | fgrep 'on / ' | fgrep '(ro,'")
```
oleorhagen: This is so much nicer!

tests/tests/test_basic_integration.py
```diff
@@ -37,43 +36,39 @@ class TestBasicIntegration(MenderTesting):
...
-                                install_image="mender-image-full-cmdline-rofs-%s.ext4" % conftest.machine_name,
-                                compression_type="gzip")
-        run("mount | fgrep 'on / ' | fgrep '(ro,'")
-
-        update_image_successful(host_ip,
-                                install_image="mender-image-full-cmdline-rofs-%s.ext4" % conftest.machine_name,
-                                compression_type="lzma")
-        run("mount | fgrep 'on / ' | fgrep '(ro,'")
-
+        update_image_successful(
```
oleorhagen: Idea:
Why not have this be a part of the device interface also? Like run above?
so:
```
mender_device.update_image_successful
```

I also think the Helper could be embedded in the device maybe? 

So that instead of `Helpers.foo.bar(device)`,
 we can do: `device.foo.bar`

tests/tests/common_update.py
```diff
@@ -103,8 +105,8 @@ def update_image_successful(host_ip,
         Logs will not be retrieved, and result in 404.
     """
 
-    previous_inactive_part = Helpers.get_passive_partition()
-    with Helpers.RebootDetector(host_ip) as reboot:
+    previous_inactive_part = Helpers.get_passive_partition(device)
```
oleorhagen: This is what I am talking about below, with regards to maybe embedding the helper in the `MenderDevice`

tests/tests/test_grouping.py
```diff
@@ -105,48 +101,43 @@ def test_update_device_group(self, standard_setup_two_clients_bootstrapped):
...
+        ip_to_device_id = Helpers.ip_to_device_id_map(mender_device_group)
+        id_alpha = ip_to_device_id[alpha.host_string]
+        id_bravo = ip_to_device_id[bravo.host_string]
+        logger.info("ID of alpha host: %s" % id_alpha)
+        logger.info("ID of bravo host: %s" % id_bravo)
 
-        ret = execute(Helpers.get_passive_partition, hosts=clients)
-        pass_part_alpha = ret[alpha]
-        pass_part_bravo = ret[bravo]
+        # TODO: paralize these using fabric.group.ThreadingGroup once we upgrade to Python 3
```
oleorhagen: Nitpick: `s/paralize/parallelize/`

tests/tests/test_os_ent_migration.py
```diff
@@ -101,7 +99,7 @@ def initialize_os_setup(env):
 
     # create and accept some devs; save tokens
     devs = []
-    for i in range(10):
+    for _ in range(10):
```
oleorhagen: :1st_place_medal:

tests/tests/test_preauth.py
```diff
@@ -222,32 +218,34 @@ def get_pub_key():
...
         while sleepsec < Client.MENDER_STORE_TIMEOUT:
             try:
-                out = run('strings {} | grep authtoken'.format(Client.MENDER_STORE))
+                out = device.run(
+                    "strings {} | grep authtoken".format(Client.MENDER_STORE)
+                )
                 return out != ''
             except:
-                output_from_journalctl = run("journalctl -u mender -l")
+                output_from_journalctl = device.run("journalctl -u mender -l")
```
oleorhagen: Note: When the renaming in `MEN-2948` is merged, this will not work anylonger. So let's be careful to rebase this if it is not merged before.

testutils/infra/device.py
```diff
@@ -0,0 +1,217 @@
...
+            self._devices.append(MenderDevice(host_string))
+
+    def __len__(self):
+        return len(self._devices)
+
+    def __getitem__(self, idx):
+        return self._devices[idx]
+
+    def run(self, cmd, **kw):
+        """Run command for all devices in group in a serialized manner
```
oleorhagen: Question: Like run them sequentially?

testutils/infra/device.py
```diff
@@ -0,0 +1,217 @@
...
+
+    fabric.api.local(
+        "%s %s %s/%s %s@%s:%s"
+        % (scp, port, local_path, file, fabric.api.env.user, host, remote_path)
+    )
+
+
+def _run(cmd, **kw):
+    if kw.get("wait") is not None:
+        wait = kw["wait"]
```
oleorhagen: Python scoping is weird

testutils/infra/device.py
```diff
@@ -0,0 +1,217 @@
+# Copyright 2020 Northern.tech AS
```
oleorhagen: Now this is beautiful!

testutils/infra/device.py
lluiscampos: That is a better term, thanks! I will amend here and in `put`

tests/tests/test_preauth.py
lluiscampos: Noted. Let's see who merges first :smile:

tests/tests/test_basic_integration.py
lluiscampos: This is a nice idea. I moved only the `ssh_is_opened` but looking at `Helpers` once again I see that absolutely everything is device specific.

Unless the other reviewers disagree, I will move all of it!

tests/tests/common_update.py
lluiscampos: Got it :+1:

testutils/infra/device.py
lluiscampos: > So much to like here!
> 
> Good job! 1st_place_medal raised_hands checkered_flag cherries
> 
> Only playing with some ideas, and some questions (:

Thank you. This is the most emojis I have ever received for a PR :muscle: 

And thanks for your comments. I locally fixed the small grammar stuff and I'll wait for one more review before doing the `Helpers` thing.

lluiscampos:
> So much to like here!
> 
> Good job! :1st_place_medal: :raised_hands: :checkered_flag: :cherries:
> 
> Only playing with some ideas, and some questions (:

Thank you @oleorhagen. This is the most emojis I have ever received for a PR :muscle:

And thanks for your comments. I locally fixed the small grammar stuff and I'll wait for one more review before doing the Helpers thing.

tests/tests/test_00_multi_tenancy.py
```diff
@@ -14,81 +14,53 @@
...
-                auth_v2.get_devices(expected_devices=0)  # make sure device not seen
-        else:
-            pytest.fail("failed to bootstrap device")
+        mender_device = MenderDevice(enterprise_no_client.get_mender_clients()[0])
+
+        mender_device.ssh_is_opened()
+        # TODO check this change with Kristian
+        mender_device.run(
+            'journalctl -u mender | grep "authentication request rejected server error message: Unauthorized"',
+            wait=70,
```
kacf: I *think* it should be ok!

tests/tests/test_grouping.py
```diff
@@ -105,48 +101,43 @@ def test_update_device_group(self, standard_setup_two_clients_bootstrapped):
...
+        ], Helpers.RebootDetector(bravo, host_ip) as reboot[bravo]:
+
+            deployment_id, expected_image_id = common_update_procedure(
+                conftest.get_valid_image(), devices=[id_alpha]
+            )
+
+            reboot[alpha].verify_reboot_performed()
+            # Extra long wait here, because a real update takes quite a lot
+            # of time.
+            reboot[bravo].verify_reboot_not_performed(300)
```
kacf: Now that these are not parallelized anymore, a small optimization you could try is to call these two in the opposite order. The thing is that `verify_reboot_not_performed` will always wait the entire interval, whereas `verify_reboot_performed` will progress while you're doing other things, so it should return immediately if you call it as number two and `verify_reboot_not_performed` has already waited for a long time.

tests/tests/test_basic_integration.py
oleorhagen: Also, I have a small request to make, and I feel now is probably a good time.

Could we change `update_image_successful` to simply `update_image`. Successful image updates is what we do for a living, right? The errors could be explicit instead.

Thoughts?

tests/tests/test_00_multi_tenancy.py
oleorhagen: I *think* journalctl can do grepping also, right?

tests/tests/test_grouping.py
lluiscampos: Good idea.

tests/tests/test_00_multi_tenancy.py
lluiscampos: Good eye @kacf, I actualy forgot about this TODO...

Let me elaborate and double check with you guys that this is okey. The previous code was grepping for "bootstrapped -> authorize-wait", which during my local run of this test didn't show up and I decided to change the text to grep for. That made me think that the old test logic was wrong, but but just reading the code it seemed to be okey. See the old logic [here](https://github.com/mendersoftware/integration/blob/beb63462cd1cd6e3772e709258486fd3d59697dc/tests/tests/test_00_multi_tenancy.py#L58-L64).

The test passes and I think it is safe to merge it, but it is still a mystery to me...

lluiscampos: (removing the TODO in the upcoming rework)

tests/tests/test_basic_integration.py
lluiscampos: I don't have an strong opinion, I will rename it. I am trying not to be too strict with these in the test framework, but I want to make you happy ;)

lluiscampos: > This is a nice idea. I moved only the `ssh_is_opened` but looking at `Helpers` once again I see that absolutely everything is device specific.

Not really everything. And also, I misunderstood your comment as was looking at `helpers.py` instead of `common_update.py`.

So I have moved many `helpers.py` methods to the `MenderDevice` class (see upcoming PR update) but nothing from `common_update.py`. The reason is that I don't want to make the `device.py` depend on MenderAPI module (basically because of it's global scope and the coupling with mender-artifact, etc).

Anyway, I think the changes I have made are nice and you will like it. And we can eventually make it even better after the MenderAPI refactor (QA-118).

lluiscampos:
Hello my dear reviewers!

The last push had minor typos and amends in the original commit, plus three more commits with improvements triggered by @oleorhagen comments.

If any, review only the new commits. We should be ready to merge if the tests pass.

Running together with https://github.com/mendersoftware/meta-mender/pull/943 in [Pipeline 125013155](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/125013155).

tests/tests/test_security.py
```diff
@@ -78,27 +75,11 @@ def test_token_token_expiration(self, standard_setup_with_short_lived_token):
...
-            if output.return_code == 0:
-                logger.info("mender logs indicate new authorization data available")
-                break
-
-        if timeout_time <= int(time.time()):
-            pytest.fail("timed out waiting for download retries")
-
+        mender_device = standard_setup_with_short_lived_token.device
+        mender_device.run(
+            'journalctl -u mender-client -l --no-pager | grep "received new authorization data"',
```
oleorhagen: Nitpick: 
`journalctl --grep="received new authorization data"` maybe?
That means the `-l` and `--no-pager` options should be no longer needed either.

tests/tests/test_update_modules.py
```diff
@@ -35,13 +31,7 @@ def test_file_update_module(self, standard_setup_one_docker_client_bootstrapped)
         """Test the file based update module, first with a failed update, then
         a successful one."""
 
-        mender_clients = standard_setup_one_docker_client_bootstrapped.get_mender_clients()
-
-        if not env.host_string:
```
oleorhagen: Not gonna miss this

testutils/infra/device.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class MenderDeviceGroup:
+    """Group of SSH accessible devices with Mender client
+
+    Currently, run/put methods are serialized, but they will be migrated to a parallel execution with Fabric 2
+    https://docs.fabfile.org/en/latest/api/group.html
+    """
+
+    def __init__(self, host_string_list, user="root"):
```
oleorhagen: Question: 
Why can't the device group take a list of `MenderDevices` instead of the `host_string_list`?

tests/tests/test_fault_tolerance.py
```diff
@@ -93,7 +127,7 @@ def test_deployed_during_network_outage(self, standard_setup_one_client_bootstra
 
         mender_device = standard_setup_one_client_bootstrapped.device
 
-        Helpers.gateway_connectivity(mender_device, False)
+        self.manipulate_network_connectivity(mender_device, False)
```
oleorhagen: Suggestion:
Since this is a static method, is not cleaner to call it with the class-name, instead of the object reference?

tests/tests/test_fault_tolerance.py
```diff
@@ -213,14 +212,14 @@ def test_image_download_retry_hosts_broken(self, standard_setup_one_client_boots
 
         mender_device = standard_setup_one_client_bootstrapped.device
 
-        inactive_part = Helpers.get_passive_partition(mender_device)
+        inactive_part = mender_device.get_passive_partition()
```
oleorhagen: :heart_eyes: :heart_eyes_cat: 
This is what I'm talking about!

oleorhagen: Semantically sensible test interfaces <3

testutils/infra/device.py
```diff
@@ -100,6 +102,135 @@ def ssh_is_opened(self, wait=60 * 60):
...
+    server = None
+    device = None
+    host_ip = None
+    # This global one is used to increment each port used.
+    port = 8181
+
+    def __init__(self, device, host_ip):
+        self.port = RebootDetector.port
+        RebootDetector.port += 1
+        self.host_ip = host_ip
```
oleorhagen: Question:
Why have a class variable, and an object variable named `host_ip` ?
Same goes for `device`

lluiscampos: Answer: it was like this already and I didn't dear to touch it at the time.

But yeah, I will fix it now :)

tests/tests/test_fault_tolerance.py
lluiscampos: Okey, it is the same to me, I will change it.

When the static method is from the same class, I usually use `self.`

testutils/infra/device.py
lluiscampos: It is a bit coupled to the way DockerNamespace and classes gather the IP address of the clients. It is actually quite ugly (as the mentioned methods append a harcoded `:8822`) so I might do a fix on this but not in this PR ;)

tests/tests/test_update_modules.py
lluiscampos: I will miss all the fun that we had to figure what this was about...

tests/tests/test_security.py
lluiscampos: I'll try this

lluiscampos: It seems like the journalctl compiled in yocto does not support this:
```
[172.21.0.5:8822] run: journalctl -u mender-client --grep="received new authorization data"
[172.21.0.5:8822] err: Compiled without pattern matching support
```

The call actually returns sucess, which was very annoying...

oleorhagen: hah, I see. Alright. Worth a shot.

tests/tests/test_update_modules.py
oleorhagen: I think this should be an interview question hereafter: "What does this piece of code do?"

lluiscampos: We can make a full quiz just with this test suite :rofl:

PR #839, opened 2020-03-09
Fix the name of the workflows-server container in enterprise composition

tranchitella:


lluiscampos:
Thank you for the fix :+1:

PR #842, opened 2020-03-09
Reuse docker-compose environments between tests

tranchitella:
This commit modifies the scope of the fixtures setting up the
docker-compose environment between test runs. Setting the scope
to session makes the tests reuse the environment between runs.

### DIFF
--- a/tests/common_setup.py
+++ b/tests/common_setup.py
@@ -24,7 +24,7 @@ from testutils.infra.container_manager import factory
 container_factory = factory.get_factory()
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="session")
 def standard_setup_one_client(request):
     env = container_factory.getStandardSetup(num_clients=1)
     request.addfinalizer(env.teardown)
@@ -39,7 +39,7 @@ def standard_setup_one_client(request):
     return env
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="session")
 def standard_setup_one_client_bootstrapped(request):
     env = container_factory.getStandardSetup(num_clients=1)
     request.addfinalizer(env.teardown)
@@ -55,7 +55,7 @@ def standard_setup_one_client_bootstrapped(request):
     return env
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="session")
 def standard_setup_one_rofs_client_bootstrapped(request):
     env = container_factory.getRofsClientSetup()
     request.addfinalizer(env.teardown)
@@ -71,7 +71,7 @@ def standard_setup_one_rofs_client_bootstrapped(request):
     return env
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="session")
 def standard_setup_one_docker_client_bootstrapped(request):
     env = container_factory.getDockerClientSetup()
     request.addfinalizer(env.teardown)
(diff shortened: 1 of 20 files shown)


tranchitella:
@kacf you are right; I've updated the PR to just run the tests for the most common setup, skipping all the others, and the test passed! (37 out of 61) Just wondering if I can re-work a bit the execution of the tests to run py.test multiple times, each of them for a specific setup/fixture. Giving it a try...

tranchitella:
@lluiscampos @kacf I'm very curious about your feedback here.
I've updated the scope of the setup fixtures to session, marked all the tests with a setup-specific name and I'm running pytest multiple times in a loop, once for each marker to share the setup. All the tests are passing. :) I'd be curious to increase the number of parallel tests, though.

tests/run.sh
```diff
@@ -169,10 +169,26 @@ if [[ -n $SPECIFIC_INTEGRATION_TEST ]]; then
...
+standard_setup_one_rofs_client_bootstrapped
+standard_setup_one_docker_client_bootstrapped
+standard_setup_two_clients_bootstrapped
+setup_with_legacy_client
+standard_setup_with_signed_artifact_client
+standard_setup_with_short_lived_token
+setup_failover
+running_custom_production_setup
+enterprise_no_client
+migrated_enterprise_setup"
```
kacf: Could we rework this list to be centrally managed in some way? It's quite easy to miss an update here, which would reduce our test coverage. One suggestion is to put this in a list in some standalone python file, which the test framework loads and adds as markers (instead of the `pytest.ini` list). A different script can then simply print the list, which would make it available for iteration here.

tests/tests/test_00_multi_tenancy.py
```diff
@@ -27,6 +28,7 @@
 
 
 class TestMultiTenancyEnterprise(MenderTesting):
+    @pytest.mark.enterprise_no_client
```
kacf: What happens to tests that aren't marked?

Probably they simply won't run ATM, but we can easily add an auto-fixture to make sure it is present. You can check out [this fixture](https://github.com/mendersoftware/meta-mender/blob/bf4be1f27d8929e9fdd706aa29559ef71f6ebe38/tests/acceptance/fixtures.py#L404) for an example of how to do it.

tests/run.sh
tranchitella: @kacf sure, this is a PoC, quick approach to have everything up and running

tests/tests/test_00_multi_tenancy.py
tranchitella: @kacf if they are not marked, the tests won't run. It is easy to run those tests using an expression, e.g.:
`-m 'not setup_failover and not running_custom_production_setup  and ...`
If we manage the markers in a python file, we can easily produce such a string to run "non-marked" tests.

kacf: But would there ever be a test should not be marked? They all will depend on at least one fixture.

tests/common_setup.py
```diff
@@ -215,14 +204,3 @@ def enterprise_no_client(request):
     reset_mender_api(env)
 
     return env
-
-
-@pytest.fixture(scope="function")
-def enterprise_no_client_smtp(request):
-    env = container_factory.getEnterpriseSMTPSetup()
```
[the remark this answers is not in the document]
tranchitella: Done

tranchitella:
@lluiscampos @kacf I've summarised my findings in Jira:
https://tracker.mender.io/browse/QA-156

lluiscampos:
We will follow-up in another JIRA ticket. Closing.

PR #845, opened 2020-03-11
removed leftover e2e tests, since they now live in the gui repo

mzedel:
Changelog: None


lluiscampos:
And thanks for the clean-up!

PR #846, opened 2020-03-11
Format all integration and testutils Python code with black formatter

lluiscampos:
Format all integration and testutils Python code with black formatter and start enforcing it in the CI Pipeline.

Even though it is still running with Python 2, it should be okey to
use Python 3 style.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,15 @@
 
 [tool.black]
 target-version = ['py38']
-include = '.*/backend-tests/.*\.py$'
+include = '''
+(
+  /(
+      backend-tests
+    | testutils
+    | tests
+  )/.*\.py$
+)
+'''
 exclude = '''
 (
   /(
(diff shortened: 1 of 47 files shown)


pyproject.toml
```diff
@@ -1,7 +1,15 @@
...
-include = '.*/backend-tests/.*\.py$'
+include = '''
+(
+  /(
+      testutils/.*\.py$
+    | backend-tests/.*\.py$
+    | tests/.*\.py$
+  )/
+)
+'''
```
lluiscampos: @oleorhagen Please pay close attention to this regex. I have low confidence in that I got it right

oleorhagen: Not completely right I am afraid, so good instincts :)
Try this:
```python
include = '''
(
  /(
      backend-tests
    | testutils
    | tests
  )/.*\.py$
)
```

lluiscampos: That works, thanks!

lluiscampos:
> I'm gonna approve in spite of the comment, just remember to fix it :)

I am pushing now the amendment and I will be merging without running tests again, as they already passed in the previous push.

Thanks!

PR #849, opened 2020-03-16
QA-170: Move backend-tests files knowledge to integration repo

lluiscampos:
This commit introduces a major rework in the ./run script so that the
knowledge of which fails goes with which suite (open/enterprise) is in
this repo rather than in the caller (i.e. mender-qa, which is not
versioned).

The motivation is to avoid once the recurrent problem of breaking old
releases when introducing/removing files

Now the wrapper takes a -s|--suite parameter that can be open (default),
enterprise or all and it takes care of the files required in each case.
The -f parameter to specify files is now deprecated

As a side effect, we have also moved the results pytest args here, so
that we can control that the "all" case does not override results.

It still accept user custom filters or other pytest parameters.

### DIFF
--- a/backend-tests/run
+++ b/backend-tests/run
@@ -5,72 +5,87 @@ HERE=$(dirname $(readlink -f "$0"))
 INTEGRATION_PATH=$(dirname "$HERE")
 export INTEGRATION_PATH
 
-COMPOSE_CMD="docker-compose -p backend-tests \
-             -f $INTEGRATION_PATH/docker-compose.yml \
-             -f $INTEGRATION_PATH/docker-compose.demo.yml \
-             -f $INTEGRATION_PATH/backend-tests/docker/docker-compose.backend-tests.yml"
+COMPOSE_CMD_OPEN="docker-compose -p backend-tests \
+        -f $INTEGRATION_PATH/docker-compose.yml \
+        -f $INTEGRATION_PATH/docker-compose.demo.yml \
+        -f $INTEGRATION_PATH/backend-tests/docker/docker-compose.backend-tests.yml \
+        -f $INTEGRATION_PATH/docker-compose.storage.minio.yml
+        "
+COMPOSE_CMD_ENTERPRISE="${COMPOSE_CMD_OPEN} -f $INTEGRATION_PATH/docker-compose.enterprise.yml \
+        -f $INTEGRATION_PATH/extra/recaptcha-testing/tenantadm-test-recaptcha-conf.yml \
+        -f $INTEGRATION_PATH/extra/smtp-testing/conductor-workers-smtp-test.yml \
+        -f $INTEGRATION_PATH/extra/stripe-testing/stripe-test.docker-compose.yml
+        "
+COMPOSE_CMD=""
 
-# by default just add minio, with COMPOSE_CMD_BASE this creates the standard onprem ST setup
-COMPOSE_FILES_DEFAULT=( "$INTEGRATION_PATH/docker-compose.storage.minio.yml" )
-COMPOSE_FILES=()
+PYTEST_FILTER_OPEN="not Enterprise and not Multitenant"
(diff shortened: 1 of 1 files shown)


lluiscampos:
Testing together with https://github.com/mendersoftware/mender-qa/pull/378 in [Pipeline 126828746](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/126828746)

lluiscampos:
I will not merge this until I don't have the same PR (modifying the files list, of course) for integration/2.3.x and integration/2.2.x.

Looking forward to your feedback

lluiscampos:
> Testing together with [mendersoftware/mender-qa#378](https://github.com/mendersoftware/mender-qa/pull/378) in [Pipeline 126828746](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/126828746)

Testing again in [Pipeline 126930954](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/126930954).

backend-tests/run
```diff
@@ -5,73 +5,91 @@ HERE=$(dirname $(readlink -f "$0"))
...
-COMPOSE_FILES_DEFAULT=( "$INTEGRATION_PATH/docker-compose.storage.minio.yml" )
-COMPOSE_FILES=()
+COMPOSE_CMD_OPEN="docker-compose -p backend-tests \
+        -f $INTEGRATION_PATH/docker-compose.yml \
+        -f $INTEGRATION_PATH/docker-compose.demo.yml \
+        -f $INTEGRATION_PATH/backend-tests/docker/docker-compose.backend-tests.yml \
+        -f $INTEGRATION_PATH/docker-compose.storage.minio.yml"
+COMPOSE_CMD_ENTERPRISE="${COMPOSE_CMD_OPEN} -f $INTEGRATION_PATH/docker-compose.enterprise.yml \
+        -f $INTEGRATION_PATH/extra/recaptcha-testing/tenantadm-test-recaptcha-conf.yml \
+        -f $INTEGRATION_PATH/extra/smtp-testing/conductor-workers-smtp-test.yml \
```
kacf: No idea how this file is used, but did we not just remove conductor?

lluiscampos: Good question ;)

So basically at the time of removing conductor, we wanted to rename this file (i.e. workflows-workers...) but then we had no mechanism in mender-qa to make in backwards compatible. The only solution was an ugly if this file exist...

That was the original motivation for this task. At the moment we cannot run 2.3.x backend-tests so that has ben the "second" motivation to do it.

kacf: Ah, but then you can rename it now, right?

lluiscampos: I guess so.

But let me first merge this one (and do similar fixes in 2.3.x and 2.2.x); then separately I will ping Fabio to see if in master the file can completely be removed or only renamed.

lluiscampos:
Check for `pytest-html` from `./run` script proved worthless, as the python dependencies are installed inside the container. I reworked then the logic to create reports unconditionally.

Running again in https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/127003010

backend-tests/run
tranchitella: @lluiscampos @kacf I confirm we can rename the file. I couldn't rename it because, as Lluis pointed out, it would have broken the tests for 2.3.x.

lluiscampos:
Thanks @tranchitella for the review.

The last force-pushed only fixed typos an newlines. Merging!

lluiscampos:
See also #850 and #851

backend-tests/run
lluiscampos: :+1: Feel free to rename it in master next time you are around that.

 Funny enough, the tests were broken for 2.3.x already when we introduced the `stripe-test.docker-compose.yml`. This will be fixed in #850

PR #851, opened 2020-03-17
QA-170: move backend knowledge tests files to integration - 2.2.x

lluiscampos:
Cherry-picking #849 (and other minor fixes) plus removing `stripe-test.docker-compose.yml` only existing in master.


lluiscampos:
Integration tests are failing. However this PR only affects backend-tests, so it is safe to merge.

PR #852, opened 2020-03-17
New back-end tests for artifact depends/provides

tranchitella:
Depends on: https://github.com/mendersoftware/deployments-enterprise/pull/79

### DIFF
--- a/backend-tests/docker/Dockerfile
+++ b/backend-tests/docker/Dockerfile
@@ -7,5 +7,7 @@ RUN apt-get -y -qq update && apt-get -qq -y install \
 COPY requirements-py3.txt .
 
 RUN pip3 install -r requirements-py3.txt
+COPY mender-artifact /usr/local/bin/mender-artifact
+RUN chmod 755 /usr/local/bin/mender-artifact
 
 ENTRYPOINT ["bash", "/tests/run.sh"]
--- a/backend-tests/tests/test_artifact.py
+++ b/backend-tests/tests/test_artifact.py
@@ -1,0 +1,500 @@
+# Copyright 2020 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        https://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import os
+import random
+import re
+import requests
+import string
+import subprocess
+import tempfile
+
+from contextlib import contextmanager
+
+from testutils.api.client import ApiClient
+from testutils.api import (
+    deployments,
+    useradm,
+    deviceauth as deviceauth_v1,
+)
+from testutils.common import create_org, mongo, clean_mongo
+
+from .test_deployments import make_accepted_device
+
+
+class TestUploadArtifactEnterprise:
+    def get_auth_token(self, username, password):
+        r = ApiClient(useradm.URL_MGMT).call(
+            "POST", useradm.URL_LOGIN, auth=(username, password)
+        )
+        assert r.status_code == 200
+        assert r.text is not None
+        assert r.text != ""
+
+        return r.text
+
+    def get_tenant_username_and_password(self):
+        tenant, username, password = (
+            "test.mender.io",
+            "some.user@example.com",
+            "secretsecret",
+        )
+        tenant = create_org(tenant, username, password)
+        return tenant, username, password
+
+    @contextmanager
+    def get_mender_artifact(
+        self,
+        artifact_name="test",
+        update_module="dummy",
+        device_types=("arm1",),
+        size=256,
+        depends=(),
+        provides=(),
+    ):
+        data = "".join(random.choices(string.ascii_uppercase + string.digits, k=size))
+        f = tempfile.NamedTemporaryFile(delete=False)
+        f.write(data.encode("utf-8"))
+        f.close()
+        #
+        filename = f.name
+        artifact = "%s.mender" % filename
+        args = [
+            "mender-artifact",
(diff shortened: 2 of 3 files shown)


backend-tests/docker/Dockerfile
```diff
@@ -7,5 +7,6 @@ RUN apt-get -y -qq update && apt-get -qq -y install \
 COPY requirements-py3.txt .
 
 RUN pip3 install -r requirements-py3.txt
+RUN wget -q https://d1b0l86ne08fsf.cloudfront.net/mender-artifact/master/linux/mender-artifact -O /usr/local/bin/mender-artifact && chmod 755 /usr/local/bin/mender-artifact
```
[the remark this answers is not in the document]
tranchitella: Ack!

backend-tests/tests/test_artifact.py
```diff
@@ -0,0 +1,352 @@
...
+        filename = f.name
+        artifact = "%s.mender" % filename
+        args = [
+            "mender-artifact",
+            "write",
+            "module-image",
+            "-o",
+            artifact,
+            "--version",
+            "3",
```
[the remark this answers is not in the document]
tranchitella: Ack!

backend-tests/tests/test_artifact.py
```diff
@@ -0,0 +1,352 @@
...
+                files=(
+                    ("description", (None, "description")),
+                    ("size", (None, str(os.path.getsize(artifact)))),
+                    (
+                        "artifact",
+                        (artifact, open(artifact, "rb"), "application/octet-stream"),
+                    ),
+                ),
+            )
+        assert r.status_code != 201
```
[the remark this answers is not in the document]
tranchitella: This is a bug in @alfrunes PR, it returns 500, he is fixing it.
I'm setting up the right status_code as soon as his PR is updated.
For the very same reason, I'm NOT running pipeline until his PR is okay.

tranchitella:
> Oh, and one more thing, this should be targeted towards the `artifact-depends-i4` branch, which is the feature branch for this epic.

Oh good, I'll take care of this (which I suppose, it means I have to close and open a new PR).

backend-tests/tests/test_artifact.py
```diff
@@ -0,0 +1,352 @@
...
+            "key2": "value2",
+        }
+        assert artifact["artifact_provides"] == {
+            "artifact_name": "test",
+            "key3": "value3",
+            "key4": "value4",
+            "key5": "value5",
+        }
+
+    def test_upload_artifact_depends_provides_conflicting(self, mongo, clean_mongo):
```
alfrunes: Nitpick (test naming): Provides cannot conflict on upload and can (in theory) be identical for different artifacts.

backend-tests/tests/test_artifact.py
```diff
@@ -0,0 +1,352 @@
...
+        api_client.headers = {}  # avoid default Content-Type: application/json
+        api_client.with_auth(auth_token)
+
+        # create and upload the mender artifact
+        for checksum in checksums:
+            with self.get_mender_artifact(
+                artifact_name="test",
+                device_types=["arm1"],
+                depends=("rootfs_image_checksum:%s" % checksum,),
+                provides=("rootfs_image_checksum:provided",),
```
alfrunes: I might be wrong, but I think `mender-artfiact` also provides this value with the actual checksum by default.

alfrunes: But if the following tests passes, then perhaps this value gets overwritten by this flag.

tranchitella: Yep, the tests pass, let's keep it as it is

backend-tests/tests/test_artifact.py
tranchitella: Renamed to: test_upload_artifact_provides_conflicting

backend-tests/tests/test_artifact.py
```diff
@@ -0,0 +1,352 @@
...
+            "3",
+            "--artifact-name",
+            artifact_name,
+            "-T",
+            update_module,
+            "-f",
+            filename,
+        ]
+        for device_type in device_types:
+            args.extend(["--t", device_type])
```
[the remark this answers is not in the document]
tranchitella: Replaced with -t, thanks!

backend-tests/tests/test_artifact.py
alfrunes: Ohh, I see my comment wasn't very clear (sorry about that). I meant something like `test_upload_artifact_depends_conflicting`.

backend-tests/tests/test_artifact.py
kacf: It only does when you use `write rootfs-image`, not when using `write module-image`. So for test purposes, the code is correct.

backend-tests/tests/test_artifact.py
```diff
@@ -0,0 +1,352 @@
...
+            deployments.URL_NEXT,
+            body={
+                "device_type": "arm1",
+                "artifact_name": "old-artifact",
+                "rootfs_image_checksum": "checksum",
+            },
+        )
+        assert r.status_code == 200
+
+    def test_upload_artifact_selection_match_depends_multiple_artifacts(
```
alfrunes: It would be useful to have a testcase where the provides satisfies multiple artifacts and verify that the artifact with the smallest file-size is picked. For instance, make a deployment with artifact depends:
```
{
    "device_type": ["acme"],
    "vcdiff": "xdelta"
} (artifact_name: "common", size 100KiB)
```
```
{
    "device_type": ["acme"]
} (artifact_name: "common", size: 256MiB)
```
And a POST /next request with provides
```
{
    "artifact_name": "foobar",
    "device_type": "acme",
    "vcdiff": "xdelta"
}
```
And check the first is picked.

backend-tests/tests/test_artifact.py
tranchitella: my fault, fixed

backend-tests/tests/test_artifact.py
tranchitella: Done!

tranchitella:
Manually started PR with deployments-enterprise/79:
https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/127774702

PR #853, opened 2020-03-19
Move test to its own file and optimize running time

kacf:


kacf:
This goes together with https://github.com/mendersoftware/mender/pull/513.

tests/tests/test_provides_depends.py
```diff
@@ -0,0 +1,112 @@
...
+
+from ..common_setup import enterprise_no_client
+from .common_update import update_image, common_update_procedure
+from .mendertesting import MenderTesting
+from ..MenderAPI import auth, auth_v2, deploy, image, logger
+
+from testutils.infra.device import MenderDevice
+
+
+class TestProvidesDependsEnterprise(MenderTesting):
```
lluiscampos: Hehehe. I was reviewing commit by commit so I added a comment asking for this test to be moved to its own file adding the Enterprise suffix... Good :+1:

tests/tests/test_provides_depends.py
```diff
@@ -0,0 +1,112 @@
...
+    def test_update_provides_depends(self, enterprise_no_client):
+        """
+        Perform two consecutive updates, the first adds virtual provides
+        to the artifact and the second artifact depends on these provides.
+        """
+
+        DEMO_POLL_INTERVAL = 5
+        IMAGE_NAME = "core-image-full-cmdline-qemux86-64.ext4"
```
lluiscampos: Please use `conftest.get_valid_image()` so that we avoid copies of this.

tests/tests/test_provides_depends.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # Create tenant user
+        auth.reset_auth_token()
+        auth.new_tenant("admin", "bob@builder.org", "secret-service")
+        token = auth.current_tenant["tenant_token"]
+
+        # Create client setup with tenant token
+        enterprise_no_client.new_tenant_docker_client("mender-client", token)
+        mender_device = MenderDevice(enterprise_no_client.get_mender_clients()[0])
+        mender_device.host_ip = enterprise_no_client.get_virtual_network_host_ip()
```
lluiscampos: I think `host_ip` does not belong go `mender_device`. That is why when I introduced this in all other tests, I always kept it independently. Better to be consistent with what we have in other tests.

tests/tests/test_provides_depends.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # Wait for ssh to be open
+        mender_device.ssh_is_opened()
+        # Check that the device has authorized with the backend.
+        device = auth_v2.get_devices(expected_devices=1)
+        device_ids = [device[0]["id"]]
```
lluiscampos: Unused

tests/tests/test_provides_depends.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # Create client setup with tenant token
+        enterprise_no_client.new_tenant_docker_client("mender-client", token)
+        mender_device = MenderDevice(enterprise_no_client.get_mender_clients()[0])
+        mender_device.host_ip = enterprise_no_client.get_virtual_network_host_ip()
+
+        # Wait for ssh to be open
+        mender_device.ssh_is_opened()
+        # Check that the device has authorized with the backend.
+        device = auth_v2.get_devices(expected_devices=1)
```
lluiscampos: `device` is unused also, consider just calling the function but ignoring return

tests/tests/test_provides_depends.py
```diff
@@ -0,0 +1,112 @@
...
+        device_ids = [device[0]["id"]]
+        auth_v2.accept_devices(1)
+        assert len(auth_v2.get_devices_status("accepted")) == 1
+
+        # Update client with and artifact with custom provides
+        def prepare_provides_artifact(artifact_file, artifact_id):
+            cmd = (
+                # Package MenderAPI in the artifact, just a random folder.
```
lluiscampos: Please use a different folder, as we have plans in QA backlog to remove this one. Maybe `/testutils` which will survive my refactoring plans or just `/tests/tests`

tests/tests/test_provides_depends.py
```diff
@@ -0,0 +1,112 @@
...
+            make_artifact=prepare_depends_artifact, verify_status=False
+        )
+
+        # Retry for at most 60 seconds checking for deployment status update
+        stat = None
+        for i in range(60):
+            time.sleep(1)
+            stat = deploy.get_statistics(deployment_id)
+            if stat.get("pending") == 0:
+                break
```
lluiscampos: For my own education: so there the transition will go "pending" -> "noartifact" without going through "inprogress", right?

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -373,6 +377,16 @@ def new_tenant_client(self, name, tenant):
...
+    def new_tenant_docker_client(self, name, tenant):
+        if not self.MT_DOCKER_CLIENT_FILES[0] in self.docker_compose_files:
+            self.extra_files += self.MT_DOCKER_CLIENT_FILES
+        logger.info("creating docker client connected to tenant: " + tenant)
+        self._docker_compose_cmd(
+            "run -d --name=%s_%s mender-client" % (self.name, name),
+            env={"TENANT_TOKEN": "%s" % tenant},
+        )
+        time.sleep(5)
+
```
lluiscampos: Nice!

tests/tests/test_provides_depends.py
kacf: But this is the IP of the device, no? Not sure if I understand the reason.

tests/tests/test_provides_depends.py
kacf: Correct. "inprogress" is only triggered when there is a download, which there isn't 
when "noartifact" is triggered.

kacf:
All issues fixed except the `host_ip` comment, which I asked about.

tests/tests/test_provides_depends.py
lluiscampos: No, is the IP of the test runner (i.e. host running docker-compose) as seen from the device (i.e. gateway). At the time, there was a naming discussion around `get_virtual_network_host_ip` due to this.

lluiscampos: The IP of the device(s) comes from `get_mender_clients()`

alfrunes: The `host_ip` is just a variable with the client's host container IP that I added to the `mender_device` object as a matter of keeping the variable associated with the device. I don't think it's defined anywhere else, and might as well be defined as it's own variable.

kacf: I see, thanks for explaining. Fixed!

tests/tests/test_provides_depends.py
```diff
@@ -0,0 +1,112 @@
...
+                "directory-artifact-gen -o %s -n %s -t docker-client -d /tmp/test_file_update_module tests -- --provides foo:bar"
+                % (artifact_file, artifact_id)
+            )
+            logger.info("Executing: " + cmd)
+            subprocess.check_call(cmd, shell=True)
+            return artifact_file
+
+        deployment_id, _ = common_update_procedure(
+            mender_device,
+            host_ip,
```
lluiscampos: Method `common_update_procedure` doesn't take any of these (no positional arguments).

It works due to an implementation detail: `make_artifact` being set makes the function ignore `install_image` (`mender_device` here). But please clean-up.

lluiscampos: Also repeated once more in this test below.

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -373,6 +377,16 @@ def new_tenant_client(self, name, tenant):
...
 
+    def new_tenant_docker_client(self, name, tenant):
+        if not self.MT_DOCKER_CLIENT_FILES[0] in self.docker_compose_files:
+            self.extra_files += self.MT_DOCKER_CLIENT_FILES
+        logger.info("creating docker client connected to tenant: " + tenant)
+        self._docker_compose_cmd(
+            "run -d --name=%s_%s mender-client" % (self.name, name),
+            env={"TENANT_TOKEN": "%s" % tenant},
+        )
+        time.sleep(5)
```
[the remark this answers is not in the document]
oleorhagen: I like this idea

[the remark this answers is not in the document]
oleorhagen: +1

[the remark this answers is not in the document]
kacf: Does it support `docker-compose` though?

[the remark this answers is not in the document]
lluiscampos: I have been looking at the SDK over the weekend. It looks pretty nice and we should use it in `docker_manager.py`, but (answering @kacf question) it does not abstract `docker-compose`, so for `docker_compose_manager.py` it will get harder. It can be workaround it (filtering from the running containers according to the project name) but I don't think it will result in cleaner code.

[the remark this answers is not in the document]
alfrunes: Unfortunately not.

tests/tests/test_provides_depends.py
kacf: Thanks, fixed!

PR #854, opened 2020-03-19
QA-132: Migrage integration tests suite to Python 3

lluiscampos:
Best reviewed commit-by-commit


lluiscampos:
It can be reviewed already even though most likely will still require some fixes here and there.

lluiscampos:
> It can be reviewed already even though most likely will still require some fixes here and there.

Like the black formatting fix that I just force-pushed ;)

tests/conftest.py
```diff
@@ -140,25 +142,41 @@ def pytest_exception_interact(node, call, report):
...
+            if isinstance(val, BaseContainerManagerNamespace)
+        ]
+        if len(env_candidates) == 1:
+            env = env_candidates[0]
+            dev_candidates = [
+                getattr(env, attr)
+                for attr in dir(env)
+                if isinstance(getattr(env, attr), MenderDevice)
+                or isinstance(getattr(env, attr), MenderDeviceGroup)
+            ]
```
kacf: Impressive!

tests/README.md
```diff
@@ -66,7 +66,7 @@ $ virtualenv <name-of-virtualenv-folder>
 
 ##### (Optional) -- Select Python Version
 ```bash
-$ virtualenv -p /usr/bin/python2.7 <name-of-virtualenv-folder>
+$ virtualenv -p /usr/bin/python3.7 <name-of-virtualenv-folder>
```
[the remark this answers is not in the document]
lluiscampos: Thanks. I'll fix.

testutils/infra/device.py
```diff
@@ -302,47 +301,68 @@ def _ssh_prep_args_impl(tool):
...
+        # are quite expensive.
+        time.sleep(sleeptime)
+        sleeptime *= 2
+
+        try:
+            result = conn.run(cmd, **kw)
+            break
+        except NoValidConnectionsError as e:
+            logger.info("Could not connect to host %s: %s", conn.host, e)
+            time.sleep(30)
```
kacf: Why do you need an explicit sleep here when you already have `time.sleep(sleeptime)`? This will probably introduce unnecessary waiting time.

testutils/infra/device.py
```diff
@@ -302,47 +301,68 @@ def _ssh_prep_args_impl(tool):
...
-            finally:
-                # Taken from disconnect_all() in Fabric.
-                from fabric.state import connections
-
-                if connections.get(fabric.api.env.host_string) is not None:
-                    connections[fabric.api.env.host_string].close()
-                    del connections[fabric.api.env.host_string]
-
-    return output
+    while True:
```
kacf: I can't see that this loop handles timing out, which the old one did. Should you perhaps check absolute time at the very beginning of the loop, before waiting?

lluiscampos: Yep, I realized this morning when I got the first tests running in CI. I have fixed it locally and will push soon. Thanks for a good finding!

testutils/infra/device.py
lluiscampos: Good point. It was a merge of two similar functions that resulted in double wait. I'll fix.

testutils/infra/device.py
lluiscampos: Pushed now, it has an extra handling for this:

```
        except UnexpectedExit as e:
            # Many tests rely on the old behaviour of "keep trying until it passes", so
            # they run commands that may return non 0 for a while
            logger.info(
                "Got UnexpectedExit while executing command in host %s: %s", conn.host, e
            )
            continue
```

It is not consistent across tests and I would love to clean-it up (many tests use "default" timeout which is ONE HOUR!). But I think I shouldn't spend too much time on this now so I added some TODOs and will address it in the future.

lluiscampos:
Hi guys and thank you for your review.

I had an stupid error in the `.run.sh` script regarding the pytest-xdist dependency check which resulted with the test run stopping first failure. So I had learned little about tests with still missing changes.

Giving now a second go and hoping for a final rework this afternoon.

testutils/infra/device.py
kacf: > many tests use "default" timeout which is ONE HOUR!

Yes, this comes from the times we used Digital Ocean. It was not that uncommon that certain tests could take up to 40 min to pass. The capacity we got in Digital Ocean was absolutely abysmal. Several times I measured a steal time (time that hypervisor "steals" from VM) of **85%**! I think the conditions in Google Cloud are much better so maybe we could reduce this to something like 20 min or something. You could perhaps pick the longest test according to the html summary and multiply it by three or something. That should give a pretty safe fallback timeout, even when conditions are bad.

lluiscampos:
Finally, all tests passed :tada:  :)

Please check the last commit fe2f301 which contains the individual tests fixes. I will squash it before merge.

tests/tests/test_fault_tolerance.py
```diff
@@ -257,10 +248,6 @@ def test_rootfs_conf_missing_from_new_update(
 
         mender_device = standard_setup_one_client_bootstrapped.device
 
-        output = mender_device.run("test -e /data/mender/mender.conf", hide=True)
-        if output.return_code != 0:
-            pytest.skip("Needs split mender.conf configuration to run this test")
```
kacf: This should probably be retained with an `&& echo true` so you can check the output. Otherwise we can't run this with thud and older.

lluiscampos:
@oleorhagen (Your comment has disappeared after my push). Look at [this formatting](https://github.com/mendersoftware/integration/pull/854/files#diff-bae596dcd07996767e2e1cb71ac5aac9R641-R647), bra nok?

lluiscampos:
My last push contain the last comments from you guys plus a rebase from master. Once tests run and pass once more, I think I can merge it.

tests/README.md
```diff
@@ -66,7 +66,7 @@ $ virtualenv <name-of-virtualenv-folder>
 
 ##### (Optional) -- Select Python Version
 ```bash
-$ virtualenv -p /usr/bin/python2.7 <name-of-virtualenv-folder>
+$ python -m venv <name-of-virtualenv-folder>
```
oleorhagen: Minor nitpick:
Should probs be `python3` on most systems, no?

lluiscampos: Hopefully this will be changing this year, right? Ubuntu had this "python2" package which removes the "python" symlink so that it can be replaced by python3. I don't think it is very important, any educated reader know what are we referring to...

oleorhagen: Alright, good to go then :smiley_cat:

PR #855, opened 2020-03-20
Backend integration tests for v2/filters/search endpoint

alfrunes:
Enterprise integration tests for v2/filters/search endpoint

changelog: none

### DIFF
--- a/backend-tests/tests/test_inventory.py
+++ b/backend-tests/tests/test_inventory.py
@@ -11,6 +11,8 @@
 #    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
+import json
+import logging
 import pytest
 import random
 import time
@@ -22,6 +24,7 @@ import testutils.api.deviceauth as deviceauth_v1
 import testutils.api.deviceauth_v2 as deviceauth_v2
 import testutils.api.useradm as useradm
 import testutils.api.inventory as inventory
+import testutils.api.inventory_v2 as inventory_v2
 import testutils.api.tenantadm as tenantadm
 import testutils.util.crypto
 from testutils.common import (
@@ -37,7 +40,7 @@ from testutils.common import (
 )
 
 
-@pytest.yield_fixture(scope="function")
+@pytest.fixture(scope="function")
 def clean_migrated_mongo(clean_mongo):
     deviceauth_cli = CliDeviceauth()
     useradm_cli = CliUseradm()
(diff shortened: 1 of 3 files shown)


alfrunes:
Added tests for the internal endpoint too.

backend-tests/tests/test_inventory.py
```diff
@@ -37,7 +40,7 @@
 )
 
 
-@pytest.yield_fixture(scope="function")
+@pytest.fixture(scope="function")
```
mchalski: this is because we bumped pytest/python itself?

backend-tests/tests/test_inventory.py
```diff
@@ -311,3 +315,579 @@ def test_fail_no_attr_value(self, user):
...
+                            and dev.inventory["version"] in ["v1.0", "v3.0"],
+                            tenant.devices,
+                        ),
+                        key=lambda dev: dev.inventory["idx"],
+                        reverse=True,
+                    )
+                ],
+            },
+            {
+                "name": "Test $nin, sort by descending idx",
```
mchalski: since we're comprehensive - could use a `$ne` test as well

backend-tests/tests/test_inventory.py
```diff
@@ -311,3 +315,579 @@ def test_fail_no_attr_value(self, user):
...
+        attr_list = []
+        for key, value in d.items():
+            attr = {"name": key, "value": value}
+            if scope is not None:
+                attr["scope"] = scope
+            attr_list.append(attr)
+
+        return attr_list
+
+    def test_search_v2(self, tenants_users):
```
mchalski: I meant to say it would be useful to test the 'plan-too-low' scenario ("os").
but actually tenants_users does create tenants with this plan.

so something is wrong - they should not be able to use the filtering api, it's PRO and ENT only. I believe KJ submitted some late patches yesterday, fixing just that (used to allow "os").

mchalski: https://github.com/mendersoftware/inventory-enterprise/pull/13

mchalski:
@alfrunes one general CR - we're doing this on dev branches now.
in `inventory`, the dev branch is called `saved-filters` (double check pls).
can we have the same branch here as well? then it's easy to identify a feature across repos.

backend-tests/tests/test_inventory.py
alfrunes: yeah, I noticed in the API docs that the `yield_fixture` is deprecated - the two decorators are the same now, but it might get removed in the future.

backend-tests/tests/test_inventory.py
alfrunes: Good point. I tested it before this was merged, I'll make a test-case for it.

alfrunes:
@mchalski I've addressed your comments and started a new pipeline https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/129203445

PR #857, opened 2020-03-23
release_tool: Make it possible to use branch names in --pr argument.

kacf:
Changelog: None


kacf:
Cancelling mender-qa pipeline. This code path is not used there.

PR #860, opened 2020-03-26
Revert "MEN-3279 RBAC do not clean rbac db"

merlin-northern:
This reverts commit 9242b848b6afae6f0d379159fae4b7f923429d30.

We decided to change the behaviour and move the roles to
per-tenant dbs


tranchitella:
@merlin-northern just add the Changelog in the commit message!

PR #865, opened 2020-04-02
Men 3310: Integrartion test for the retry-failed deployments functionality

oleorhagen:
Some misc nitpicks, and the integration test commit.

The test currently only tests the successful case, with a retry count of 3. My thought was, that the failure case is covered in the acceptance test, but if you want it covered here also, then it should not be much work to add more cases, with different number of retries, and a failure case.

-O


oleorhagen:
Last pipeline (https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/132490524) passed on the added test, but failed on one other integration test. This test has not failed on any of the other runs, and I am thus treating it as spurious.

Therefore merging.

PR #866, opened 2020-04-02
tests: Take into account multiple mender.service names.

kacf:
Due to integration needing to be tested with meta-mender branches
warrior and older, we need to be prepared for the client being named
both "mender.service" (the old name), and "mender-client.service" (the
new name).

Changelog: None

### DIFF
--- a/testutils/infra/device.py
+++ b/testutils/infra/device.py
@@ -126,6 +126,8 @@ class MenderDevice:
     def get_reboot_detector(self, host_ip):
         return RebootDetector(self, host_ip)
 
+    def get_client_service_name(self):
+        return self.run("if test -e /lib/systemd/system/mender.service; then echo mender; else echo mender-client; fi").strip()
 
 class RebootDetector:
     # This global one is used to increment each port used.
@@ -271,6 +273,12 @@ class MenderDeviceGroup:
         """
         for dev in self._devices:
             dev.ssh_is_opened(wait)
+
+    def get_client_service_name(self):
+        # We assume that the service name is always the same across all devices,
+        # so it's enough to return the first one.
+        assert len(self._devices) > 0
+        return self._devices[0].get_client_service_name()
 
 
 def _ssh_prep_args(device):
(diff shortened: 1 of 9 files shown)


kacf:
Note that this needs to pass both of these pipeline (it won't show up correctly in Github status; one will shadow the other):

* [With meta-mender/master](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/132177744)
* [With meta-mender/warrior](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/132180822)

testutils/infra/device.py
```diff
@@ -126,6 +126,9 @@ def get_passive_partition(self):
     def get_reboot_detector(self, host_ip):
         return RebootDetector(self, host_ip)
 
+    def get_client_service_name(self):
+        output = self.run("if test -e /lib/systemd/system/mender.service; then echo mender; else echo mender-client; fi")
+        return output.stdout
```
lluiscampos, line 131: This should be wrong: `self.run` is already returning `.stdout` (this is a key difference between integration tests and the more modern suites like mender-binary-delta)

testutils/infra/device.py
```diff
@@ -272,6 +275,13 @@ def ssh_is_opened(self, wait=60 * 60):
         for dev in self._devices:
             dev.ssh_is_opened(wait)
 
+    def get_client_service_name(self):
+        # We assume that the service name is always the same across all devices,
+        # so it's enough to return the first one.
+        assert len(self._devices) > 0
+        for dev in self._devices:
+            return dev.get_client_service_name()
```
lluiscampos, line 283: This is fine, but for simplicity I would have used `return self._devices[0].get_client_service_name()`

kacf:
> How is that we haven't seen this before? I am a bit confused. Any test on warrior should have been failing for a while now, right?

Any meta-mender/warrior build, yes. But we build meta-mender/master, usually. Although they both use poky/warrior as their base, they have quite different content, and master will soon become zeus.

testutils/infra/device.py
kacf: Yes, you're right, somehow I got the idea that this was a map. Will fix.

testutils/infra/device.py
kacf: Ah good catch. I didn't see that. Will fix!

kacf:
Updated!

PR #867, opened 2020-04-03
Smoke test: run all tests using only enterprise images

lluiscampos:
This is POC meant to showcase how the tests could be run in integration
staging branch.

IOW, we want to avoid creating meaningless staging branches in OS repos
which are not used in HM stating.


tranchitella:
@lluiscampos this was in my TODO list, too! well done! :)

lluiscampos:
All integration tests have actually passed :rocket:

Expect a follow-up PR with all the required changes

PR #869, opened 2020-04-03
Fix incorrect 'pass' parsing when more than one entry matches gitlab.com.

kacf:
Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -226,9 +226,6 @@ def get_value_from_password_storage(server, key):
     else:
         keys = key
 
-    # Regex to extract the alphanumeric key from the tree output
-    pass_tree_re = re.compile(r".* (?:\x1b\[[\d;]*m)?([\w\d\.]+).*")
-
     try:
         # Remove https prefix.
         if server.startswith("https://"):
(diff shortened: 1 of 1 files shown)


kacf:
Cancelling pipeline. This is not going to affect it.

kacf:
@lluiscampos: It was inevitable it would happen at some point! :-)

extra/release_tool.py
```diff
@@ -237,18 +234,34 @@ def get_value_from_password_storage(server, key):
...
-        if count == 0:
-            return
+        pass_dir = os.getenv("PASSWORD_STORE_DIR")
+        if not pass_dir:
+            pass_dir = os.path.join(os.getenv("HOME"), ".password-store")
+
+        server_path_str = os.getenv("PASS_GITLAB_COM")
+
+        if not server_path_str:
+            output = subprocess.check_output(["find", pass_dir, "-type", "f", "-path", "*%s*" % server]).decode()
```
lluiscampos, line 244: Aha! So using `find -path` is what I was missing when I did the first implementation. I couldn't figure out how to parse the tree like output of pass...

Thanks!

kacf: Hehe yes, this was after several iterations of frustratingly trying to parse `pass` output. It really has a very difficult to parse format, `find` is much easier.

PR #876, opened 2020-04-07
Misc test fixes and optimization.

lluiscampos:
See individual commits.

With regards to 3d4cc2a, we usually don't do this as we branch out `integration` repo to follow the release branches.

@kacf @oleorhagen You can argue that is not really necessary and we can live with these two tests failing. Generally speaking: is is a good idea to have known tests failing in staging or should be do these kind of fixes when we find problems?

### DIFF
--- a/testutils/infra/device.py
+++ b/testutils/infra/device.py
@@ -60,6 +60,7 @@ class MenderDevice:
             connect_kwargs={"password": "", "banner_timeout": 60, "auth_timeout": 60},
         )
         self._conn.client.set_missing_host_key_policy(IgnorePolicy())
+        self._service_name = None
 
     @property
     def host_string(self):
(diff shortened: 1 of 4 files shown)


lluiscampos:
> LGTM :)
> 
> I don't fully understand how in: [53951cf](https://github.com/mendersoftware/integration/commit/53951cfe334e7ee67a2ca716d908bcf9aa7860ce) you go from `device.get_client_service_name()` to a hardcoded value, whilst in [1a75d0c](https://github.com/mendersoftware/integration/commit/1a75d0cca35c1207665dbc005dd80efd209ae031) you go the other way though.

Oh, I see there is a typo in my commit message for the latter (I'll amend). But the point is that the "service" there is referring to the Docker container name, does nothing to do with systemd. Actually, that test uses a non quemu client...

testutils/infra/device.py
```diff
@@ -127,9 +128,11 @@ def get_reboot_detector(self, host_ip):
         return RebootDetector(self, host_ip)
 
     def get_client_service_name(self):
-        return self.run(
-            "if test -e /lib/systemd/system/mender.service; then echo mender; else echo mender-client; fi"
-        ).strip()
+        if self._service_name is None:
```
kacf, line 131: :+1:

lluiscampos:
Last push is only an amend in the commit message. Not running tests again.

PR #877, opened 2020-04-10
[release tool] Add staging and feature/* branches to -f option

lluiscampos:
Add branches to --integration-versions-including option so that we can
launch builds with this specials branches on PRs to each repo.

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -1887,6 +1887,8 @@ def do_integration_versions_including(args):
                  "--sort=-version:refname:short",
                  "refs/tags/*",
                  "refs/remotes/%s/master" % remote,
+                 "refs/remotes/%s/staging" % remote,
+                 "refs/remotes/%s/feature/*" % remote,
                  "refs/remotes/%s/[1-9]*" % remote]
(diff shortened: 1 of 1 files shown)


lluiscampos:
Just FYI @kacf

extra/release_tool.py
```diff
@@ -1887,6 +1887,8 @@ def do_integration_versions_including(args):
                  "--sort=-version:refname:short",
                  "refs/tags/*",
                  "refs/remotes/%s/master" % remote,
+                 "refs/remotes/%s/staging" % remote,
+                 "refs/remotes/%s/feature/*" % remote,
```
kacf, line 1891: I *think* this may cause our integration-test-runner to build all feature branches that contain master, which we probably don't want. We may need to find a better way to handle this.

lluiscampos: Okey. The `feature/*` part is not in use yet, so I will verify your thinking and just remove it until I find a better solution.

As for the staging part: here it won't be a problem as integration/staging will not be using any component in its master version.

Thanks!

PR #878, opened 2020-04-10
[release tool] Support branch integrations using tags for -g option

lluiscampos:
Some integration branch (i.e. staging) will be using tags for some of
its components (in staging branch, mender or mender-artifact).

Before this change --version-of would be prepending the remote if the
input had a remote; now it will avoid prepending the remote to a tag,
which was resulting in invalid git syntax for a revision.


lluiscampos:
Also FYI @kacf

PR #879, opened 2020-04-15
[release_tool] Fixes for versions with `/` (branches) and remove feature/*

lluiscampos:
- [release_tool] Fixes to support versions with `/` (branches)
Couple of fixes to make the tool compatible with versions (branches)
containing slash character.

- [release tool] Remove feature/* branches for -f option
This will not work as expected, as it will make integration-test-runner
to build all these feature branches every time there is a PR to any
component that is in master.
We will need to find an smarter and efficient way to do this.

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -1899,7 +1899,6 @@ def do_integration_versions_including(args):
                  "refs/tags/*",
                  "refs/remotes/%s/master" % remote,
                  "refs/remotes/%s/staging" % remote,
-                 "refs/remotes/%s/feature/*" % remote,
                  "refs/remotes/%s/[1-9]*" % remote]
     if args.all:
         git_query += ["refs/heads/**"]
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -346,11 +346,12 @@ def get_docker_compose_data_from_json_list(json_list):
             full_image = cont_info.get('image')
             if full_image is None or ("mendersoftware" not in full_image and "mender.io" not in full_image):
                 continue
-            split = full_image.rsplit("/", 1)
-            prefix = split[0]
-            split = split[1].split(":", 1)
-            image = split[0]
+            split = full_image.split(":", 1)
+            prefix_and_image = split[0]
             ver = split[1]
+            split = prefix_and_image.rsplit("/", 1)
+            prefix = split[0]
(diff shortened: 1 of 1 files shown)


lluiscampos:
This PR was motivated by this comment: https://github.com/mendersoftware/integration/pull/877#discussion_r407906769

extra/release_tool.py
```diff
@@ -418,12 +419,15 @@ def version_of(integration_dir, yml_component, in_integration_version=None):
             if len(rev_range) > 1:
                 range_type = ".."
         repo_range = []
+
+        remote = ""
+        match = re.match("^(.*)/", rev_range[0])
+        if match is not None:
+            remote_candidate = match.group(1)
+            if subprocess.call("git remote  | egrep -q ^%s$" % remote_candidate, shell=True) == 0:
+                remote = remote_candidate + "/"
```
kacf, line 428: I have a slightly cleaner approach:

```bash
remote_candidate = match.group(1)
if subprocess.call("git rev-parse -q --verify refs/heads/%s > /dev/null" % remote_candidate, shell=True) != 0:
    remote = remote_candidate + "/"
```

First off, it's a little bit more targeted, and second, note that I flipped the condition, which matches the Git standard behavior, which is to match local branches first.

lluiscampos: Thank you. `rev-parse` and `for-each-ref` are two commands that I really need to learn...

lluiscampos: But wait, this doesn't work. Locally running it:
```
$ git rev-parse --verify refs/heads/lluiscampos
fatal: Needed a single revision
```
So maybe what you intend is to pass the rest of the string, so that if the input is `lluiscampos/feature/fancy` we call `git rev-parse ... refs/heads/feature/fancy`?

lluiscampos: Okey, so looking at it with fresh eyes I understood what you meant. You can ignore my previous comment; you would expect that to fail, of course.

But your logic still wasn't 100% correct, as branches with `/ ` in their name will also fail this check and then append the first part it as a false remote.

Anyway, I fixed it in an slight different way. Pleas see my last force-push.

kacf: Of course, doh! Good that you fixed my broken example! :smile:

PR #882, opened 2020-04-17
[release_tool] Strict the release branches match for -f

lluiscampos:
Restrict the release branches match for -f option so that it doesn't match arbitrary branches starting with numbers, only release branches.


lluiscampos:
@kacf Good feedback, thanks.

PR #886, opened 2020-04-23
Expose tenantadm device limit parameters in enterprise template

alfrunes:


production/config/enterprise.yml.template
```diff
@@ -14,3 +14,12 @@ services:
...
             DEVICEAUTH_DEFAULT_TENANT_TOKEN: ''
+
+    mender-tenantadm:
+        environment:
+            # The TENANTADM_DEVICE_LIMIT_* variables sets the default upper 
+            # limit on number of devices the tenant is able to accept
+            # (0: no limit).
+            TENANTADM_DEVICE_LIMIT_STARTER_DEFAULT: 50
+            TENANTADM_DEVICE_LIMIT_PROFESSIONAL_DEFAULT: 250
+            TENANTADM_DEVICE_LIMIT_ENTERPRISE_DEFAULT: 0
```
kacf, line 25: Isn't unlimited already the default if Enterprise is selected? Just wondering if we need to expose this to users if it's only relevant for Hosted Mender.

alfrunes: It's actually 500 currently...

kacf: Ok. IMHO we should not add this to `enterprise.yml.template`, but instead change the default for the Enterprise plan. Not only is this something we should not burden the user with, but it's a file that we cannot easily change down the road, because this file is meant to be *copied* and then kept out-of-repo.

alfrunes: Ok, so if I understand you correctly; I should close this PR and proceed with the [other PRs](https://tracker.mender.io/browse/MEN-3352?focusedCommentId=106643#comment-106643)?

kacf: Yes, exactly! 😀

alfrunes:
Other PRs are merged, closing this one; not including the parameters to the template...

PR #887, opened 2020-04-23
fix backend integration tests after introducing new deployments flow

kjaskiewiczz:


backend-tests/tests/test_deployments.py
```diff
@@ -1400,7 +1400,18 @@ def do_test_deployment_status_update(self, clean_mongo, user_token, devs):
             substate="bar",
         )
 
-        # second device
+        # Try to retrieve next update and assert expected status code
+        resp = deploymentsd.with_auth(devs[3].token).call(
```
merlin-northern, line 1404: is 3 a magic number, or is it itentional?

kjaskiewiczz: there are 5 devices; 4 are part of the deployment
after the 4th device (dev[3]) will finish the deployment (3 devices finished it at this point) we can check the deployment status (should be finished)

PR #888, opened 2020-04-24
[extra tools] Black-format release_tool and dump-gitlab-env

lluiscampos:
And enforce formatting for all tools in extra/ dir except for gitdm.


lluiscampos:
I skimmed the changes and I didn't see anything suspicious. I will let all tests run just in case...

PR #889, opened 2020-04-27
Move dependency between deployments and storage-proxy to demo yaml

tranchitella:
The docker-compose service mender-deployments depends on storage-proxy,
which is defined in docker-compose.demo.yml. This causes docker-compose
commands, as well as the logging command executed when the user presses
ENTER after running ./demo up, to fail with the following message:

ERROR: Service 'mender-deployments' depends on service 'storage-proxy'
which is undefined.

This commit moves the dependency on storage-proxy to the file where it
really belong to: docker-compose.demo.yml

Changelog: None

### DIFF
--- a/docker-compose.demo.yml
+++ b/docker-compose.demo.yml
@@ -58,6 +58,8 @@ services:
             DEPLOYMENTS_AWS_AUTH_KEY: minio
             DEPLOYMENTS_AWS_AUTH_SECRET: minio123
             DEPLOYMENTS_AWS_URI: https://s3.docker.mender.io:9000
+        depends_on:
+            - storage-proxy
 
     minio:
         networks:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,6 @@ services:
             - mender
         depends_on:
             - mender-mongo
-            - storage-proxy
 
     #
     # mender-gui


tranchitella:
I'm restarting the test_full_integration job, it failed for unrelated reason but I prefer to be sure my change does not affect it.

kacf:
I suspect this is not working correctly. Take a look at this error from mender-docs, which runs the production environment: https://gitlab.com/Northern.tech/Mender/mender-docs/-/jobs/528170811. See in particular this error:

```
mender-deployments_1             | caused by: Put https://s3.docker.mender.io:9000/mender-artifact-storage: dial tcp: lookup s3.docker.mender.io on 127.0.0.11:53: no such host
```

Now it can't find the storage anymore. I can't reproduce it locally, but it's probably a race condition, as it is possible the storage has had time to start, but not always. We may need to put this back, and rather find out why the demo script is not using the `docker-compose.demo.yml` in all the places it should. The bare `docker-compose.yml` strictly speaking isn't, and never was, a complete service definition.

@tranchitella, @lluiscampos: Thoughts?

tranchitella:
@kacf what about this diff?

```diff --git a/production/config/prod.yml.template b/production/config/prod.yml.template
index 4273697..aee2119 100644
--- a/production/config/prod.yml.template
+++ b/production/config/prod.yml.template
@@ -111,6 +111,8 @@ services:
             # devices will access storage using https://s3.acme.org:9000, then
             # set this to https://s3.acme.org:9000
             DEPLOYMENTS_AWS_URI: https://set-my-alias-here.com
+        depends_on:
+            - storage-proxy
         logging:
             options:
                 max-file: "10"```

From what I see, in the prod environment demo is not included, thus there is no "depends_on" enforced between deployments and storage-proxy. As deployments is configured to use the storage proxy, adding the depends_on there seems logical to me.

kacf:
We should not add this to a template, because the user is expected to copy this elsewhere and keep it across upgrades. Essentially it would be impossible to remove it again.

Was it only the demo script, or was there another motive for the original change?

tranchitella:
@kacf only the demo, here you have the reference:
https://hub.mender.io/t/unable-to-start-demo/1740/14

PR #891, opened 2020-04-28
Update migration scripts (dump/migrate/restore)

tranchitella:
The dump-db, migrate-db and restore-db in migration/ needs to be updated to reflect the new back-end components. Additionally, use mongodb 3.6 utilities to dump/restore mongo database.

### DIFF
--- a/migration/dump-db
+++ b/migration/dump-db
@@ -8,6 +8,9 @@ SERVICES="\
   mender-useradm \
   mender-device-auth \
   mender-deployments \
+  mender-create-artifact-worker \
+  mender-workflows-server \
+  mender-workflows-worker \
 "
 DB_SERVICES="\
   mender-mongo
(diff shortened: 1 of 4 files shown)


tranchitella:
As a side note: I've actually tested the scripts locally, this is not a fire-and-forget PR. :)

tranchitella:
Stopped the pipeline to save $$$, we don't need it for this PR.

tranchitella:
@lluiscampos well, in 2.3 we don't have workflows-worker, shall I split it in a different commit? I think so...

lluiscampos:
> @lluiscampos well, in 2.3 we don't have workflows-worker, shall I split it in a different commit? I think so...

I see. Well, I think the easiest is merge this one as-is (we don't need changelog here) and just launch an independent fix for 2.3.x

PR #892, opened 2020-04-28
[production_test_env] Update shebang to use Python 3

lluiscampos:
Update shebang to use Python 3 and adjust the help text for what the script actually supports.


lluiscampos:
The failing test is unrelated.

PR #894, opened 2020-04-29
test dynamic deployments

mchalski:
https://tracker.mender.io/browse/MEN-3332

what's tested:
- predicates handling (single, combined, on single and many attrs)
- deployments lifecycle (status, stats changes, both for unbounded and bounded deployments)
- ordering of dynamic deployments
- basic dynamic phased rollouts 

---
after huge amounts of back and forth, this should finally pass together with KJs fixes in deployments-enterprise: https://github.com/mendersoftware/deployments-enterprise/pull/130

combined pipeline:
https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/141548963
(should get unstuck when we purchase the minutes...)

### DIFF
--- a/backend-tests/tests/test_deployments.py
+++ b/backend-tests/tests/test_deployments.py
@@ -14,7 +14,9 @@ import testutils.api.deviceauth as deviceauth_v1
 import testutils.api.deviceauth_v2 as deviceauth_v2
 import testutils.api.useradm as useradm
 import testutils.api.inventory as inventory
+import testutils.api.inventory_v2 as inventory_v2
 import testutils.api.deployments as deployments
+import testutils.api.deployments_v2 as deployments_v2
 
 from testutils.api.client import ApiClient
 from testutils.common import (
@@ -27,7 +29,9 @@ from testutils.common import (
     create_authset,
     change_authset_status,
     clean_mongo,
+    mongo_cleanup,
     mongo,
+    get_mender_artifact,
 )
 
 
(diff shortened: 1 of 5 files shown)


testutils/api/inventory_v2.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2019 Northern.tech AS
```
merlin-northern, line 1: will something not complain about that year?

backend-tests/tests/test_deployments.py
```diff
@@ -1446,3 +1450,589 @@ class TestDeploymentsStatusUpdateEnterprise(TestDeploymentsStatusUpdateBase):
...
+
+        res = api_dep_v2.with_auth(utoken).call(
+            "POST", deployments_v2.URL_DEPLOYMENTS, deployment_req
+        )
+
+        assert res.status_code == status_code
+        if status_code != 201:
+            return None
+
+        depid = res.headers["Location"].split("/")[5]
```
merlin-northern, line 1523: I would consider checking here fi indeed the Location matches what we expect, and get explicit error if not. As another test for the correct Location:.

backend-tests/tests/test_deployments.py
```diff
@@ -1446,3 +1450,589 @@ class TestDeploymentsStatusUpdateEnterprise(TestDeploymentsStatusUpdateBase):
...
+        assert_get_next(200, dev.token, "foo3")
+
+    def test_phased_rollout(self, mongo):
+        """ Check phased rollouts with and without max_devices.
+        """
+
+        cases = [
+            # without max_devices
+            {
+                "name": "with max_devices",
```
merlin-northern, line 1944: this one is `"without max_devices"`, right?

testutils/api/inventory_v2.py
lluiscampos: We don't run copyright checks in tests.

merlin-northern: but still that is not a true year of Copyright ;-)

mchalski: no thing complained... but fixed regardless, thanks

backend-tests/tests/test_deployments.py
mchalski: yup, thanks

backend-tests/tests/test_deployments.py
mchalski: well, if the location is not what we assume - it will blow up the test and pretty quickly tell us where it failed, so I don't think it's necessary

backend-tests/tests/test_deployments.py
```diff
@@ -1446,3 +1450,613 @@ class TestDeploymentsStatusUpdateEnterprise(TestDeploymentsStatusUpdateBase):
...
+    return d
+
+
+def submit_inventory(attrs, token):
+    invd = ApiClient(inventory.URL_DEV)
+    r = invd.with_auth(token).call("PATCH", inventory.URL_DEVICE_ATTRIBUTES, attrs)
+    assert r.status_code == 200
+
+
+def update_deployment_status(deployment_id, status, token):
```
kjaskiewiczz, line 1581: if I understand correctly this should be more like update_device_deployment_status(device_deployment_id,...
in the deployments service, we have deployments and device deployments and it would be good not to mix those things here

PR #895, opened 2020-04-29
MEN-4366: New versioning scheme for backend Docker tags

lluiscampos:
So here goes all my work around release_tool. Best reviewed commit by commit, I think.

From the interface point of view, it touches two parts of the tool.
- CLI options `--version-of` and `--set-version-of`: they use a new file `internal-vesions.yml` to figure out where to get/set the Git or Docker version.
- Release process: new flag in `components-maps,yml` informs the tool about which repos follow Mender product version and which Git tags when publishing Docker images.

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -1768,16 +1768,19 @@ def set_docker_compose_version_to(dir, repo, tag, git_tag=None):
 
         if git_tag is not None:
             # Replace Git tag with a new one.
-            with open(filename) as fd:
-                full_content = "".join(fd.readlines())
-            with open(filename, "w") as fd:
-                fd.write(
-                    re.sub(
-                        r"(\s*%s:[\n\s]+git-version:\s+)(.*)" % re.escape(repo.git()),
-                        r"\g<1>%s" % git_tag,
-                        full_content,
+            assosiated_repos = repo.associated_components_of_type("git")
+            for assosiated_repo in assosiated_repos:
+                with open(filename) as fd:
+                    full_content = "".join(fd.readlines())
+                with open(filename, "w") as fd:
+                    fd.write(
+                        re.sub(
+                            r"(\s*%s:[\n\s]+git-version:\s+)(.*)"
+                            % re.escape(assosiated_repo.git()),
+                            r"\g<1>%s" % git_tag,
+                            full_content,
+                        )
                     )
-                )
 
 
 def purge_build_tags(state, tag_avail):
@@ -2477,7 +2480,9 @@ def do_set_version_to(args):
         sys.exit(1)
 
     repo = Component.get_component_of_any_type(args.set_version_of)
-    set_docker_compose_version_to(integration_dir(), repo, args.version)
+    set_docker_compose_version_to(
+        integration_dir(), repo, args.version, git_tag=args.version
+    )
 
 
 def is_marked_as_releaseable_in_integration_version(
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -1,4 +1,17 @@
 #!/usr/bin/env python3
+# Copyright 2020 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        https://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
 
 import argparse
 import copy
@@ -362,8 +375,10 @@ def docker_compose_files_list(dir):
     """Return all docker-compose*.yml files in given directory."""
     list = []
     for entry in os.listdir(dir):
-        if entry == "other-components.yml" or (
-            entry.startswith("docker-compose") and entry.endswith(".yml")
+        if (
+            entry == "git-versions.yml"
+            or entry == "other-components.yml"
+            or (entry.startswith("docker-compose") and entry.endswith(".yml"))
         ):
             list.append(os.path.join(dir, entry))
     return list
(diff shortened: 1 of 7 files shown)


extra/release_tool.py
```diff
@@ -1870,12 +1914,20 @@ def push_latest_docker_tags(state, tag_avail):
...
+                    ]
+                else:
+                    minor_version = overall_minor_version
 
             prefix = compose_data[image.docker_image()]["image_prefix"]
 
+            if image.is_independent_component():
+                build_tag = tag_avail[repo.git()]["build_tag"]
+            else:
+                build_tag = state["image_version"]
```
kacf, line 1944: Should use `tag_avail`.

extra/release_tool.py
```diff
@@ -1999,6 +2056,12 @@ def do_docker_compose_branches_from_follows(state):
                 bare_branch = branch
 
             set_docker_compose_version_to(checkout, repo, bare_branch)
```
kacf, line 2063: This is repeated below. Remove?

extra/release_tool.py
```diff
@@ -1989,6 +2041,11 @@ def do_docker_compose_branches_from_follows(state):
         state, "integration", state["integration"]["following"]
     )
 
+    # For the Docker images, use M.N as the release branch
+    mender_release_minor = re.match(
+        r"^([0-9]+\.[0-9]+)\.[0-9]+.*$", state["version"]
+    ).group(1)
```
kacf, line 2052: Won't this be missing the `.x`? The `M.N` Docker tags are stable releases, the `M.N.x` tags are unstable rolling branches. Probably we won't use the former anymore, but I think we should still stick to the convention.

extra/release_tool.py
```diff
@@ -1912,14 +1913,26 @@ def push_latest_docker_tags(state, tag_avail):
...
+            elif tip.isnumeric():
+                if image.is_independent_component():
+                    new_version = state[repo.git()]["version"][
+                        0 : state[repo.git()]["version"].index(".")
+                    ]
+                else:
+                    new_version = overall_major_version
+            else:
+                raise Exception(
+                    "Unrecognized tip %s, expected 'latest', minor-like ormajor-like"
```
kacf, line 1935: `s/ormajor/or major/`

extra/release_tool.py
```diff
@@ -1912,14 +1913,26 @@ def push_latest_docker_tags(state, tag_avail):
...
+            elif tip.count(".") == 1:
                 if image.is_independent_component():
-                    minor_version = state[repo.git()]["version"][
+                    new_version = state[repo.git()]["version"][
                         0 : state[repo.git()]["version"].rindex(".")
                     ]
                 else:
-                    minor_version = overall_minor_version
+                    new_version = overall_minor_version
+            elif tip.isnumeric():
```
kacf, line 1926: I found this particular call confusing, especially in contrast to the previous `elif`. Is it better to use `tip.count(".") == 0`?

extra/release_tool.py
```diff
@@ -1758,17 +1758,19 @@ def set_docker_compose_version_to(dir, repo, tag, internal=None):
```
kacf, line 1758: Commit message: `s/set-verion-of/set-version-of/`

extra/release_tool.py
```diff
@@ -71,7 +71,7 @@ class NotAVersionException(Exception):
 
 
 class Component:
-    COMPONENT_MAPS = None
+    COMPONENT_MAPS = {}
```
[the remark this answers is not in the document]
lluiscampos: For the error checking, I don't think it is a big difference, as an [empty dict evaluates to False in Python](https://docs.python.org/3/library/stdtypes.html#truth-value-testing). I did the change to avoid the [local] pylint warning that COMPONENT_MAPS was a "unsubscriptable" (and we first define it as None, then as a dict).

I can revert this if you wish, no strong opinion.

extra/release_tool.py
```diff
@@ -384,29 +400,47 @@ def get_docker_compose_data_from_json_list(json_list):
...
+                    data[container] = dict()
+                if data[container].get("internal_version") is not None:
+                    raise Exception(
+                        (
+                            "More than one service specifying internal-version for '%s'. "
+                            + "The tool currently does not support this."
+                        )
+                        % container
+                    )
+                data[container].update(
```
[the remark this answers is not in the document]
lluiscampos: So here `container` is a bad variable name, really. What you are getting are the "services" of a given yaml file, which in the case of "internal-versions.yml" (future git-versions) is really the repo names, not the image nor the container.

Does it makes sense?

I will change the for loop to be something like for `services in ...`

extra/release_tool.py
```diff
@@ -1164,6 +1179,12 @@ def generate_new_tags(state, tag_avail, final):
                 )
             print()
 
+    if final:
+        image_version = "mender-" + state["version"]
+    else:
+        image_version = "mender-%s-build%d" % (state["version"], highest + 1)
+    update_state(state, ["image_version"], image_version)
```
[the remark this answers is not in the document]
lluiscampos: This makes a lot of sense to me. In my first approach I went that way, but I was setting them at single repo level, which lead me to lots of trouble. But I think I know how to do it now, I will rework this part; thanks!

extra/release_tool.py
```diff
@@ -1705,13 +1734,13 @@ def tag_or_followed_branch(repo_git):
...
 
 
-def set_docker_compose_version_to(dir, repo, tag):
+def set_docker_compose_version_to(dir, repo, tag, internal=None):
     """Modifies docker-compose files in the given directory so that repo_docker
     image points to the given tag."""
 
-    for yml in repo.yml_components():
-        compose_files = docker_compose_files_list(dir)
-        for filename in compose_files:
```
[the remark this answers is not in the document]
lluiscampos: It is not for fun: I did it to be able to iterate over the files but not over the `yml_components` in the below block `if internal is not None: ...`

extra/release_tool.py
lluiscampos: Yes, I also think it is confusing. What I don't like of `tip.count(".") == 0` (my first approach, actually) is that it would go for any other generic string. But yeah, as we are defining the list in this very function, it is kind of safe so to avoid confusion I'll change it.

extra/release_tool.py
lluiscampos: I thought about this too. The thing is that we don't have a way now to publish the M.N.x tags of the different services (following Mender product versions).

Thinking about it now, maybe this is something that we should do? It should be fairy simple from the template to query release_tool for the integration version and publish the mender-N.M.x images. Unless you disagree I'll do this also (this was a TBD point in the JIRA ticket, actually)

lluiscampos: Note to self: the version here is missing the `mender-` prefix.

lluiscampos:
@kacf Thank you for a detailed and very valuable review. Everything not replied means "Thanks, I will fix".

extra/release_tool.py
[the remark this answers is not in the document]
lluiscampos: This I made in a different commit for you to give it a close look. I will squash it before final merge.

extra/release_tool.py
lluiscampos: Fixed here; but I have now a TODO PR in `mendertesting` for these Docker tags to be actually updated.

lluiscampos:
@kacf I pushed a new rework addressing most of your comments (I hope!) This time I haven't run through a full release simulation, but I am confident enough with the code now.

I have to fix one more thing with the unit tests (they will not pass in any branch but master, which will be an ugly problem in release branches), but the tool code can be reviewed as-is. Thanks!

extra/release_tool.py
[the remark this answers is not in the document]
kacf: My idea is that if you define it to `None`, you get exceptions if it's uninitialized, whereas if you set it to `[]`, you get silent misbehavior.

extra/release_tool.py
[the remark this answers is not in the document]
kacf: Oh, actually I thought this optimized the section by only opening the file once, but it still opens it the same number of times as before. Just disregard my comment, I misunderstood.

extra/release_tool.py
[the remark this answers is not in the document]
lluiscampos: Good point, I like that. I'll revert this and find a way to silent my pylint warning :)

PR #899, opened 2020-05-06
Update versions for all backend images to mender-master

lluiscampos:
MEN-3466: Update backend images to use version mender-master

Introducing a new versioning schema, from this release on the Docker
images for the backend repositories will be published in their
corresponding registries following the Mender product version.

This means tags `<service>:mender-<mender-version>` instead of the old
tags `<service>:<service-version>`, which will eventually be deprecated.

Changelog: Commit

### DIFF
--- a/docker-compose.client.rofs.yml
+++ b/docker-compose.client.rofs.yml
@@ -5,4 +5,4 @@ services:
     # mender-client
     #
     mender-client:
-        image: mendersoftware/mender-client-qemu-rofs:master
+        image: mendersoftware/mender-client-qemu-rofs:mender-master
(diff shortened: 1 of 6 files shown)


lluiscampos:
Depends on https://github.com/mendersoftware/mendertesting/pull/66 and https://github.com/mendersoftware/workflows/pull/47 (and a similar to the latter for `workflows-enterprise`).

It should be safe, but I will let the integration tests run just in case.

lluiscampos:
@kacf @tranchitella I decided to add a Changelog for this; namely for users doing custom updates of their on-prem installations. I would appreciate a close look at it before I merge. Thanks!

Stopping the last build. The previous one (no code changes) was successful @ [Pipeline 143213751](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/143213751).

PR #900, opened 2020-05-06
[release_tool] Make -f option run regardless of the directory name

lluiscampos:
The previous implementation would not work on an integration repo
checked out in a directory named something else.

The rest of the uses of find_upstream_remote are linked to the release
process or to --select-test-suite, and in both cases all repositories are
required (hence is okey to require the same directory name).

Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -871,16 +871,22 @@ def cleanup_temp_git_checkout(tmpdir):
     shutil.rmtree(tmpdir, ignore_errors=True)
 
 
-def find_upstream_remote(state, repo_git):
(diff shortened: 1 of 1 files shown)


lluiscampos:
@kacf Turned out that https://github.com/mendersoftware/mendertesting/pull/66 showed a bug (rather a weakness) in release_tool. See the error [here](https://gitlab.com/Northern.tech/Mender/deployments/-/jobs/540578093).

lluiscampos:
The other obvious fix would be to change the template to check out the repo in `/tmp/integration` or similar. But I thought it was better to not require an specific dir name for these independent CLI functionalities that the tool provides.

PR #902, opened 2020-05-07
[release_tool] Fixes for --integration-versions-including

lluiscampos:
* Compose files fix: Only backend shall use version mender-master
Fixes commit 56140bf; where following its own commit message only the
backend components should have been updated to "mender-master" version.

* [release_tool] Fix -f option to prefer git versions over docker
The use case is for the user (or bot) to pass a git version in -v
parameter. So compare against internal git version when present.
Added a unit tests for it.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -32,6 +32,8 @@ test:extra-tools:release-tool:
 
   before_script:
     - pip install pytest pyyaml
+    - git remote add github https://github.com/mendersoftware/integration.git
+    - git fetch github
 
   script:
     # Run release-tool unit tests.
--- a/docker-compose.client.rofs.yml
+++ b/docker-compose.client.rofs.yml
@@ -5,4 +5,4 @@ services:
     # mender-client
     #
     mender-client:
-        image: mendersoftware/mender-client-qemu-rofs:mender-master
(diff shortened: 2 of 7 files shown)


lluiscampos:
@kacf The unit tests fail because they cannot find a remote pointing to github.com, as the remote points to gitlab.com. Should I modify the regex to match any git server?

kacf:
> The unit tests fail because they cannot find a remote pointing to github.com, as the remote points to gitlab.com. Should I modify the regex to match any git server?

If by "any" you mean either `github.com/mendersoftware` or `gitlab.com/Northern.tech/Mender`, then sure. I don't think it should *any* server, because it should filter out personal forks of which I have tons of locally.

lluiscampos:
> > The unit tests fail because they cannot find a remote pointing to github.com, as the remote points to gitlab.com. Should I modify the regex to match any git server?
> 
> If by "any" you mean either `github.com/mendersoftware` or `gitlab.com/Northern.tech/Mender`, then sure. I don't think it should _any_ server, because it should filter out personal forks of which I have tons of locally.

Correct. I thought about that afterwards. And even only with those (which I have in many repos) will be problematic as the tool still has to chose one.

So I modified the GitLab CI script instead, to add a GitHub based remote before running the tests. That should do the trick :)

PR #906, opened 2020-05-07
[release-tool/compose files]: Fix for mender-api-gateway-docker

lluiscampos:
The actual fix in this commit is renaming "mender-api-gateway-docker"
(i.e. git repo name) to "api-gateway" (i.e. docker image name) in
git-versions.yml file.

Added some code comments around to clarify the design, as well as an
strict control for CLI option --version-of to disallow ambiguous uses.
Few components have different name for the repo than for the Docker
image (workflows-worker, api-gateway) which will make this option return
unexpected versions. Throw an informative error for these cases.

Unit tests added for it.

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -401,7 +401,7 @@ def get_docker_compose_data_from_json_list(json_list):
             "container": container_name,
             "image_prefix": "mendersoftware/" or "someserver.mender.io/blahblah",
             "version": version,
-            "git_version": git_version,
+            "git_version": git version of the assoiated repo,
         }
     }
     """
@@ -448,6 +448,9 @@ def get_docker_compose_data_from_json_list(json_list):
                         )
                         % service
                     )
+                # NOTE: For yaml files containing "git-version" fields (namely git-versions.yml),
+                # these hang on "services" keys representing Docker image names (instead of Docker
+                # containers names). Hence we merge the dicts here assuming "service" == image.
                 data[service].update(
                     {"git_version": git_version,}
                 )
@@ -584,11 +587,18 @@ def do_version_of(args):
         print("Unrecognized repository: %s" % args.version_of)
         sys.exit(1)
 
-    yml_component = comp.yml_components()[0]
-
     assert args.version_type in ["docker", "git"], (
         "%s is not a valid name type!" % args.version_type
     )
+
+    if comp.type in ["docker_image", "docker_container"] and args.version_type == "git":
+        print(
+            "Unsuported %s %s for --version-type git. Use --version-type docker instead"
+            % (comp.type, args.version_of)
+        )
+        sys.exit(1)
+
+    yml_component = comp.yml_components()[0]
 
     print(
         version_of(
--- a/git-versions.yml
+++ b/git-versions.yml
@@ -1,17 +1,17 @@
 # This file lists all software components that are part of a Mender
-# release with their internal versions. The keys represent images
-# rather than "services", but we keep it for the sake of parsing it
-# the same way as other-components and docker-compose files.
+# release with their git versions. The keys represent images rather
+# than "services" (or containers), but we keep it for parsing it the
+# same way as other-components and docker-compose files.
 services:
 
     #
-    # backend open source services
+    # backend open source images
     #
     deployments:
         git-version: master
     gui:
         git-version: master
-    mender-api-gateway-docker:
+    api-gateway:
         git-version: master
     deviceauth:
         git-version: master
@@ -25,7 +25,7 @@ services:
         git-version: master
 
     #
-    # backend enterprise services
+    # backend enterprise images
     #
     deployments-enterprise:
         git-version: master
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,9 +36,7 @@ test:extra-tools:release-tool:
     - git remote add github https://github.com/mendersoftware/integration.git
     - git fetch github
     # Fetch master branch for tests using --in-integration-version
-    - if [ $CI_COMMIT_REF_NAME != "master" ]; then
-    -  git fetch origin master:master
-    - fi
+    - git fetch origin master:master
 
   script:
     # Run release-tool unit tests.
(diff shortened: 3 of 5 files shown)


lluiscampos:
I will let the full pipeline run this time to make sure `mender-qa` is happy with this new CLI level check.

extra/release_tool.py
```diff
@@ -401,7 +401,7 @@ def get_docker_compose_data_from_json_list(json_list):
             "container": container_name,
             "image_prefix": "mendersoftware/" or "someserver.mender.io/blahblah",
             "version": version,
-            "git_version": git_version,
+            "git_version": git version of the assoiated repo,
```
kacf, line 404: `s/assoiated/associated/`

extra/release_tool.py
```diff
@@ -584,12 +587,19 @@ def do_version_of(args):
...
 
-    yml_component = comp.yml_components()[0]
-
     assert args.version_type in ["docker", "git"], (
         "%s is not a valid name type!" % args.version_type
     )
 
+    if comp.type in ["docker_image", "docker_container"] and args.version_type == "git":
+        print(
+            "Unsuported %s %s for --version-type git. Use --version-type docker instead"
```
kacf, line 596: `s/Unsuported/Unsupported/`

git-versions.yml
```diff
@@ -1,17 +1,17 @@
...
     #
-    # backend open source services
+    # backend open source images
     #
     deployments:
         git-version: master
     gui:
         git-version: master
-    mender-api-gateway-docker:
+    api-gateway:
```
kacf, line 14: I think this should be the container name (`mender-api-gateway`), not the image name, so that the format of this file matches what is in docker-compose.yml. The previous `--version-of` command worked for all three names, so I think it should keep working like that.

In relation to that I just thought of a completely different fix which is way simpler than the current approach: Instead of loading both YML files and keeping separate entries, just load *either* `docker-compose.yml` or `git-versions.yml` depending on which one is desired, and return direct results from there. No need to have separate `git-version` entries, just make sure the files are the same format. With this we could remove quite a lot of special case logic. What do you think?

extra/release_tool.py
```diff
@@ -584,12 +587,19 @@ def do_version_of(args):
...
+            % (comp.type, args.version_of)
+        )
+        sys.exit(1)
```
kacf, line 599: I disagree with this check. All containers and images have a Git repository, so there should never be a reason not to return a result for `version_type == "git"`. See previous comment also.

git-versions.yml
lluiscampos: Thanks. I think is good that we are iterating the design so that we might end up with a simpler one :)

> I think this should be the container name (`mender-api-gateway`), not the image name, so that the format of this file matches what is in docker-compose.yml.

In the first iteration I was using repos names here, but it was very hard to merge the info while parsing the rest of the yaml files (see below wrt re: parsing them separately).

The problem with your suggestion is that the container names are not unique. In the old code, for each container we extract the prefix/image/version parameters and then use the "image" as the unique key for our data. We could take some shortcuts, for example if we force enterprise and os to have always the same versions, then we could unify the container names.

> The previous `--version-of` command worked for all three names, so I think it should keep working like that.

Yes, I would like to keep it too. It is clear that I added the restriction to compensate the shortcoming of this design. If we manage to change it, we can take out the restriction.

> 
> In relation to that I just thought of a completely different fix which is way simpler than the current approach: Instead of loading both YML files and keeping separate entries, just load _either_ `docker-compose.yml` or `git-versions.yml` depending on which one is desired, and return direct results from there. No need to have separate `git-version` entries, just make sure the files are the same format. With this we could remove quite a lot of special case logic. What do you think?

And this is something that in principle I like too, and maybe you have a good suggestion for the following problem.

The issue here is that not all components exist in the git-versions.yml file. The reasoning by me was to not duplicate version information: we would be using git-versions.yml only for the images that use different tags for docker/git (namely backend images) while still taking the version from the image specification for the rest (mender client, mender-cli, mender-artifact).

extra/release_tool.py
lluiscampos: Related to the above. Let's keep the discussion there: if we find a good solution I will happily remove the restriction.

extra/release_tool.py
lluiscampos: I am done with being embarrassed about these: I just installed an spell checker plugin to my editor :smiley:

git-versions.yml
kacf: > The problem with your suggestion is that the container names are not unique. In the old code, for each container we extract the prefix/image/version parameters and then use the "image" as the unique key for our data.

Can't we still do that? Use the exact same format, as in:

```
docker-compose.yml:
    mender-api-gateway:
        image: mendersoftware/api-gateway:mender-master

git-versions.yml:
    mender-api-gateway:
        image: mendersoftware/api-gateway:master
```

Then `image` can be used as the unique identifier as before. There will be one piece of duplicated information, and that is the container name, but I think that price is worth paying for keeping the parsing logic the same everywhere, as it's already complicated enough. Besides, the container name already exists as an "extra" component in the `other-components.yml` file, so it's not really new.

> The issue here is that not all components exist in the git-versions.yml file. The reasoning by me was to not duplicate version information: we would be using git-versions.yml only for the images that use different tags for docker/git (namely backend images) while still taking the version from the image specification for the rest (mender client, mender-cli, mender-artifact).

I agree that we should not duplicate version information. But I don't think we need to, we can just maintain a map of docker-only and git-only files:

```
DOCKER_ONLY_YML = ["docker-compose.yml", "docker-compose.enterprise.yml"]
GIT_ONLY_YML = ["git-versions.yml"]
```

and the rest of the YML files we always parse. This would ensure that the non-docker tools are always loaded, because they are in other files (`other-components.yml` and `docker-compose.client.yml`), and the docker-versions and git-versions are loaded only when they are needed.

Does this solve the problem, you think?

lluiscampos: Yes, that will overall work. The only side effect is that we need a separate `git-versions-enterprise.yml` to not override containers keys, but we can live with that.

lluiscampos:
See new PR https://github.com/mendersoftware/integration/pull/907

PR #908, opened 2020-05-11
MEN-3458: Backend-integration; limit acces by device group

alfrunes:


alfrunes:
NOTE: I left a comment in one of the test-cases which fails but I think it shouldn't.

testutils/util/artifact.py
```diff
@@ -0,0 +1,206 @@
...
+            getattr(fd, "name", "rootfs-%04d.ext4" % random.randint(0, 10000)),
+        )
+
+        self._filenames.append(filename)
+        self._payloads[filename] = fd
+        self._payload_types[filename] = payload_type
+        if isinstance(depends, dict):
+            self._depends[filename] = depends
+        if isinstance(provides, dict):
+            self._provides[filename] = provides
```
lluiscampos, line 64: It they are not `None`, I would assert that they are dict instances, to make sure we generate an error if a user passes a list or something else.

testutils/util/artifact.py
```diff
@@ -0,0 +1,206 @@
...
+
+        for filename in sorted(self._payloads.keys()):
+            path_prefix = os.path.join(
+                "headers", os.path.basename(os.path.dirname(filename))
+            )
+            typeinfo = {"type": self._payload_types[filename]}
+            if filename in self._depends:
+                typeinfo["artifact_depends"] = self._depends[filename]
+            if filename in self._provides:
+                typeinfo["artifact_provides"] = self._provides[filename]
```
lluiscampos, line 171: This could override an existing depends/provides, but I guess that is the correct behavior. Does mender-artifact give an error in these cases? Could we add some error here?

testutils/util/artifact.py
```diff
@@ -0,0 +1,206 @@
...
+import hashlib
+import json
+
+
+class Artifact:
+    """
+    Artifact provides a very simplistic implementation of mender artifact
+    that allows creating simple buffered artifact file objects that
+    resides in memory.
+    """
```
lluiscampos, line 14: Nice work!

testutils/util/artifact.py
alfrunes: That's true. I think we should update the artifact v3 specification as well as it doesn't say anything that the `type-info` provides/depends has to be globally unique.

testutils/util/artifact.py
alfrunes: :+1:

backend-tests/tests/test_rbac.py
```diff
@@ -0,0 +1,257 @@
...
+                "user": {"name": "test2@username.org", "pwd": "password"},
+                "permissions": [
+                    UserPermission("CREATE_DEPLOYMENT", "DEVICE_GROUP", "test")
+                ],
+                "device_groups": {"test": 5, "production": 25},
+                "deploy_groups": ["production"],
+                "status_code": 405,
+            },
+            # { TODO: This test-case should pass, but deployments only
+            #         expect device IDs to come from the same group
```
merlin-northern, line 178: yes, thats true, until we get MEN-3411 in. and when deployments restrictions are on with RBAC there is no other way: but to restrict the deployment constructor to only one group.

alfrunes: I'll just leave this test case commented out until this is implemented.

alfrunes:
Had to rebase on top of master as the QA pipeline was out of sync.

backend-tests/tests/test_rbac.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, action, permission_type, permission_target):
```
merlin-northern, line 30: not an easy commit/diff to read, but seems fine.

PR #910, opened 2020-05-16
Fix TestDevicePatchAttributes.test_ok, new system scope

tranchitella:
Changelog: none


merlin-northern:
@tranchitella should it somehow agree with https://github.com/mendersoftware/integration/pull/904?

tranchitella:
Duplicated of https://github.com/mendersoftware/integration/pull/904

PR #911, opened 2020-05-18
release_tool: Fix remote-name being dropped from version ranges.

kacf:
This happens when using -g and -i together with a range where one end
uses a remote reference and the other one doesn't. We are checking
whether the supplied reference is a remote or not, but only for the
first element of a range. We need to check it for all elements in the
range.

Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -532,24 +532,25 @@ def version_of(
             if len(rev_range) > 1:
                 range_type = ".."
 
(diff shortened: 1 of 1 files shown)


kacf:
@lluiscampos: I thought about adding a test for this, but since it is dependent on a remote, it's a bit tricky. Not impossible though, but I couldn't spot any other remote tests so far, so I thought maybe it's a bit too much effort for this internal tool. What do you think?

PR #912, opened 2020-05-18
statistics-generator: Make sure OS code is not counted twice for Ent repos.

kacf:
Many commits will be in both repos, but they should only be counted in
the Open Source repos.

Changelog: None

### DIFF
--- a/extra/statistics-generator
+++ b/extra/statistics-generator
@@ -61,27 +61,48 @@ collect_changes() {
     if [ "$REPO_ONLY" = 1 ]; then
         REPO_LIST=.
     else
-        REPO_LIST="$("$RELEASE_TOOL" --list git)"
+        # Sorting is important because we need "*-enterprise" repositories to
+        # come after their Open Source counterparts (see below).
+        REPO_LIST="$("$RELEASE_TOOL" --list git | sort)"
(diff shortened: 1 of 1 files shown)


lluiscampos:
> Many commits will be in both repos, but they should only be counted in
> the Open Source repos.

Why, really? Do we assume that an enterprise repo has always all the os functionality? I guess this will not always be true, when there the same API endpoint can do different things.

kacf:
> > Many commits will be in both repos, but they should only be counted in
> > the Open Source repos.
> 
> Why, really? Do we assume that an enterprise repo has always all the os functionality? I guess this will not always be true, when there the same API endpoint can do different things.

I realize now that this was very badly phrased. What I meant was:

"Many commits will be in both repos, but they should only be counted once. However, commits in Open Source will appear twice, since all commits in Open Source are also in Enterprise. Therefore subtract those from the Enterprise set, so that we have a single set of Open Source commits, from the Open Source repository, and a single set of Enterprise-only commits, from the Enterprise repository."

Is that clearer? AFAIK, Enterprise is a more or less strict superset of Open Source.

kacf:
Good, updated commit message!

PR #916, opened 2020-05-20
demo: Do not download the demo artifact if it is already downloaded.

drewmoseley:
Currently curl will fail if the file exists with the error message:
     Failed to download the demo Artifact

Changelog: None

### DIFF
--- a/demo
+++ b/demo
@@ -57,7 +57,7 @@ done
 # Check if the demo-Artifact has been downloaded,
 # or if there exists a newer one in storage.
 DEMO_ARTIFACT_NAME="mender-demo-artifact.mender"
-curl -sz mender-demo-artifact.mender -o mender-demo-artifact.mender https://dgsbl4vditpls.cloudfront.net/${DEMO_ARTIFACT_NAME}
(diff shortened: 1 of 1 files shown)


kacf:
Ah, hold on, you say that it failed. This should not happen. What is the message you get from curl?

drewmoseley:
I invoke
```
$ ./demo up
```
and the demo environment is launched properly.  I then hit ctrl-c and try again:
```
$ ./demo up
Failed to download the demo Artifact
```

lluiscampos:
> I invoke
> 
> ```
> $ ./demo up
> ```
> 
> and the demo environment is launched properly. I then hit ctrl-c and try again:
> 
> ```
> $ ./demo up
> Failed to download the demo Artifact
> ```

@drewmoseley I tried your instructions in `master` and in `2.4.0b1-build1` tag, and it works for me in both cases. The second `./demo up` fails on creating the user (as expected) but no error regarding the Artifact download...

kacf:
Yes, it works for me too. Perhaps you can add `-v` to curl to see what is happening?

drewmoseley:
Looks like it was something in my ~/.curlrc.  I updated the script to simply bypass the ~/.curlrc file.

PR #919, opened 2020-06-03
docker compose: minio now restart on-failure

clementperon:
Mender storage minio doesn't restart in case of failure.

Fix this by adding the restart rule in the docker compose file.

Changelog: Title

### DIFF
--- a/docker-compose.storage.minio.yml
+++ b/docker-compose.storage.minio.yml
@@ -6,6 +6,7 @@ services:
     #
     minio:
         image: minio/minio:RELEASE.2018-09-25T21-34-43Z
+        restart: on-failure
         networks:
             mender:
                 aliases:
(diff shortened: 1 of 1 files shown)


kacf:
@lluiscampos, @tranchitella: Do you have any idea why the backend tests are failing? I've run it several times, all the test cases are passing, and I cannot see a failure in the log. But still it seems to report failure at the end.

tranchitella:
@kacf I've restarted the backend tests job, it should be fine now because it looks like a spurious failure unrelated to this PR.

tranchitella:
@kacf it failed again, but the issue could be solved rebasing this PR against the latest master. I believe we can just merge the PR as it is, I'm 100% sure the pipeline is going to be green.

kacf:
Ok, thanks!

PR #921, opened 2020-06-05
MEN-3637 identity attributes count test update.

merlin-northern:
ChangeLog:none

### DIFF
--- a/backend-tests/tests/test_inventory.py
+++ b/backend-tests/tests/test_inventory.py
@@ -277,19 +277,19 @@ class TestDevicePatchAttributes:
 
             api_dev = r.json()
             # Expected inventory count per scope:
-            # {"inventory": 3, "identity": 1, "system": 2}
-            assert len(api_dev["attributes"]) == 6
+            # {"inventory": 3, "identity": 1+2, "system": 2} # +2 comes from the id_data see MEN-3637
+            assert len(api_dev["attributes"]) == 8
             # new scopes: identity and system holding authset status and
             #             time-stamp values respectively
 
             for a in api_dev["attributes"]:
-                if a["name"] == "mac":
+                if a["name"] == "mac" and a["scope"] == "inventory":
                     assert a["value"] == "mac-new-" + str(api_dev["id"])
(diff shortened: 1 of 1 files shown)


backend-tests/tests/test_inventory.py
```diff
@@ -277,13 +277,13 @@ def test_ok(self, user):
 
             api_dev = r.json()
             # Expected inventory count per scope:
-            # {"inventory": 3, "identity": 1, "system": 2}
-            assert len(api_dev["attributes"]) == 6
+            # {"inventory": 3, "identity": 1+2, "system": 2} # +2 comes from the id_data see MEN-3637
+            assert len(api_dev["attributes"]) == 8
```
lluiscampos, line 281: I don't understand why the test passed in https://github.com/mendersoftware/deviceauth/pull/351 and then failed in master.

backend-tests/tests/test_inventory.py
```diff
@@ -277,13 +277,13 @@ def test_ok(self, user):
...
             # new scopes: identity and system holding authset status and
             #             time-stamp values respectively
 
             for a in api_dev["attributes"]:
-                if a["name"] == "mac":
+                if a["name"] == "mac" && a["scope"] == "inventory":
```
lluiscampos, line 286: Mmm, then we are not checking any more `mac` for other scopes? Also, this change is silent (i.e. no mention in the commit) so I don't really understand the context.

backend-tests/tests/test_inventory.py
merlin-northern: I think it is because 351 ran without inventory changes merged in.

lluiscampos: That could be, but then the inventory changes PR should have failed...

Anyway, for the next time we should use a feature branch for these kind of multi repos changes.

backend-tests/tests/test_inventory.py
lluiscampos: @merlin-northern Thank you for a beautiful commit message :heart:

PR #922, opened 2020-06-05
release_tool: Fix exception when pushing Docker final release tags.

kacf:
When looking for tags, we need to keep track of whether any repository
has a build tag. If none have it, then it is a final release, and
global image_tag must be set specifically for that.

Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -950,6 +950,7 @@ def check_tag_availability(state):
 
     tag_avail = {}
     highest_overall = -1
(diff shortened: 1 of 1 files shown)


kacf:
> Obviously, I didn't try pushing final tags while developing this :)

Of course! All in all, the new tags worked remarkably well. IIRC, this is only issue I faced.

kacf:
Cherry-picked to 2.4.x as well.

PR #923, opened 2020-06-10
Restore docker-compose.storage.s3.yml

lluiscampos:
The file was removed by mistake at bc39515 as it was mistakenly
concluded that was used only for tests. However it is used by on-prem
setups for some community users.

### DIFF
--- a/docker-compose.storage.s3.yml
+++ b/docker-compose.storage.s3.yml
@@ -1,0 +1,23 @@
+version: '2'
+services:
+
+    #
+    # mender-deployments
+    #
+
+    #
+    # storage backend proxy used in conjunction with minio, applies
+    # rate & connection limiting
+    #
+
+    mender-deployments:
+        # S3 access configuration - override with your own values
+        # Keys have to grant access to default bucket: mender-artifact-storage
+        environment:
+            DEPLOYMENTS_AWS_TAG_ARTIFACT: "true"
+            DEPLOYMENTS_AWS_AUTH_KEY: ${AWS_ACCESS_KEY_ID}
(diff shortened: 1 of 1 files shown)


docker-compose.storage.s3.yml
```diff
@@ -0,0 +1,23 @@
...
+    mender-deployments:
+        # S3 access configuration - override with your own values
+        # Keys have to grant access to default bucket: mender-artifact-storage
+        environment:
+            DEPLOYMENTS_AWS_TAG_ARTIFACT: "true"
+            DEPLOYMENTS_AWS_AUTH_KEY: ${AWS_ACCESS_KEY_ID}
+            DEPLOYMENTS_AWS_AUTH_SECRET: ${AWS_SECRET_ACCESS_KEY}
+            DEPLOYMENTS_AWS_REGION: us-west-1
+            DEPLOYMENTS_AWS_URI: https://s3-us-west-1.amazonaws.com
+            DEPLOYMENTS_AWS_BUCKET: mender-artifacts-int-testing-us
```
merlin-northern, line 22: those variables look ok.

merlin-northern: the names look ok, I am assuming the values are "arbitrary".

lluiscampos: @merlin-northern Yes. I assume the users replace the values :)

alfrunes: You might want to add `DEPLOYMENTS_AWS_FORCE_PATH_STYLE: "false"` as this will no longer be supported by AWS in the future.

PR #940, opened 2020-06-12
MEN-3606: Implement `./demo --enterprise-testing` flag.

kacf:
```
commit cfa1b40e20b2afca30f52d45017524e4bd3dd8c6
Date:   Fri Jun 12 14:53:53 2020

    MEN-3606: Implement `./demo --enterprise-testing` flag.
    
    This required some significant refactoring because with Enterprise we
    have to launch the client in a separate step, instead of together as
    in Open Source. Took the opportunity to also clean up a little and
    standardize on number based truth values (both numbers and strings
    were used before).
    
    This is for internal testing and is not meant for consumption by
    users. The client launching is not fully aligned with the onboarding
    experience, because it instructs you to launch the client in a
    separate step, whereas the demo script can only launch the backend and
    the client in the same step (though inside the script it is two steps,
    as described above).
    
    Changelog: None
    
commit 8f6a9880bbea1f4ad7eaf81dccf883804eac3974
Date:   Fri Jun 12 15:02:22 2020

    tests: Fix infinite loop when demo script returns error.
    
    This is not actually used in the tests, but I hit the bug while
    debugging, so figured I would fix it. The built-in fd reader is much
    smarter than an `iter` with a sentinel.
    
    Changelog: None
    
```

### DIFF
--- a/demo
+++ b/demo
@@ -2,24 +2,6 @@
 
 ./verify-docker-versions
 
-usage() {
-    cat <<EOF
-$(basename $0) [options] docker-options
-
---client    Enable emulated client
-
-All other arguments passed to this command are passed directly to
-docker-compose, if you want to run the demo, run:
-
-'$(basename $0) up'
-EOF
-}
-
-if [[ "$#" -eq 0 ]] || [[ "$1" = "-h" ]] || [[ "$1" = "--help" ]]; then
-    usage
-    exit 1
-fi
-
 # For the integration tests the name of the docker-compose project is generated
 # by pytest. For users, we use the folder name, but strip out a few characters
 # that aren't allowed.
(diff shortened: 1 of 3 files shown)


tests/tests/test_demo_artifact.py
```diff
@@ -71,7 +76,7 @@ def run_demo_script_up(exit_cond=exit_cond):
             logger.info("Started the demo script")
             password = ""
             time.sleep(60)
-            for line in iter(proc.stdout.readline, ""):
+            for line in proc.stdout:
```
lluiscampos, line 79: At least once in the last couple of months, I have seen this test in CI busy in an infinite loop. So thanks for fixing this!

demo
```diff
@@ -211,45 +233,72 @@ done
...
+        echo "is started, when the first user is created. If you wish to start from scratch,"
+        echo "replace \`up\` with \`down -v\` first to reset, then rerun."
     fi
-    RETRIES=$((RETRIES+1))
-    sleep 5
-done
+fi
 
-if [[ $UPLOAD_ARTIFACT == "true" ]]; then
+if [[ $UPLOAD_ARTIFACT -eq 1 ]]; then
```
lluiscampos, line 301: For consistency: use `[ ` here too.

demo
```diff
@@ -211,45 +233,72 @@ done
...
-            echo "If you don't remember the password, you can run '$(basename $0) down' to delete"
-            echo "the old user and rerun '$(basename $0) up' to create a new one."
-            echo "Please note that all data will be deleted from the old demo server."
-            break ;;
-        *) echo "docker exec error: " $retval; exit $retval ;;
-    esac
-    if [[ $RETRIES -ge $RETRY_LIMIT ]]; then
-        echo "Retried $RETRIES times without success. Giving up."
-        exit 1
+    EXISTS_ERROR=5
```
lluiscampos, line 261: So the retcode for existing user is different from Enterprise to OS?

demo
```diff
@@ -172,7 +194,7 @@ docker-compose \
     -f docker-compose.storage.minio.yml \
     -f docker-compose.demo.yml \
     -p ${DOCKER_COMPOSE_PROJECT_NAME} \
-    $CLIENT \
+    $EXTRA_FILES \
```
lluiscampos, line 197: Aren't we missing here `USER_EXTRA_FILES` ? Not that I have a good use case for it (it was used basically for enterprise). But if we keep supporting it in the script I think we should then add it to all `docker-compose` calls.

demo
```diff
@@ -211,45 +233,72 @@ done
...
+                | jq -r .tenant_token)
+        # Now that we have the tenant token we can enable the client.
+        EXTRA_FILES="$EXTRA_FILES $CLIENT_FILES"
+        TENANT_TOKEN=$TENANT_TOKEN docker-compose \
+                -f docker-compose.yml \
+                -f docker-compose.storage.minio.yml \
+                -f docker-compose.demo.yml \
+                -p ${DOCKER_COMPOSE_PROJECT_NAME} \
+                $EXTRA_FILES \
+                "$@" -d mender-client
```
lluiscampos, line 292: Also here: I think we are missing `USER_EXTRA_FILES`.

demo
```diff
@@ -329,5 +393,5 @@ docker-compose \
     -f docker-compose.storage.minio.yml \
     -f docker-compose.demo.yml \
     -p ${DOCKER_COMPOSE_PROJECT_NAME} \
-    $CLIENT \
+    $EXTRA_FILES \
```
lluiscampos, line 396: And here, missing `USER_EXTRA_FILES`.

tests/tests/test_demo_artifact.py
```diff
@@ -170,3 +175,10 @@ def demo_up_down_up(self, run_demo_script):
         # the environment up a second time
         self.demo_artifact_upload(run_demo_script, exit_cond="The user already exists")
         logger.info("Finished")
+
+
+class TestDemoArtifactEnterprise(TestDemoArtifact):
+    """A subclass of the TestDemoArtifact class for testing the demo-Artifact
+    upload in Enterprise mode."""
```
lluiscampos, line 182: :clap:

demo
oleorhagen: I guess the docker-compose calls should be collected into one place somewhow then?

demo
```diff
@@ -211,45 +233,72 @@ done
 
 
 echo "Creating a new user..."
-UPLOAD_ARTIFACT="true"
-RETRIES=0
-while :
-do
+UPLOAD_ARTIFACT=1
```
oleorhagen, line 236: Why the integer booleans, instead of idioms like `true` and `false`?

demo
```diff
@@ -211,45 +233,72 @@ done
...
+PRINT_LOGIN_INFO=0
+PRINT_USER_EXISTS=1
+if [ $ENTERPRISE = 1 ]; then
```
oleorhagen, line 239: arithmetic comparison with `=`

demo
```diff
@@ -211,45 +233,72 @@ done
...
+elif [ $retval -eq $EXISTS_ERROR ]; then
+    # If the user exists, skip uploading the Artifact
+    UPLOAD_ARTIFACT=0
+    PRINT_USER_EXISTS=1
+else
+    echo "docker exec error: " $retval
+    exit $retval
+fi
+
+if [ $ENTERPRISE = 1 -a $CLIENT = 1 ]; then
```
oleorhagen, line 274: arithmetic comparison

demo
```diff
@@ -211,45 +233,72 @@ done
...
+    if [ $(docker ps -q -f name=${DOCKER_COMPOSE_PROJECT_NAME}_mender-client_1 | wc -l) -gt 0 ]; then
```
oleorhagen, line 275: Suggestion:
```bash
if ! $(docker ps -q -f name=${DOCKER_COMPOSE_PROJECT_NAME}_mender-client_1 | wc -l); then
...
```

demo
kacf: Yes, I actually started by trying to unify the two binaries' return codes, but there are TONS of different error codes in both of them, 5 is already used for something else in tenantadm, and the operations don't map well to each other. In the end I gave it up, and took it into account here instead.

demo
```diff
@@ -54,6 +69,14 @@ while [ -n "$1" ]; do
     shift
 done
 
+if [ $CLIENT = 1 ]; then
```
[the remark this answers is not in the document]
kacf: Makes sense to change to arithmetic comparison, but I have never seen that syntax before. This is arithmetic comparison to me:

```
if [ $CLIENT -eq 1 ]; then
```

This is used in a lot of places, so I would favor that instead of something else which does the same thing.

demo
kacf: No particular reason. Somewhat subjectively I would argue that in shell scripts integer style is more common, but I have no data to back that up. In any case, my point here was just to pick one and use only that, and the existing true/false variant was implemented incorrectly (it set the `true` value, but used empty value instead of `false`).

demo
kacf: That will not work, it will evaluate to either `0` or `1` and then try to execute that `0` or `1`.

demo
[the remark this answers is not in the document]
oleorhagen: Just as good :)

demo
oleorhagen: Alright :)

demo
kacf: I'm trying to fix this, but I'm a bit unsure about this one:

```bash
if [[ "$@" == *"-f docker-compose.no-ssl.yml"* ]]
```

I understand what it's trying to do, but this looks like incorrect syntax to me. I can't find a reference in the bash documentation that wild cards actually work on arbitrary strings (except in `case` that is). AFAIK they only work on file names that are on disk. And in that context, is `[[` important, or not? @lluiscampos, @oleorhagen: Do any of you know?

oleorhagen: Hmm, yes, this one struck me as a bit odd also. How about first:
` "$@" -> "$*"` ?
Then:
```bash
if [[ "$*" =~ <regex> ]]; then
```
The intended use-case above I think needs `extglob` enabled: https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html to enable this sort of pattern matching.

However, I might be wrong...

kacf: Hmm, the documentation for that option is quite confusing, it doesn't really explain whether the parentheses are part of the expression or just a way to denote the syntax. And there are zero examples... I think I will just rewrite it into a `case` match instead, which is a bit more well known, and doesn't require a special `shopt` option.

oleorhagen: Sounds good to me :)

kacf:
> * The script is becoming quite big and unruly, like all the docker-compose calls are hard to keep unified, like @lluiscampos pointed out. Should we take a little effort and try and amend this?
> 
> * This is now almost 400 lines of bash, maybe it has actually outgrown bash soon?

I understand what you mean, but I'm not convinced it would be significantly better in a different language. Apart from argument parsing and retrying, there are no loops, and very little repetition, so it's not very "abstractable". It's really just a long recipe.

But you're right that it's kind of unruly. What about refactoring into some functions so one can more easily follow the flow of what's happening? I'll see if I can do that without too much effort.

> * A total of 11(!) contributors. Use of `[` and `[[` all over the place. Some people like integers as booleans, some like strings, and some, yeh. A little bit of both?

I'm cleaning this up!

> * Parsing of arguments is done in multiple places :/

Good point, I will move this.

> * Lots of variables, maybe we should disallow undeclared variables, and collect them into one place?

Good idea, I will do that too!

demo
kacf: Decided to leave it as is after discussing with @oleorhagen.

kacf: Regarding, use of `[[` and `[` in general, @oleorhagen convinced me. The double variant is less portable, but let's stick with it as long as we know we are in bash, since it is more well behaved with regards to quoting, and has more matching options.

oleorhagen:
> > * The script is becoming quite big and unruly, like all the docker-compose calls are hard to keep unified, like @lluiscampos pointed out. Should we take a little effort and try and amend this?
> > * This is now almost 400 lines of bash, maybe it has actually outgrown bash soon?
> 
> I understand what you mean, but I'm not convinced it would be significantly better in a different language. Apart from argument parsing and retrying, there are no loops, and very little repetition, so it's not very "abstractable". It's really just a long recipe.
> 
> But you're right that it's kind of unruly. What about refactoring into some functions so one can more easily follow the flow of what's happening? I'll see if I can do that without too much effort.

Sounds like a good idea to me :)

> 
> > * A total of 11(!) contributors. Use of `[` and `[[` all over the place. Some people like integers as booleans, some like strings, and some, yeh. A little bit of both?
> 
> I'm cleaning this up!
> 
> > * Parsing of arguments is done in multiple places :/
> 
> Good point, I will move this.
> 
> > * Lots of variables, maybe we should disallow undeclared variables, and collect them into one place?
> 
> Good idea, I will do that too!

Then I guess this script will shine soon :100:

kacf:
> Then I guess this script will shine soon

Now it does!

demo
```diff
@@ -182,214 +202,254 @@ exitfunc() {
...
-        JWT=$(curl --silent -k -X POST -u ${USER}:${PASSWORD}\
-            --fail\
-            --connect-timeout 5\
-            $MENDER_SERVER_URI/api/management/v1/useradm/auth/login)
+    # Block until the useradm service returns an HTTP 4xx response
+    local RETRIES=0
+    while :
+    do
+        curl --silent -k -X POST -u ${USER}:${PASSWORD} \
+             --fail\
```
oleorhagen, line 227: Nitpick:
5 spaces

demo
```diff
@@ -182,214 +202,254 @@ exitfunc() {
...
-                    --header "Authorization: Bearer ${JWT}"\
-                    --form "size=${ARTIFACT_SIZE_BYTES}"\
-                    --form "artifact=@${DEMO_ARTIFACT_NAME}"\
-                    $MENDER_SERVER_URI/api/management/v1/deployments/artifacts)
+create_user() {
+    echo "Creating a new user..."
+
+    if [[ $ENTERPRISE -eq 1 ]]; then
+        TENANT_ID=$(docker exec \
+                           ${DOCKER_COMPOSE_PROJECT_NAME}_mender-tenantadm_1 \
```
oleorhagen, line 250: Nitpick:
7 spaces

demo
```diff
@@ -182,214 +202,254 @@ exitfunc() {
...
+    fi
+}
 
-    errout=$(jq '.error' <<< $cout)
+maybe_upload_artifact() {
+    if [[ $UPLOAD_ARTIFACT -eq 1 ]]; then
+
+        local RETRIES=0
+        local retval=0
+        until [[ "$JWT" != "" ]]; do
```
oleorhagen, line 319: I know this is not you (In fact I think it is me), but maybe change to `-n` for consistency?

demo
```diff
@@ -182,214 +202,254 @@ exitfunc() {
...
+maybe_upload_artifact
+print_info
+wait_for_user
+
+# ------------------------------------------------------------------------------
+#
+# We will only get here if the user presses Enter.
+#
+# ------------------------------------------------------------------------------
+follow_logs
```
oleorhagen: Sooooooo nice!!!

demo
```diff
@@ -182,214 +202,254 @@ exitfunc() {
...
+#
+# Start execution
+#
+#-------------------------------------------------------------------------------
+
+check_tools
+parse_args "$@"
+enterprise_client_early_handling
+download_demo_artifact
+platform_tweaks
```
oleorhagen, line 431: Suggestion:
`platform_dependent_setup` ?

demo
```diff
@@ -15,160 +35,160 @@ docker-compose, if you want to run the demo, run:
...
-        break
-    elif [[ "$1" = "up" ]]; then
-        RUN_UP=1
-        # Not a flag, so we should break out of the loop.
-        break
-    else
-        break
+enterprise_client_early_handling() {
+    if [[ $CLIENT -eq 1 ]]; then
+        if [[ $ENTERPRISE -ne 1 ]] || [[ $RUN_UP -ne 1 ]]; then
```
oleorhagen, line 105: Suggestion:
I guess this might as well be an explicit zero comparison?

demo
```diff
@@ -15,160 +35,160 @@ docker-compose, if you want to run the demo, run:
...
+    export INTEGRATION_VERSION=$(git describe --tags --abbrev=0)
+    # Parse the Mender-Artifact version used from the other-components.yml file's image tag
+    export MENDER_ARTIFACT_VERSION=$(awk -F':' '/mendersoftware\/mender-artifact/ {print $3}' other-components.yml)
+    # Parse the mender version from docker-compose.yml mender image's tag
+    export MENDER_VERSION=$(awk -F':' '/mendersoftware\/mender-client/ {print $3}' docker-compose.client.yml)
+    export MENDER_DEB_PACKAGE_VERSION=$MENDER_VERSION
+
+    MENDER_SERVER_URI="https://localhost"
+    # use http when providing no-ssl config
+    if [[ "${ARGS[@]}" == *"-f docker-compose.no-ssl.yml"* ]]
```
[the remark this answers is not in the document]
kacf: Yes! Good catch!

demo
kacf: This is completely inconsistent throughout the whole script, even in the original. I'll make it 8 spaces, double of the base indentation, as long as it's the same concatenated line.

demo
kacf: Yes, good point, this is just confusing.

kacf:
Fixed!

oleorhagen:
> Fixed!

Then all good to go :checkered_flag:

PR #944, opened 2020-06-15
backend tests: devauth tests cleanup

kjaskiewiczz:


tests/tests/test_os_ent_migration.py
```diff
@@ -21,7 +21,7 @@
 from testutils.api.client import ApiClient
 from testutils.infra.cli import CliTenantadm
 import testutils.api.deviceauth as deviceauth_v1
-import testutils.api.deviceauth_v2 as deviceauth_v2
+import testutils.api.devicedevauth as devicedevauth
```
lluiscampos: Where is this `devicedevauth` coming from?

lluiscampos: I see you removed it in a later commit, so probably just a temporary hack for search+replace later...

backend-tests/tests/test_deployments.py
```diff
@@ -10,8 +10,7 @@
 from datetime import datetime, timedelta
```
lluiscampos, line 10: Once we are on it: please add the Copyright notice here

backend-tests/tests/test_rbac.py
```diff
@@ -5,7 +5,6 @@
```
lluiscampos, line 5: And here

backend-tests/tests/test_tenantadm.py
```diff
@@ -5,10 +5,14 @@
```
lluiscampos, line 5: And here :)

backend-tests/tests/test_tenantadm_v2.py
```diff
@@ -8,8 +8,14 @@
```
lluiscampos, line 8: Here too

backend-tests/tests/test_devauth.py
```diff
@@ -12,22 +12,1401 @@
...
+        dev.authsets.append(new_set)
+
+    dev.status = "pending"
+
+    return dev
+
+
+def make_accepted_device_with_multiple_authsets(
+    utoken, num_auth_sets=1, num_accepted=1, tenant_token=""
+):
```
lluiscampos, line 405: Just curious, but what is the motivation for renaming this helper? The other ones (`make_pending_device`, for instance) also seem to work with multiple authsets.

mchalski: +1

tests/tests/test_os_ent_migration.py
kjaskiewiczz: I didn't remove it, I've combined _deviceauth.py_ and _deviceauth_v2.py_ into one _deviceauth.py_ file in testutils/api/

backend-tests/tests/test_devauth.py
kjaskiewiczz: there is already `make_accepted_device` method in the testutils.common and the tests imports it so I had a method name conflict

kjaskiewiczz: and with that method, it's not possible to create a device with multiple authsets

tests/tests/test_os_ent_migration.py
lluiscampos: My point is that `devicedevauth` does did not exist, nor it existed at any point on the Git history. You probably replaced "auth_v2" with "devauth" and hence ended up with this "devicedevauth" here.

PR #945, opened 2020-06-17
[changelog-generator] Accept "Changelog:All" and inline entries

lluiscampos:
This adds to changes to changelog-generator to accommodate commits by
dependabot, which use titles like "Changelog:All : Important info here"

See new entry in the test for better understanding.

### DIFF
--- a/extra/changelog-generator/changelog-generator
+++ b/extra/changelog-generator/changelog-generator
@@ -148,14 +148,20 @@ for repo in repos:
                 title = line
                 title_fetched = True
 
-            match = re.match("^ *Changelog: *(.*)", line, re.IGNORECASE)
-            if match:
+            match_changelog = re.match("^ *Changelog: *(.*)", line, re.IGNORECASE)
+            if match_changelog:
                 if log_entry:
                     add_entry(sha, log_entry)
                     log_entry = ""
                 log_entry_local = False
 
-                if re.match("^(Title|Commit|None)[ .]*$", match.group(1), re.IGNORECASE):
+                # The logic of this regex is:
+                # - Shall start with Title|Commit|All|None
+                # - From here, if [.:]+ then capture the rest in a new group
+                # - Else just expect end line
+                match_keyword = re.match("^(Title|Commit|All|None) *(?:[.:]+ *(.*)$|$)", match_changelog.group(1), re.IGNORECASE)
+
+                if match_keyword:
                     if exclusive_tag_seen:
                         # This doesn't really mean that the tags are exclusive, but rather that it
                         # is quite uncommon to see any of them together, and might indicate a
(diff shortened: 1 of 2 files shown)


lluiscampos:
Stopping the pipeline, this tool is not used there.

lluiscampos:
See also https://github.com/mendersoftware/mendertesting/pull/84

extra/changelog-generator/changelog-generator
```diff
@@ -148,14 +148,15 @@ for repo in repos:
...
-            if match:
+            match_changelog = re.match("^ *Changelog: *(.*)", line, re.IGNORECASE)
+            if match_changelog:
                 if log_entry:
                     add_entry(sha, log_entry)
                     log_entry = ""
                 log_entry_local = False
 
-                if re.match("^(Title|Commit|None)[ .]*$", match.group(1), re.IGNORECASE):
+                match_keyword = re.match("^(Title|Commit|All|None)[ .:]*(.*)$", match_changelog.group(1), re.IGNORECASE)
```
kacf, line 158: Unless we really can't because of dependabot, I think we should require colon after the tag, otherwise this entry will be misinterpreted:

```
Changelog: All crashing bugs fixed
```

You might even consider a test case for this.

extra/changelog-generator/changelog-generator
```diff
@@ -166,15 +167,20 @@ for repo in repos:
                                                  % sha)
                     exclusive_tag_seen = True
 
-                if re.match("^Title[ .]*$", match.group(1), re.IGNORECASE):
-                    log_entry = title
-                elif re.match("^Commit[ .]*$", match.group(1), re.IGNORECASE):
-                    log_entry_commit = True
-                elif re.match("^None[ .]*$", match.group(1), re.IGNORECASE):
-                    log_entry_commit = False
+                    if match_keyword.group(1).lower() == "Title".lower():
```
kacf, line 170: The second `lower()` is very overkill, why not just type `title`? Same for the others.

extra/changelog-generator/changelog-generator
lluiscampos: Yes, I see what you mean, thanks.

I will replace it with: `^(Title|Commit|All|None)[ .:]+(.*)$`

dependabot uses something like:
````
Changelog:All : Bump bla bla bla
```

lluiscampos: @kacf It took me a while to fine tune this regex... I added more test cases that I hope cover the relevant corner cases.

extra/changelog-generator/changelog-generator
```diff
@@ -148,14 +148,20 @@ for repo in repos:
...
                     add_entry(sha, log_entry)
                     log_entry = ""
                 log_entry_local = False
 
-                if re.match("^(Title|Commit|None)[ .]*$", match.group(1), re.IGNORECASE):
+                # The logic of this regex is:
+                # - Shall start with Title|Commit|All|None
+                # - From here, if [.:]+ then capture the rest in a new group
+                # - Else just expect end line
+                match_keyword = re.match("^(Title|Commit|All|None) *([.:]+)? *((?(2).*$|$))", match_changelog.group(1), re.IGNORECASE)
```
kacf, line 162: I actually have not seen the `(?(` notation before, I had to look it up.

Question is, is this an easier way to express the same thing?

`^(Title|Commit|All|None) *(?:[.:]+ *(.*)$|$))`

If it doesn't work, I'm fine with the original expression.

extra/changelog-generator/changelog-generator
```diff
@@ -166,15 +172,20 @@ for repo in repos:
...
-                    log_entry_commit = True
-                elif re.match("^None[ .]*$", match.group(1), re.IGNORECASE):
-                    log_entry_commit = False
+                    if match_keyword.group(1).lower() == "title":
+                        log_entry = title
+                    elif match_keyword.group(1).lower() == "none":
+                        log_entry_commit = False
+                    elif match_keyword.group(1).lower() in ["commit", "all"]:
+                        log_entry_commit = True
+                        if match_keyword.group(3):
```
kacf, line 181: If you change the expression above, this will be group 2.

extra/changelog-generator/changelog-generator
```diff
@@ -166,15 +172,20 @@ for repo in repos:
...
+                            # Log the rest of the line
+                            if commit_msg:
+                                commit_msg += "\n"
+                            commit_msg += match_keyword.group(3)
```
kacf, line 185: And here.

extra/changelog-generator/changelog-generator
lluiscampos: Your regex also works (all tests pass, which is all the evidence I need). Thank you!

PR #951, opened 2020-07-02
2.3.x update for release tool

kacf:
Should mainly affect release process, but I'm letting the tests run just in case.


kacf:
Yep, good point, done!

PR #915, opened 2020-05-20
[staging] Update integration from latest master

lluiscampos:
Brings tests updates and `release_tool.py` changes related to versioning.

Follow-up from #901.

PR #956, opened 2020-07-02
Fix TestUpdateModules for old Mender client

lluiscampos:
For Mender client 2.2.x the warnings are printed into stdout, so the
test receives the warning of not being able to connect to syslog
together with the desired output.


lluiscampos:
@kacf This is the fix I was talking you about. Technically speaking we don't need it in master, only in `integration/2.3.x` and `integration/staging` but I don't think it hurts anyway...

lluiscampos:
Correct. I didn't see the original fix. Closing. I will cherry-pick #949 to #915 instead.

PR #957, opened 2020-07-03
Fix production/run script parsing of versions

lluiscampos:
The files are in the parent directory.


lluiscampos:
I need to cherry-pick this to `2.4.x` before the release too.

lluiscampos:
> How did we not catch this before! astonished

Because we don't have a good test for this. I created a new ticket: [QA-202](https://tracker.mender.io/browse/QA-202)

PR #960, opened 2020-07-06
Fix production/run script parsing of versions - 2.4.x

lluiscampos:
Cherry-picking #957


lluiscampos:
Stopping integration pipeline.

PR #963, opened 2020-07-14
introduce new, private repository - auditlogs

kjaskiewiczz:


docker-compose.auditlogs.yml
```diff
@@ -0,0 +1,12 @@
+version: '2.1'
+services:
+
+    #
+    # auditlogs
+    #
+    mender-auditlogs:
+        image: registry.mender.io/mendersoftware/auditlogs:mender-master
```
kacf, line 8: Why not in `docker-compose.enterprise.yml`?

kjaskiewiczz: we don't want it in every enterprise setup; on-prem enterprise customer may not want to have audit logs at all
we'll introduce configuration in other services that will enable/disable audit logging and I was thinking of setting the appropriate envs here (when we'll add the configuration)

kacf: :+1:

PR #966, opened 2020-07-22
Update release_tool.py to use saas-vYYYY.MM.DD tags by default for hM

tranchitella:
Changelog: none


lluiscampos:
Thanks @tranchitella! This was in my "Back from vacation TODO" list :)

PR #976, opened 2020-08-03
[integration staging] Update mender-cli to latest stable

lluiscampos:


lluiscampos:
I will follow-up with the broken tests independently. Merging.

PR #977, opened 2020-08-03
inventory search endpoint tests + deployments tests refactor

kjaskiewiczz:


backend-tests/tests/test_inventory.py
```diff
@@ -108,48 +114,11 @@ def make_pending_device(utoken, tenant_token=""):
     return dev
```
lluiscampos, line 115: Se can remove `make_pending_device` from this file too.

backend-tests/tests/test_inventory.py
```diff
@@ -77,6 +82,7 @@ def tenants_users(clean_migrated_mongo_mt):
         username = "user@%s.com" % n
         password = "correcthorse"
         tenant = create_org(n, username, password)
+        tenants.append(tenant)
```
lluiscampos, line 85: So this was not working at all before? :exploding_head:

backend-tests/tests/test_inventory.py
```diff
@@ -60,12 +65,12 @@ def clean_migrated_mongo_mt(clean_mongo):
...
+@pytest.fixture(scope="function")
 def user(clean_migrated_mongo):
     yield create_user("user-foo@acme.com", "correcthorse")
 
 
-@pytest.yield_fixture(scope="function")
+@pytest.fixture(scope="function")
 def tenants_users(clean_migrated_mongo_mt):
     cli = CliTenantadm()
     api = ApiClient(tenantadm.URL_INTERNAL)
```
lluiscampos, line 76: Unused

backend-tests/tests/test_inventory.py
```diff
@@ -321,3 +293,278 @@ def test_fail_no_attr_value(self, user):
...
+                            "attribute": "idx",
+                            "value": 1,
+                            "scope": "inventory",
+                        }
+                    ],
+                },
+                "status_code": 200,
+                "response": [
+                    {
+                        "id": str(tenants[0].devices[1].id),
```
lluiscampos, line 381: For consistency,
```suggestion
                        "id": str(tenant.devices[1].id),
```

backend-tests/tests/test_inventory.py
```diff
@@ -321,3 +293,278 @@ def test_fail_no_attr_value(self, user):
...
+                assert len(body) == len(test_case["response"]), (
+                    "Unexpected number of results: %s != %s"
+                    % (test_case["response"], body)
+                )
+
+                if len(body) > 0:
+                    for dev in test_case["response"]:
+                        for api_dev in body:
+                            if dev["id"] == api_dev["id"]:
+                                assert_device_attributes(dev, api_dev)
```
lluiscampos, line 562: We need to assert that all `dev` devices are present in `api_dev`.

If for some strange reason you get the same number of results but no match in any of the devices, then this would result in a false positive.

backend-tests/tests/test_inventory.py
```diff
@@ -321,3 +293,278 @@ def test_fail_no_attr_value(self, user):
...
+        NUM_DEVICES = 100
+
+        useradmm = ApiClient(useradm.URL_MGMT)
+        devauthd = ApiClient(deviceauth.URL_DEVICES)
+        devauthm = ApiClient(deviceauth.URL_MGMT)
+        invd = ApiClient(inventory.URL_DEV)
+        invm_v2 = ApiClient(inventory_v2.URL_MGMT)
+
+        # Initialize devices and API tokens
+        tenants = [t for t in tenants_users]
```
lluiscampos, line 327: Unnecessary, we can use `tenant_users`, right?

backend-tests/tests/test_inventory.py
```diff
@@ -321,3 +293,278 @@ def test_fail_no_attr_value(self, user):
...
+            {"version": "v3.0", "grp1": "baz", "idx": 2},
+            {"version": "v1.0", "grp2": "foo", "idx": 3},
+            {"version": "v2.0", "grp2": "bar", "idx": 4},
+            {"version": "v3.0", "grp2": "baz", "idx": 5},
+            {"version": "v1.0", "grp3": "foo", "idx": 6},
+            {"version": "v2.0", "grp3": "bar", "idx": 7},
+            {"version": "v3.0", "grp3": "baz", "idx": 8},
+        ]
+
+        # Setup an identical inventory environment for both tenants.
```
lluiscampos, line 341: Why? There is only one tenant in use, that I can see. I would define here `tenant = tenants[0]` and skip the for loop.

kjaskiewiczz:
@lluiscampos thanks for the review!
yes, I was originally planning to open the second PR where I will move some tests form integration to backend integration (as mentioned on standup);
 
some of the tests, like `test_clients_exclusive_to_user`, are using `enterprise_no_client` but during the test, we create a client using `new_tenant_client` methods; my understanding was that the tests which are using the client should stay in the integration;

I'll take another look at the test_inventory and address your comments; tbh I just took it from a feature branch (we forgot to merge it at some point) and, since it was already reviewed, I didn't bother to review it once again, but after looking at your comments I know I should do that;

lluiscampos:
> @lluiscampos thanks for the review!
> yes, I was originally planning to open the second PR where I will move some tests form integration to backend integration (as mentioned on standup);
> 

Right, you said so. Sorry.

> some of the tests, like `test_clients_exclusive_to_user`, are using `enterprise_no_client` but during the test, we create a client using `new_tenant_client` methods; my understanding was that the tests which are using the client should stay in the integration;
> 

Actually, this specific test does not create "devices", just emulates clients via APIs. But in general you are right in that there are some tests that use `enterprise_no_client` and then create a client/device.

> I'll take another look at the test_inventory and address your comments; tbh I just took it from a feature branch (we forgot to merge it at some point) and, since it was already reviewed, I didn't bother to review it once again, but after looking at your comments I know I should do that;

Just apply the comments that you find valuable. Nothing of what I said is critical.

backend-tests/tests/test_inventory.py
kjaskiewiczz: fixed

backend-tests/tests/test_inventory.py
kjaskiewiczz: it looks so...

backend-tests/tests/test_inventory.py
kjaskiewiczz: fixed

backend-tests/tests/test_inventory.py
kjaskiewiczz: fixed

backend-tests/tests/test_inventory.py
kjaskiewiczz: fixed

backend-tests/tests/test_inventory.py
kjaskiewiczz: yes, fixed

backend-tests/tests/test_inventory.py
kjaskiewiczz: fixed

kjaskiewiczz:
apart of addressing comments, PR includes an additional commit which moves test_artifacts_exclusive_to_user from integration to backend integration

PR #983, opened 2020-08-10
MEN-3828: Merge "saved-filters" feature branch to master

alfrunes:


backend-tests/tests/test_inventory.py
```diff
@@ -529,14 +826,539 @@ def test_search_v2(self, tenants_users):
...
+            )
+
+    def test_saved_filters_dynamic(self, clean_mongo, mongo):
+        """
+        Check that the saved filters return the correct set of
+        devices when the inventory changes dynamically. That is,
+        when devices modify their inventory or get removed entirely.
+        """
+
+        devauthd = ApiClient(deviceauth.URL_DEVICES)
```
lluiscampos, line 1108: Unused

backend-tests/tests/test_inventory.py
```diff
@@ -529,14 +826,539 @@ def test_search_v2(self, tenants_users):
...
+            devs_recv = sorted([dev["id"] for dev in rsp.json()])
+            devs_exct = sorted(
+                [dev.id for dev in filter(test_case["filter"], tenant.devices)]
+            )
+            assert devs_recv == devs_exct, (
+                "Unexpected device set returned by saved filters, "
+                + "expected: %s, received: %s" % (devs_recv, devs_exct)
+            )
+
+            # Perform perturbations to the device set.
```
lluiscampos, line 1312: :+1:

backend-tests/tests/test_inventory.py
```diff
@@ -529,14 +826,539 @@ def test_search_v2(self, tenants_users):
...
+            devs_recv = sorted([dev["id"] for dev in rsp.json()])
+            devs_exct = sorted(
+                [dev.id for dev in filter(test_case["filter"], tenant.devices)]
+            )
+            assert devs_recv == devs_exct, (
+                "Unexpected device set returned by saved filters, "
+                + "expected: %s, received: %s" % (devs_recv, devs_exct)
+            )
+
+            mongo_cleanup(mongo)
```
lluiscampos, line 1361: Isn't this done automatically by `clean_mongo` fixture?

backend-tests/tests/test_inventory.py
```diff
@@ -529,14 +826,539 @@ def test_search_v2(self, tenants_users):
...
+                            "value": False,
+                        },
+                    ],
+                },
+                "status_codes": [201, 200, 200, 200],
+                "result": [
+                    {"id": dev.id, "attributes": dict_to_inventoryattrs(dev.inventory),}
+                    for dev in filter(
+                        lambda dev: "py3" not in dev.inventory
+                        and 2 <= dev.inventory["idx"] < 8
```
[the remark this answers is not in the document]
lluiscampos: This is important; the rest of the review is minor stuff.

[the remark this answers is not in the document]
alfrunes: You're right. The test passes because device at index 8 is excluded by another filtering operation, I'm changing this test-case because it doesn't make much sense after I changed the device set.

backend-tests/tests/test_inventory.py
```diff
@@ -529,14 +826,539 @@ def test_search_v2(self, tenants_users):
...
+                    "terms": [
+                        {
+                            "type": "$ne",
+                            "scope": "inventory",
+                            "attribute": "artifact",
+                            "value": "v2",
+                        }
+                    ],
+                },
+                "status_codes": [409, 200, 404, 404],
```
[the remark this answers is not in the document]
alfrunes: Yeah, this is not obvious at all. However, I do want to include a case for getting/executing a filter that does not exist. I'll add a comment explaining that `filter_id` will be replaced by "foobar" if the first call is not successful.

backend-tests/tests/test_inventory.py
alfrunes: This was just a work-around running mongo_cleanup for each test case. (`clean_mongo` only runs before and after all test cases).

backend-tests/tests/test_inventory.py
```diff
@@ -281,60 +320,285 @@ def logger(self):
...
 
-            attrs = self.dict_to_inventoryattrs(inv)
-            rsp = invd.with_auth(device.token).call(
-                "PATCH", inventory.URL_DEVICE_ATTRIBUTES, body=attrs
-            )
-            assert rsp.status_code == 200
-            device.inventory = inv
+        invm_v2 = ApiClient(inventory_v2.URL_INTERNAL)
 
         test_cases = [
```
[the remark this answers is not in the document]
alfrunes: This has changed since the tests were written, now search is also available in open-source. I changed the tenant for the first test case to the open-source tenant just to include all plans.

backend-tests/tests/test_inventory.py
lluiscampos: Of course. I failed to see that this was per test case.

PR #985, opened 2020-08-17
MEN-2894: Make integration test which tests old clients with new server

alfrunes:
Added a new compatibility docker-compose setup which runs all clients within this major release.
Made integration tests verifying the expected client behavior (viewed from the API), that is:

- Setup test environment with one instance of each client defined in extra/integration-testing/docker-compose.compat.yml
- Accept all devices and check that all devices updates their inventory after authorizing.
- Create a deployment with a rootfs artifact with an empty file -- effectively performing a noop update to the identical inactive partition.
- Verify that all devices gets "updated" successfully.

### DIFF
--- a/backend-tests/tests/test_account_suspension.py
+++ b/backend-tests/tests/test_account_suspension.py
@@ -69,7 +69,7 @@ def tenants_users_devices(tenants, mongo):
 
 class TestAccountSuspensionEnterprise:
     def test_user_cannot_log_in(self, tenants):
-        tc = ApiClient(tenantadm.URL_INTERNAL)
+        tc = ApiClient(tenantadm.URL_INTERNAL, host=tenantadm.HOST, schema="http://")
 
         uc = ApiClient(useradm.URL_MGMT)
 
@@ -106,7 +106,7 @@ class TestAccountSuspensionEnterprise:
             assert r.status_code == 200
 
     def test_authenticated_user_is_rejected(self, tenants):
-        tc = ApiClient(tenantadm.URL_INTERNAL)
+        tc = ApiClient(tenantadm.URL_INTERNAL, host=tenantadm.HOST, schema="http://")
         uc = ApiClient(useradm.URL_MGMT)
         dc = ApiClient(deviceauth.URL_MGMT)
 
@@ -141,7 +141,7 @@ class TestAccountSuspensionEnterprise:
         dacd = ApiClient(deviceauth.URL_DEVICES)
         devauthm = ApiClient(deviceauth.URL_MGMT)
         uc = ApiClient(useradm.URL_MGMT)
-        tc = ApiClient(tenantadm.URL_INTERNAL)
+        tc = ApiClient(tenantadm.URL_INTERNAL, host=tenantadm.HOST, schema="http://")
 
         # accept a dev
         device = tenants_users_devices[0].devices[0]
@@ -182,7 +182,7 @@ class TestAccountSuspensionEnterprise:
         dacd = ApiClient(deviceauth.URL_DEVICES)
         devauthm = ApiClient(deviceauth.URL_MGMT)
         uc = ApiClient(useradm.URL_MGMT)
-        tc = ApiClient(tenantadm.URL_INTERNAL)
+        tc = ApiClient(tenantadm.URL_INTERNAL, host=tenantadm.HOST, schema="http://")
         dc = ApiClient(deployments.URL_DEVICES)
 
         # accept a dev
--- a/demo
+++ b/demo
@@ -126,7 +126,7 @@ download_demo_artifact() {
 platform_dependent_setup() {
     if [[ "$OSTYPE" == "darwin"* ]]; then
         ARTIFACT_SIZE_BYTES=$(stat -f %z ${DEMO_ARTIFACT_NAME}) # BSD is not GNU -_-
-        export GATEWAY_IP=$(ifconfig | sed -En 's/127.0.0.1//;s/.*inet (addr:)?(([0-9]*\.){3}[0-9]*).*/\2/p' | head -1)
+        export GATEWAY_IP=$(ifconfig $(netstat -rn | grep default | head -1 | awk '{print($NF)}') inet | grep -F 'inet '| sed -e 's/.*inet \([^ ]*\) .*/\1/')
     else
         ARTIFACT_SIZE_BYTES=$(stat -c %s ${DEMO_ARTIFACT_NAME})
         export GATEWAY_IP=$(ip route get 1 | awk '{print $7;exit}')
--- a/extra/integration-testing/docker-compose.compat.yml
+++ b/extra/integration-testing/docker-compose.compat.yml
@@ -1,0 +1,44 @@
+version: '2.1'
+services:
+  #
+  # Setup including older clients: 2.3 -> 2.0
+  #
+  mender-client-2-3:
+    image: mendersoftware/mender-client-qemu:2.3
+    networks:
+       - mender
+    stdin_open: true
+    tty: true
+    privileged: true
+    environment:
+      TENANT_TOKEN: ""
+
+  mender-client-2-2:
+    image: mendersoftware/mender-client-qemu:2.2
+    networks:
+       - mender
+    stdin_open: true
+    tty: true
+    privileged: true
+    environment:
+      TENANT_TOKEN: ""
+
+  mender-client-2-1:
+    image: mendersoftware/mender-client-qemu:2.1
+    networks:
+       - mender
+    stdin_open: true
+    tty: true
+    privileged: true
+    environment:
+      TENANT_TOKEN: ""
+
+  mender-client-2-0:
+    image: mendersoftware/mender-client-qemu:2.0
+    networks:
+       - mender
+    stdin_open: true
+    tty: true
+    privileged: true
+    environment:
+      TENANT_TOKEN: ""
--- a/tests/tests/test_compat.py
+++ b/tests/tests/test_compat.py
@@ -1,0 +1,296 @@
+# Copyright 2020 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        https://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import pytest
+import time
+
+from datetime import datetime, timedelta, timezone
+
+from testutils.util.artifact import Artifact
+from testutils.infra.container_manager import factory
+from testutils.infra.device import MenderDevice
+from testutils.common import create_org, create_user
+from testutils.api.client import ApiClient
+from testutils.api import deviceauth, useradm, inventory, deployments
+
+from .. import conftest
+
+container_factory = factory.get_factory()
+
+TIMEOUT = timedelta(minutes=5)
+
+
+@pytest.fixture(scope="function")
+def setup_os_compat(request):
+    env = container_factory.getCompatibilitySetup()
+    request.addfinalizer(env.teardown)
+    env.setup()
+
+    env.user = create_user(
+        "test@mender.io", "correcthorse", containers_namespace=env.name
+    )
+    env.populate_clients()
+
+    clients = env.get_mender_clients()
+    assert len(clients) > 0, "Failed to setup clients"
+    env.devices = []
+    for client in clients:
+        dev = MenderDevice(client)
+        dev.ssh_is_opened()
+        env.devices.append(dev)
+
+    return env
+
+
+@pytest.fixture(scope="function")
+def setup_ent_compat(request):
+    env = container_factory.getCompatibilitySetup(enterprise=True)
+    request.addfinalizer(env.teardown)
+    env.setup()
+
+    env.tenant = create_org(
+        "Mender",
+        "test@mender.io",
+        "correcthorse",
+        containers_namespace=env.name,
+        container_manager=env,
+    )
+    env.user = env.tenant.users[0]
+
+    env.populate_clients(tenant_token=env.tenant.tenant_token)
+
+    clients = env.get_mender_clients()
+    assert len(clients) > 0, "Failed to setup clients"
+    env.devices = []
+    for client in clients:
+        dev = MenderDevice(client)
+        dev.ssh_is_opened()
+        env.devices.append(dev)
+
+    return env
+
+
+def accept_devices(api_deviceauth, devices=None):
+    """
+    Update the device status for the given set of devices to "accepted"
+
+    :param api_deviceauth: testutils.api.client.ApiClient setup and authorized
+                           to use the deviceauth management api,
+                           i.e. api_client.with_auth(api_token)
+    :param devices: list of dict-type devices as returned by
+                    GET /api/management/v1/devauth/devices
+                    If left None, all pending devices are accepted.
+    """
+    if devices is None:
+        rsp = api_deviceauth.call(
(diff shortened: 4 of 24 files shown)


testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -335,7 +334,7 @@ def __init__(
 
 class DockerComposeEnterpriseSetup(DockerComposeNamespace):
     def __init__(self, name, num_clients=0):
-        self.num_clients = num_clients
+        self.num_clientt = num_clients
```
lluiscampos, line 337: Typo?

tests/tests/test_compat.py
```diff
@@ -0,0 +1,291 @@
...
+    This class contains compatibility tests for assessing server compatibility
+    with older clients.
+    """
+
+    def test_open_source_compatibility(self, setup_os_compat):
+        self.compatibility_test_impl(setup_os_compat)
+
+    def test_enterprise_compatibility(self, setup_ent_compat):
+        self.compatibility_test_impl(setup_ent_compat)
+
```
lluiscampos, line 200: There is a test suite selection in the integration tests. The enterprise ones need `Enterprise` in the class name.

lluiscampos:
I will review this properly tomorrow. My first impression is good though :smiley:

tests/tests/test_compat.py
```diff
@@ -0,0 +1,295 @@
...
+                "GET", deployments.URL_DEPLOYMENTS_STATISTICS.format(id=deployment_id),
+            )
+            assert rsp.status_code == 200
+            assert rsp.json()["failure"] == 0
+            assert rsp.json()["success"] == dpl["device_count"]
+            break
+        elif datetime.now() > deadline:
+            pytest.fail("timeout: Waiting for devices to update")
+        else:
+            time.sleep(1)
```
lluiscampos, line 186: You could also do while/else:

```suggestion
    while datetime.now() <= deadline:
        rsp = api_deployments.call(
            "GET", deployments.URL_DEPLOYMENTS_ID.format(id=deployment_id)
        )
        assert rsp.status_code == 200

        dpl = rsp.json()
        if dpl["status"] == "finished":
            rsp = api_deployments.call(
                "GET", deployments.URL_DEPLOYMENTS_STATISTICS.format(id=deployment_id),
            )
            assert rsp.status_code == 200
            assert rsp.json()["failure"] == 0
            assert rsp.json()["success"] == dpl["device_count"]
        else:
            time.sleep(1)
    else:
        pytest.fail("timeout: Waiting for devices to update")
    
```

lluiscampos: It is just a suggestion, but if you decide to apply follow-up on while loop in `compatibility_test_impl` too.

tests/tests/test_compat.py
```diff
@@ -0,0 +1,295 @@
...
+    This class contains compatibility tests implementation for assessing
+    server compatibility with older clients.
+    """
+
+    def compatibility_test_impl(self, env):
+        """
+        The actual test implementation:
+         - Accept devices
+         - Verify devices patches inventory
+         - Perform a noop rootfs update and verify the update was successful
```
lluiscampos, line 200: I think we should do a full rootfs update and check from the clients side that it did reboot. I don't have evidence to point that just exercising the API is not enough, but it just feels so weak as a test. Can you convince me that this is enough?

The added complexity then would be to extract the filesystem from the container to prepare the Artifact (currently we do that for master images in `run.sh` wrapper). Then you could feed it to the Artifact class and still keep using `testutils.api.client.ApiClient` through the test.

The test is very much designed as a backend only test, so I understand that a "noop" update is easier, but if you give up with the we do have helpers in the integration test framework to do the work for us (of course this will

tests/tests/test_compat.py
```diff
@@ -0,0 +1,295 @@
...
+        URL_MGMT_USERADM = useradm.URL_MGMT.replace("mender-api-gateway", gateway_addr)
+        URL_MGMT_DEVAUTH = deviceauth.URL_MGMT.replace(
+            "mender-api-gateway", gateway_addr
+        )
+        URL_MGMT_INVNTRY = inventory.URL_MGMT.replace(
+            "mender-api-gateway", gateway_addr
+        )
+        URL_MGMT_DPLMNTS = deployments.URL_MGMT.replace(
+            "mender-api-gateway", gateway_addr
+        )
```
lluiscampos, line 212: Wouldn't it be nice that `testutils.client.GATEWAY_URL` could be parametrized? But then, maybe that is that it lead to the current `MenderAPI` class with global scope...

testutils/util/artifact.py
```diff
@@ -55,7 +75,21 @@ def __init__(
         if payload is not None:
             self.add_payload(payload, payload_type, depends, provides)
 
-    def add_payload(self, fd, payload_type="rootfs_image", depends=None, provides=None):
+    def add_state_script(self, state, script):
```
lluiscampos, line 78: This is fine, but unrelated to the PR. Why?

tests/tests/test_compat.py
lluiscampos: This is something I would like to discuss further. Here in the test is okey, but putting a similar `replace` in `testutils.common.create_org` is rather hacky.

It goes all the way to the base design of `client.ApiClient` class, where the full url is expected to be passed to the constructor, but then these URLs are taken from the micro service itself  prepending a GATEWAY_URL that comes from the `client`. So this is a circular dependency that we should break.

A backwards compatible solution could be to add some methods in the micro services API files like:

```
def url_mgmt(base_url=testutils.api.client.GATEWAY_URL):
    return base_url + "/api/management/v1/useradm"

URL_MGMT = management_url()
```

So that backend-tests could use `URL_MGMT` while integration tests can use `url_mgmt(custom_gateway)`.

This is also an step forward for backend-test in staging environment project ;)

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -119,7 +121,7 @@ def _docker_compose_cmd(self, arg_list, env=None):
 
                 except subprocess.CalledProcessError as e:
                     logger.info(
-                        "failed to run docker-compose: error follows:\n%s" % (e.output)
+                        'failed to run "%s": error follows:\n%s' % (cmd, e.output)
```
lluiscampos, line 124: Thanks :+1:

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -138,7 +140,6 @@ def _wait_for_containers(self, expected_containers):
             )
             running_countainers_count = len(out.split())
             if running_countainers_count == expected_containers:
-                time.sleep(60)
```
lluiscampos, line 141: This I don't feel confident with merging, as it is one of this sleeps that removing them made tests flaky. I suggest you submit this in a separate PR and run the full test suite like 10 times.

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -280,9 +281,7 @@ def __init__(self, name, num_clients=1):
             DockerComposeNamespace.__init__(self, name, self.QEMU_CLIENT_FILES)
 
     def setup(self):
-        self._docker_compose_cmd("up -d")
-        if self.num_clients > 1:
-            self._docker_compose_cmd("scale mender-client=%d" % self.num_clients)
+        self._docker_compose_cmd("up -d --scale mender-client=%d" % self.num_clients)
```
lluiscampos, line 284: Thanks!

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -383,6 +382,64 @@ def setup(self):
...
+        client_services = self.client_services()
+        compose_cmd = "run -d"
+        if tenant_token != "":
+            compose_cmd += " -e TENANT_TOKEN={tkn}".format(tkn=tenant_token)
+        if name is not None:
+            compose_cmd += " --name '{name}'".format(name=name)
+
+        compose_cmd += " {service}"
+
+        for i in range(replicas):
```
lluiscampos, line 417: Style:

```suggestion
        for _ in range(replicas):
```

tests/tests/test_compat.py
```diff
@@ -0,0 +1,295 @@
...
+        "test@mender.io", "correcthorse", containers_namespace=env.name
+    )
+    env.populate_clients()
+
+    clients = env.get_mender_clients()
+    env.devices = []
+    for client in clients:
+        dev = MenderDevice(client)
+        dev.ssh_is_opened()
+        env.devices.append(dev)
```
lluiscampos, line 50: As the code is now, these are unused. It only makes sense to have it if we have client-side checks for the test.

demo
```diff
@@ -126,7 +126,7 @@ download_demo_artifact() {
 platform_dependent_setup() {
     if [[ "$OSTYPE" == "darwin"* ]]; then
         ARTIFACT_SIZE_BYTES=$(stat -f %z ${DEMO_ARTIFACT_NAME}) # BSD is not GNU -_-
-        export GATEWAY_IP=$(ifconfig | sed -En 's/127.0.0.1//;s/.*inet (addr:)?(([0-9]*\.){3}[0-9]*).*/\2/p' | head -1)
+        export GATEWAY_IP=$(ifconfig $(netstat -rn | grep default | head -1 | awk '{print($NF)}') inet | grep -F 'inet '| sed -e 's/.*inet \([^ ]*\) .*/\1/')
```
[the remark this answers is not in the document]
alfrunes: I must've accidentally changed the hash of that commit while rebasing.

extra/integration-testing/docker-compose.compat.yml
```diff
@@ -0,0 +1,44 @@
+version: '2.1'
+services:
+  #
+  # Setup including older clients: 2.3 -> 2.0
+  #
```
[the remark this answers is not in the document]
alfrunes: ... but isn't 2.3 the integration version 2.4? There are no 2.4 image on docker hub. Or am I missing something?

tests/tests/test_compat.py
alfrunes: I just interpreted the ticket as making a test that make sure that the server never breaks the API for older clients. I can make the client checks more verbose, but it'll come at a cost of significantly increased execution time.

tests/tests/test_compat.py
alfrunes: Yeah, this is pretty ugly. Maybe we could reduce the `URL_{MGMT,DEVICE}` to only hold the base url path and let the `gateway_host` be a separate argument (defaulting to "mender-api-gateway") when initializing the ApiClient class?

testutils/util/artifact.py
alfrunes: I was thinking of making a state-script no-op artifact, but figured out an easier way once I had already implemented it.

extra/integration-testing/docker-compose.compat.yml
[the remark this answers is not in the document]
lluiscampos: You are completely right! I got confused, sorry.

tests/tests/test_compat.py
lluiscampos: That is also a valid idea and will result in a cleaner interface, I think.

You would then need to modify integration test `test_os_ent_migration` [here](https://github.com/mendersoftware/integration/blob/master/tests/tests/test_os_ent_migration.py#L92).

tests/tests/test_compat.py
lluiscampos: I don't have a good argument, really. It might be just enough as-is and it is me being biased by the principle of "an integration test must do stuff on the client".

@kacf Do you have an opinion on this?

kacf: > I don't have a good argument, really. It might be just enough as-is and it is me being biased by the principle of "an integration test must do stuff on the client".

Well, if I understand it correctly, this *is* doing things on the client, it's just a write optimization. I think the execution time will not be affected that much by doing a real update though. By far the longest time in client tests is the initialization, initial boot, key generation and first login. The rootfs writing is over literally in seconds (this is x86 and HW accelerated), and the reboot you have to do regardless.

@alfrunes is right that it is primarily the API compatibility which is being tested here, so the type of update is not that important. Personally I considered using a `single-file` update, which doesn't require rebooting. I think it boils down to what is easiest and most maintainable in the test framework, so from my PoV a noop update is fine as long as it doesn't complicate things too much. I think my biggest unease would be that it is a bit hacky, but not so much that I would object to doing it! :-)

kacf (commented):
I didn't review the framework changes very deeply, although I did take a quick look and it looks alright to me. I assume Lluis has looked at that in more detail.

tests/tests/test_compat.py
lluiscampos: Just to clarify; I know that the test does an update, what I meant is that we don't verify it from the client point of view. For me that is the key difference between integration tests and pure API tests, that in the integration ones we do use the SSH interface to verify the client (what I consider the lower tester).

But then, to use the existing helpers (for the update, or to check the reboot) we would need to extract the file systems from the Docker images of the old clients; which complicates this test framework required for this test. So the noop rootfs update is quite smart, I think.

I am okey with leaving it as is.

testutils/common.py
```diff
@@ -127,12 +127,27 @@ def create_user(name, pwd, tid="", containers_namespace="backend-tests"):
...
+    )
     user_id = None
     tenant_id = cli.create_org(name, username, password, plan=plan)
     tenant_token = json.loads(cli.get_tenant(tenant_id))["tenant_token"]
-    api = ApiClient(useradm.URL_MGMT)
+    url_useradm = useradm.URL_MGMT
+    if container_manager is not None:
+        url_useradm = url_useradm.replace(
+            "mender-api-gateway",
+            container_manager.get_ip_of_service("mender-api-gateway")[0],
```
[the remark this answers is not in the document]
alfrunes: The ApiClient certificate verification.

testutils/infra/container_manager/docker_compose_manager.py
alfrunes: I reverted this, perhaps we can use the new liveliness endpoints to wait for the back-end to start?

tests/tests/test_compat.py
alfrunes: I did not know this was a think in python. However, I prefer having it as is since I find it easier to read as a do-while loop.

alfrunes:
@lluiscampos, @kacf I added a reboot check for all the clients and did some refactoring of `testutils.api` module to easier integrate with full integration tests. I'm keeping the noop artifact for now since I don't believe it adds any coverage by cloning the clients' rootfs.

tests/tests/test_compat.py
lluiscampos: I think the while/else is very Pythony and reads good. But I respect your preference :)

lluiscampos:
> I'm keeping the noop artifact for now since I don't believe it adds any coverage by cloning the clients' rootfs.

Makes sense :+1:

tests/tests/test_compat.py
```diff
@@ -0,0 +1,333 @@
...
+        else:
+            time.sleep(1)
+
+
+def _watch_clients(host_addr, clients, timeout=TIMEOUT):
+    """
+    Watches for clients to reboot in separate threads
+    """
+
+    class _RebootThread(threading.Thread):
```
lluiscampos, line 197: In a later iteration, this could be moved to `testutils/infra/device.py` and be reused by other tests.

lluiscampos (approved):
The testutils refactor and the reboot detection from client side look great to me. Nice job.

Some comments below but I think it looks good enough for merging.

backend-tests/tests/test_account_suspension.py
```diff
@@ -69,7 +69,7 @@ def tenants_users_devices(tenants, mongo):
 
 class TestAccountSuspensionEnterprise:
     def test_user_cannot_log_in(self, tenants):
-        tc = ApiClient(tenantadm.URL_INTERNAL)
+        tc = ApiClient(tenantadm.URL_INTERNAL, host=tenantadm.HOST, schema="http://")
```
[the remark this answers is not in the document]
alfrunes: I'm merging this PR, but I can follow up with a separate PR with this small refactor.

PR #986, opened 2020-08-19
MEN-3636: Combine dependabot entries into a single changelog entry

oleorhagen:
Sad that this took this long, but I was struggling with understanding this script at first.

There are still some things I'm wondering. Like in the added unit test, a changelog is collected, even in the case in which the dependabot commit has no `Changelog` entry. Is this not unexpected behaviour?

See below for some questions.


extra/changelog-generator/changelog-generator
```diff
@@ -22,6 +22,19 @@ SHA_TO_TRACKER = {}
 JIRA_REGEX = r"(?:Jira:? *)?(?:https?://tracker.mender.io/browse/)?((?:INF|ARCHIVE|MEN|QA|MC)(?:-| )[0-9]+)"
 TRACKER_REGEX = r"\(?(?:Ref:? *)?%s\)?:? *" % (JIRA_REGEX)
 
+def aggregate_dependabot_changelogs(entry_list):
+    new_entry_list = []
+    aggregated_dependabot_changelogs = "Aggregated Dependabot Changelogs:\n"
+    for entry in entry_list:
+        if re.match("[Bb]umps? .* from .* to .*", entry):
```
oleorhagen, line 29: Question: Is this good enough as a matcher for dependabot commits?

Alternative is the `Signed-off-by` tag maybe?

extra/changelog-generator/test-changelog-generator
```diff
@@ -252,6 +252,16 @@ Changelog: commit'
 
 ################################################################################
 
+git commit --allow-empty -m 'Bump github.com/stretchr/testify from 1.5.1 to 1.6.0
```
oleorhagen, line 255: I was not expecting this to show up in the `Changelog` :/

pyproject.toml
```diff
@@ -9,6 +9,7 @@ include = '''
     | tests
     | extra
   )/.*\.py$
+  | extra/changelog-generator/changelog-generator
```
lluiscampos, line 12: I thought I had already included it, as there is the `extra` folder above. What I didn't realize is that the regex only matches `.py` files...

Anyway, thanks!

oleorhagen: I noticed :)

extra/changelog-generator/test-changelog-generator
```diff
@@ -399,6 +429,12 @@ _Released xx.xx.xxxx_
 
 #### test-changelog-generator.$$ (HEAD)
 
+* Aggregated Dependabot Changelogs:
+  Bump github.com/klauspost/pgzip from 1.2.3 to 1.2.4
+  Bumps [github.com/klauspost/pgzip](https://github.com/klauspost/pgzip) from 1.2.3 to 1.2.4.
+  - [Release notes](https://github.com/klauspost/pgzip/releases)
+  - [Commits](klauspost/pgzip@v1.2.3...v1.2.4)Bump golang from 1.14 to 1.15.0
+  Bumps golang from 1.14 to 1.15.0.
```
kacf, line 437: There is some issue with double entries and newlines on the last and second to last line there!

Also, this should be formatted slightly differently. Since this ends up in markdown, if you do not add bullet points, most of that content will render on one line, with little to separate the entries. I suggest using this as a target test case:

```
* Aggregated Dependabot Changelogs:
  * Bump github.com/klauspost/pgzip from 1.2.3 to 1.2.4
  * Bumps [github.com/klauspost/pgzip](https://github.com/klauspost/pgzip) from 1.2.3 to 1.2.4.
    - [Release notes](https://github.com/klauspost/pgzip/releases)
    - [Commits](klauspost/pgzip@v1.2.3...v1.2.4)
  * Bump golang from 1.14 to 1.15.0
```

extra/changelog-generator/changelog-generator
kacf: I think this is alright, it's pretty explicit about which kinds of messages it is processing.

extra/changelog-generator/changelog-generator
```diff
@@ -23,6 +23,19 @@ JIRA_REGEX = r"(?:Jira:? *)?(?:https?://tracker.mender.io/browse/)?((?:INF|ARCHI
...
+            aggregated_dependabot_changelogs += entry
+        else:
+            new_entry_list.append(entry)
+    if aggregated_dependabot_changelogs != "":
+        new_entry_list.append(aggregated_dependabot_changelogs)
```
kacf, line 35: A nice solution!

kacf (changes requested):

extra/changelog-generator/test-changelog-generator
kacf: Oh wait, now I understand why it's double, it's because it appears like that in the commit message. Maybe just drop the title line then, since it is repeated in the body.

kacf: So the output would be this:

```
* Aggregated Dependabot Changelogs:
  * Bumps [github.com/klauspost/pgzip](https://github.com/klauspost/pgzip) from 1.2.3 to 1.2.4.
    - [Release notes](https://github.com/klauspost/pgzip/releases)
    - [Commits](klauspost/pgzip@v1.2.3...v1.2.4)
  * Bumps golang from 1.14 to 1.15.0
```

extra/changelog-generator/test-changelog-generator
oleorhagen: And it doesn't. Sry

oleorhagen:
I'm sorry for the force-push, but history is cleaner like this, and the change is small

lluiscampos (approved):
Nice job. LGTM

PR #989, opened 2020-08-21
refactor: Add convenience functions for creating ApiClients

alfrunes:
Added new subclasses of the ApiClient for internal apis.
Ref [comment from #985](https://github.com/mendersoftware/integration/pull/985#discussion_r474005397)

### DIFF
--- a/testutils/api/tenantadm.py
+++ b/testutils/api/tenantadm.py
@@ -12,6 +12,8 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+from testutils.api.client import ApiClient as _ApiClient
+
 HOST = "mender-tenantadm:8080"
 URL_INTERNAL = "/api/internal/v1/tenantadm"
 URL_MGMT = "/api/management/v1/tenantadm"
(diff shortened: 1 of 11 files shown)


testutils/api/tenantadm.py
```diff
@@ -24,3 +26,13 @@
...
 def req_status(status):
     return {"status": status}
+
+
+class InternalApiClient(_ApiClient):
+    def __init__(self, host=HOST, scheme="http://"):
+        super().__init__(URL_INTERNAL, host, scheme)
+
+
+class ApiClient(_ApiClient):
```
lluiscampos, line 36: Nitpick: I think this should have an explicit Management (or Mgmt) in the class name.

Even though in `tenantadm` there is no Devices API, I think is best to be explicit. Otherwise I would be asking myself "why don't we then have an ApiClient also in all the other services?" (answer being that we would need both ManagementApiClient, DevicesApiClient bringing little value).

lluiscampos (approved):
Real nice. Thank you very much!

lluiscampos:
@alfrunes This became a bit outdated now, but I still think would be nice to merge it... Do you have energy to rework and submit?

alfrunes:
> @alfrunes This became a bit outdated now, but I still think would be nice to merge it... Do you have energy to rework and submit?

Actually, I took a quick peek at this PR yesterday. I'll address it tomorrow if I have time.

PR #990, opened 2020-08-27
Misc text tweak on tests/run.sh usage.

merlin-northern:
ChangeLog:none

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -16,6 +16,8 @@ usage() {
     echo "    --                               Seperates 'run.sh' arguments from pytest arguments"
     echo "    <pytest-args>                    Passes these arguments along to pytest"
     echo "    tests/<testfile.py>              Name the test-file to run"
+    echo "    -k TestNameToRun                 Name of the test class, method, or module to run"
+
     echo
     echo "Recognized Environment Variables:"
     echo


merlin-northern:
stopped pipelines.

tests/run.sh
```diff
@@ -15,7 +15,7 @@ usage() {
     echo "    --get-requirements               Download the external binary requirements into ./downloaded-tools and exit"
     echo "    --                               Seperates 'run.sh' arguments from pytest arguments"
     echo "    <pytest-args>                    Passes these arguments along to pytest"
-    echo "    tests/<testfile.py>              Name the test-file to run"
+    echo "    -k tests/<testfile.py>           Name the test-file to run"
```
lluiscampos, line 18: That is not correct. if you use `-k` you cannot specify a test file like `<dir>/<testfile>`, but only by method/class/module name. The module happens to be the file name but without py.

I don't think this is necessary, as it is part of the "pytest-args" above. But in any case:
```suggestion
    echo "    tests/<testfile.py>           Name the test-file to run"
    echo "    TestNameToRun            Name of the test class, method or module to run"
```

Something like that

lluiscampos (changes requested):

tests/run.sh
merlin-northern: right! but when you want to use `TestNameToRun` you have to add -k and it will not be added by run.sh (that was my initial intention) so it has to be:
```
-k TestNameToRun            Name of the test class, method or module to run
```
right?

lluiscampos: @merlin-northern Yes! That was what I intended to write :) Good review of a review comment.

oleorhagen: Doesn't it have to be:
` -- -k TestNameToRun` ?

kacf: Yes, I believe so (as experienced the hard way).

lluiscampos: @oleorhagen @kacf Yes, but the `--` is already documented, just couple of lines above.

In a way this was already correct, as "<pytest-args>" could include `-k ...`. I think the point of @merlin-northern was to make this explicit to save other to experience the pain :)

merlin-northern: right!

lluiscampos (approved):

PR #999, opened 2020-09-01
[changelog-generator] Search the dependabot message in all commit entry

lluiscampos:
Some times dependabot titles the commit with "Bump `<something>` in
`<somewhere>`", while only having the "from `<this>` to `<that>`" in the
commit body.


oleorhagen (approved):
Nice!

Figured it was some matching issue.

PR #997, opened 2020-08-31
upgrade MongoDB from version 3.6 to version 4.4 (latest stable)

kjaskiewiczz:

PR #1000, opened 2020-09-01
MongoDB upgrade and introduce additional deviceauth dependency

kjaskiewiczz:


kjaskiewiczz:
@lluiscampos I have one additional commit here which I'm using to test changes in the deviceauth, so it would be great if I could keep it at least till the end of the day.

lluiscampos:
> @lluiscampos I have one additional commit here which I'm using to test changes in the deviceauth, so it would be great if I could keep it at least till the end of the day.

@kjaskiewiczz I see. Just let me know explicitly if we need this commit also in the release (#997 is in the release)

kjaskiewiczz:
@lluiscampos we don't need it in the release

PR #1001, opened 2020-09-02
MEN-3905 test_mtls_enterprise_hsm

merlin-northern:
Use SoftHSM2 to emulate private key from hardware security module in the mTLS test.

* setup the SoftHSM
* initialize token
* store private key in the slot
* get the key PKCS#11 URI
* configure OpenSSL
* configure Mender to use the key via URI and SSLEngine configured
* configure the mender-client service to be aware of SoftHSM emulation
* run the test (only RSA keys are supported currently)
* test_mtls_enterprise_hsm is a copy of test_mtls_enterprise with use_hsm->True

ChangeLog:none

### DIFF
--- a/tests/tests/test_mtls.py
+++ b/tests/tests/test_mtls.py
@@ -24,7 +24,7 @@ from testutils.infra.container_manager import factory
 from testutils.infra.device import MenderDevice
 
 from .. import conftest
-from ..MenderAPI import reset_mender_api, auth, deploy, devauth
+from ..MenderAPI import reset_mender_api, auth, deploy, devauth, logger
 from .common_artifact import get_script_artifact
 from .mendertesting import MenderTesting
 
@@ -76,7 +76,61 @@ exit 0
 
 
 class TestClientMTLSEnterprise:
-    def common_test_mtls_enterprise(self, setup_ent_mtls, algorithm=None):
+    def hsm_setup(self, pin, ssl_engine_id, device):
+        algorithm = "rsa"
+        key = f"/var/lib/mender/client.1.{algorithm}.key"
+        script = f"""\
+#!/bin/bash
+echo "module: /usr/lib/softhsm/libsofthsm2.so" > /usr/share/p11-kit/modules/softhsm2.module
+mkdir -p /softhsm/tokens
+echo "directories.tokendir = /softhsm/tokens" > /softhsm/softhsm2.conf
+export SOFTHSM2_CONF=/softhsm/softhsm2.conf;
+softhsm2-util --init-token --free --label unittoken1 --pin {pin} --so-pin 0002
+pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin {pin} --write-object "{key}" --type privkey --id 0909 --label privatekey
+"""
+        tmpdir = tempfile.mkdtemp()
+        initialize_hsm_script = os.path.join(tmpdir, "init-hsm.sh")
+        try:
+            with open(initialize_hsm_script, "w") as fd:
+                fd.write(script)
+            device.put(
+                os.path.basename(initialize_hsm_script),
+                local_path=os.path.dirname(initialize_hsm_script),
+                remote_path="/tmp",
+            )
+            device.run("chmod 755 /tmp/" + os.path.basename(initialize_hsm_script))
+            device.run("/tmp/" + os.path.basename(initialize_hsm_script))
+        finally:
+            shutil.rmtree(tmpdir)
+
+    def hsm_get_key_uri(self, pin, ssl_engine_id, device):
+        key_uri = device.run(
+            "export SOFTHSM2_CONF=/softhsm/softhsm2.conf; p11tool --login --provider=/usr/lib/softhsm/libsofthsm2.so --set-pin="
+            + pin
+            + " --list-all-privkeys | grep URL | sed -e 's/.*URL: //'"
+        ).rstrip("\n")
+        key_uri = key_uri + ";pin-value=" + pin
+        device.run("cp /etc/ssl/openssl.cnf /etc/ssl/openssl.cnf.backup")
+        device.run(
+            'echo -ne "[openssl_init]\nengines=engine_section\n\n[engine_section]\npkcs11 = pkcs11_section\n\n[pkcs11_section]\nengine_id = '
+            + ssl_engine_id
+            + '\nMODULE_PATH = /usr/lib/softhsm/libsofthsm2.so\ninit = 0\n" >> /etc/ssl/openssl.cnf'
+        )
+        device.run(
+            'sed -i.backup -e "/\\[Service\\]/ a Environment=SOFTHSM2_CONF=/softhsm/softhsm2.conf" /lib/systemd/system/mender-client.service'
+        )
+        return key_uri
+
+    def hsm_cleanup(self, device):
+        device.run(
+            "mv /lib/systemd/system/mender-client.service.backup /lib/systemd/system/mender-client.service"
+        )
+        device.run("rm -Rf /softhsm")
+        device.run("mv /etc/ssl/openssl.cnf.backup /etc/ssl/openssl.cnf")
+
+    def common_test_mtls_enterprise(
+        self, setup_ent_mtls, algorithm=None, use_hsm=False
+    ):
         tenant = setup_ent_mtls.tenant
 
         # stop the api gateway
@@ -90,33 +144,6 @@ class TestClientMTLSEnterprise:
             setup_ent_mtls.get_mender_clients()[0]
         )
         device.ssh_is_opened()
-        device.run("systemctl stop mender-client")
-        tmpdir = tempfile.mkdtemp()
-        try:
-            # retrieve the original configuration file
-            output = device.run("cat /etc/mender/mender.conf")
-            config = json.loads(output)
-            # replace mender.conf with an mTLS enabled one
-            config["ServerURL"] = "https://mtls-ambassador:8080"
-            config["ServerCertificate"] = "/etc/mender/server.crt"
-            config["SkipVerify"] = True
-            if algorithm is not None:
-                config["HttpsClient"] = {
-                    "Certificate": f"/var/lib/mender/client.1.{algorithm}.crt",
-                    "Key": f"/var/lib/mender/client.1.{algorithm}.key",
-                }
-            if "ArtifactVerifyKey" in config:
-                del config["ArtifactVerifyKey"]
-            mender_conf = os.path.join(tmpdir, "mender.conf")
-            with open(mender_conf, "w") as fd:
-                json.dump(config, fd)
-            device.put(
-                os.path.basename(mender_conf),
-                local_path=os.path.dirname(mender_conf),
-                remote_path="/etc/mender",
-            )
-        finally:
-            shutil.rmtree(tmpdir)
 
         # upload the certificates
         basedir = os.path.join(os.path.dirname(__file__), "..", "..",)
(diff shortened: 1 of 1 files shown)


lluiscampos:
@merlin-northern This looks very promising, I think you are in the right track. Just ping me if you run into trouble.

merlin-northern:
the diff is much smaller in reality, I moved one line up, and it looks like I modified a lot -- which is not the case. this mainly added the initialization of the SoftHSM and OpenSSL and the mender config changes.

tests/tests/test_mtls.py
```diff
@@ -76,7 +76,9 @@ def make_script_artifact(artifact_name, device_type, output_path):
 
 
 class TestClientMTLSEnterprise:
-    def common_test_mtls_enterprise(self, setup_ent_mtls, algorithm=None):
+    def common_test_mtls_enterprise(
+        self, setup_ent_mtls, algorithm=None, use_hsm=False
+    ):
```
lluiscampos, line 81: The new added logic here is independent of what was before (IOW, we have `if ...:` with no `else`). What was your motivation for putting it here?

I suggest we leave this function as it was, and move all the hsm related part in a different helper function. If I haven't overlooked something, this should all be doable independently (before calling `common_test_mtls_enterprise`)

merlin-northern: > The new added logic here is independent of what was before (IOW, we have `if ...:` with no `else`). What was your motivation for putting it here?

1. the fact that once we get the key URI we put it in the config, and then start the client.
2. the minimal required changes. it only changed a couple of lines there. (the diff got confused a bit by me moving one line up.)

let me think how to address your comment. I will come up with something.

tests/tests/test_mtls.py
```diff
@@ -154,9 +129,84 @@ def common_test_mtls_enterprise(self, setup_ent_mtls, algorithm=None):
...
+            )
+        finally:
+            shutil.rmtree(tmpdir)
+
+        device.run(
+            'echo -ne "[openssl_init]\nengines=engine_section\n\n[engine_section]\npkcs11 = pkcs11_section\n\n[pkcs11_section]\nengine_id = pkcs11\nMODULE_PATH = /usr/lib/softhsm/libsofthsm2.so\ninit = 0\n" >> /etc/ssl/openssl.cnf'
+        )
+        device.run(
+            'sed -i -e "/\[Service\]/ a Environment=SOFTHSM2_CONF=/softhsm/softhsm2.conf" /lib/systemd/system/mender-client.service'
+        )
```
kacf, line 204: I think we should have some cleanup of these changes. Currently we make a new set of containers for every test, but we may not always stick to that, especially when we consider the staging environment. Better to reduce that technical debt now than later.

So at least I would make a backup of the `/lib/systemd/system/mender-client.service` and `/etc/ssl/openssl.cnf` and restore them in a cleanup function. Similarly, I would also delete `/softhsm` on the device.

kacf (changes requested):
Impressive test! The testing itself looks good, but I think we should have some cleanup, see below.

tests/tests/test_mtls.py
merlin-northern: got it! will do!

tests/tests/test_mtls.py
```diff
@@ -191,6 +259,40 @@ def test_mtls_enterprise(self, setup_ent_mtls, algorithm):
...
+        deploy.check_expected_statistics(deployment_id, "success", 1)
+        deploy.check_expected_status("finished", deployment_id)
+
+        # verify the update was actually installed on the device
+        out = device.run(
+            "export SOFTHSM2_CONF=/softhsm/softhsm2.conf; mender -show-artifact"
+        ).strip()
+        assert out == "mtls-artifact"
+
+        self.hsm_cleanup(device)
```
lluiscampos, line 294: This one should be called in a `try/finally` fashion, as the test might abort before reaching here.

tests/tests/test_mtls.py
```diff
@@ -191,6 +259,40 @@ def test_mtls_enterprise(self, setup_ent_mtls, algorithm):
...
+            "mtls-test", artifact_name="mtls-artifact", devices=devices,
+        )
+
+        # now just wait for the update to succeed
+        deploy.check_expected_statistics(deployment_id, "success", 1)
+        deploy.check_expected_status("finished", deployment_id)
+
+        # verify the update was actually installed on the device
+        out = device.run(
+            "export SOFTHSM2_CONF=/softhsm/softhsm2.conf; mender -show-artifact"
```
lluiscampos, line 290: Why do we need `export SOFTHSM2_CONF=/softhsm/softhsm2.conf` here?

tests/tests/test_mtls.py
merlin-northern: @lluiscampos ok, this is the best I could do, I separated most of the hsm logic into separate function. I hope it is ok!

tests/tests/test_mtls.py
merlin-northern: Kristian: I added the cleanup call.

tests/tests/test_mtls.py
```diff
@@ -76,7 +76,59 @@ def make_script_artifact(artifact_name, device_type, output_path):
 
 
 class TestClientMTLSEnterprise:
-    def common_test_mtls_enterprise(self, setup_ent_mtls, algorithm=None):
+    def get_key_uri_hsm(self, ssl_engine_id, device):
```
lluiscampos, line 79: I would separate this in two methods: `hsm_setup` (analog to `hsm_cleanup`) and `hsm_get_key_uri`

tests/tests/test_mtls.py
lluiscampos: I wanted to start the conversation, no need to force-push so quickly.

> 1. the fact that once we get the key URI we put it in the config, and then start the client.

I thought that it is not a big deal to write into the config file twice.

> 2. the minimal required changes. it only changed a couple of lines there. (the diff got confused a bit by me moving one line up.)

IMHO this is not a priority: we "pay" adding more complexity to the method.

Being said the above, it is okey leaving it as it is, this is a low priority conversation.

tests/tests/test_mtls.py
```diff
@@ -76,7 +76,59 @@ def make_script_artifact(artifact_name, device_type, output_path):
...
+        algorithm = "rsa"
+        tmpdir = tempfile.mkdtemp()
+        key = f"/var/lib/mender/client.1.{algorithm}.key"
+        pin = "0001"
+        script = f"""\
+#!/bin/bash
+echo "module: /usr/lib/softhsm/libsofthsm2.so" > /usr/share/p11-kit/modules/softhsm2.module
+mkdir -p /softhsm/tokens
+echo "directories.tokendir = /softhsm/tokens" > /softhsm/softhsm2.conf
```
lluiscampos, line 88: Ideally, this should have been part of the `meta-mender` PR, not the test.

lluiscampos: (but is okey! I only mention it for future reference)

tests/tests/test_mtls.py
```diff
@@ -76,7 +76,59 @@ def make_script_artifact(artifact_name, device_type, output_path):
...
+        with open(initialize_hsm_script, "w") as fd:
+            fd.write(script)
+        device.put(
+            os.path.basename(initialize_hsm_script),
+            local_path=os.path.dirname(initialize_hsm_script),
+            remote_path="/tmp",
+        )
+        device.run("chmod 755 /tmp/" + os.path.basename(initialize_hsm_script))
+        device.run("/tmp/" + os.path.basename(initialize_hsm_script))
+        shutil.rmtree(tmpdir)
```
lluiscampos, line 103: This should also be in a `try/finally` block, to ensure that we remove the tmpdir

tests/tests/test_mtls.py
```diff
@@ -76,7 +76,59 @@ def make_script_artifact(artifact_name, device_type, output_path):
...
+        ).rstrip("\n")
+        key_uri = key_uri + ";pin-value=" + pin
+        device.run("cp /etc/ssl/openssl.cnf /etc/ssl/openssl.cnf.backup")
+        device.run(
+            'echo -ne "[openssl_init]\nengines=engine_section\n\n[engine_section]\npkcs11 = pkcs11_section\n\n[pkcs11_section]\nengine_id = '
+            + ssl_engine_id
+            + '\nMODULE_PATH = /usr/lib/softhsm/libsofthsm2.so\ninit = 0\n" >> /etc/ssl/openssl.cnf'
+        )
+        device.run(
+            'sed -i.backup -e "/\[Service\]/ a Environment=SOFTHSM2_CONF=/softhsm/softhsm2.conf" /lib/systemd/system/mender-client.service'
```
lluiscampos, line 118: Thank you so much for this! I didn't know about the `a` command in sed, and this is something I have been looking for for long...

tests/tests/test_mtls.py
```diff
@@ -191,6 +259,40 @@ def test_mtls_enterprise(self, setup_ent_mtls, algorithm):
...
+        deploy.check_expected_status("finished", deployment_id)
+
+        # verify the update was actually installed on the device
+        out = device.run(
+            "export SOFTHSM2_CONF=/softhsm/softhsm2.conf; mender -show-artifact"
+        ).strip()
+        assert out == "mtls-artifact"
+
+        self.hsm_cleanup(device)
+
```
lluiscampos, line 295: Is it enough with checking that the Artifact was installed? Probably yes, but I was expecting to check something from the Mender client logs that would ensure we are loading the keys actually from the HSM.

lluiscampos (changes requested):
Looks promising!

See comments and questions below.

tests/tests/test_mtls.py
merlin-northern: this is one of the follow ups I mentioned, once mender is called it will print errors about the keys/certificates. I thought they were coming from the daemon, but it was this place. when using SoftHSM2 we have to have the `SOFTHSM2_CONF` in the env, otherwise it is impossible for the engine to locate the keys. the follow-up would be: what to do in case we have mtls enabled and we are calling -show-artifact, should we print the errors or even load the keys? (low priority I think for now)

tests/tests/test_mtls.py
merlin-northern: you are reading my mind, thats the second followup I mentioned during su. I will take care of it.

lluiscampos: Nice. So the idea could be:

```
output = device.run("journalctl -u %s" % device.get_client_service_name())
assert "something" in output
```

tests/tests/test_mtls.py
merlin-northern: done.

tests/tests/test_mtls.py
merlin-northern: done.

tests/tests/test_mtls.py
merlin-northern: I do not think so -- I thought about it, and this is specific configuration for the package, it may some in handy but also may be disruptive if someone would like to do different tests. /softhsm/softhsm2.conf is specific to this test.

tests/tests/test_mtls.py
merlin-northern: done.

tests/tests/test_mtls.py
merlin-northern: that makes two of us :) I always look for it every time I want to use it

tests/tests/test_mtls.py
merlin-northern: done. now this I mentioned during su as a follow up. the story: I had put in for this very purpose a like of logs in the mender PR from previous week, but there is a misspell there, it says:
```
time="2020-09-04T09:51:32Z" level=info msg="laoded private key: '&{0xc43560 0xc000010a30}' from 'pkcs11'."
```
so I will file a PR and then correct also here. right now I am checking for "ded private key: " to be present in the logs.

and the test will not pass if the key cannot be loaded -- thats the whole idea of mtls test and thats why i chose to add HSM private key loading testing in here.

merlin-northern:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/185910404

tests/tests/test_mtls.py
```diff
@@ -76,7 +76,61 @@ def make_script_artifact(artifact_name, device_type, output_path):
...
+        key_uri = key_uri + ";pin-value=" + pin
+        device.run("cp /etc/ssl/openssl.cnf /etc/ssl/openssl.cnf.backup")
+        device.run(
+            'echo -ne "[openssl_init]\nengines=engine_section\n\n[engine_section]\npkcs11 = pkcs11_section\n\n[pkcs11_section]\nengine_id = '
+            + ssl_engine_id
+            + '\nMODULE_PATH = /usr/lib/softhsm/libsofthsm2.so\ninit = 0\n" >> /etc/ssl/openssl.cnf'
+        )
+        device.run(
+            'sed -i.backup -e "/\\[Service\\]/ a Environment=SOFTHSM2_CONF=/softhsm/softhsm2.conf" /lib/systemd/system/mender-client.service'
+        )
```
lluiscampos, line 121: This could have been in `hsm_setup`

tests/tests/test_mtls.py
lluiscampos: Got it. Thanks!

tests/tests/test_mtls.py
lluiscampos: Thank you, I understand now. This is a good question, indeed we can discuss in a client team meeting if you want. But I don't think it is a big deal, this is quite an advance use case and we should assume the user knows enough of OpenSSL.

lluiscampos (approved):
Looks good to me now. Thank you for the rework.

Overall great work with this issue, @merlin-northern. After having done an integration test with changes in meta-mender, I can say you are a fully empowered client developer now :)

lluiscampos:
Tests are still running. But I think we can merge these two PRs tomorrow morning. I can take care of it. Thanks again!

PR #1013, opened 2020-09-22
Fix race condition in 'test_compat'

oleorhagen:
Changelog: None

### DIFF
--- a/tests/tests/test_compat.py
+++ b/tests/tests/test_compat.py
@@ -131,7 +131,7 @@ def assert_inventory_updated(api_inventory, num_devices, timeout=TIMEOUT):
             pytest.fail("timeout waiting for devices to submit inventory")
 
         rsp = api_inventory.call(
-            "GET", inventory.URL_DEVICES, qs_params={"per_page": num_devices * 2},
+            "GET", inventory.URL_DEVICES, qs_params={"per_page": num_devices * 2}
         )
         assert rsp.status_code == 200
         dev_invs = rsp.json()
@@ -177,7 +177,7 @@ def assert_successful_deployment(api_deployments, deployment_id, timeout=TIMEOUT
         dpl = rsp.json()
         if dpl["status"] == "finished":
             rsp = api_deployments.call(
-                "GET", deployments.URL_DEPLOYMENTS_STATISTICS.format(id=deployment_id),
+                "GET", deployments.URL_DEPLOYMENTS_STATISTICS.format(id=deployment_id)
             )
             assert rsp.status_code == 200
             assert rsp.json()["failure"] == 0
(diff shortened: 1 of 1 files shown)


tests/tests/test_compat.py
```diff
@@ -189,40 +189,68 @@ def assert_successful_deployment(api_deployments, deployment_id, timeout=TIMEOUT
...
+                RebootDetectorThread(host_ip, client, timeout, self.enter_barrier)
+            )
+
+    def __enter__(self):
+        for d in self.reboot_detectors:
+            d.start()
+
+        # Wait for all reboot-detectors to finish their setup, before starting
+        # the code within the 'with' statement
+        self.enter_barrier.wait()
```
alfrunes, line 247: Is it 100% certain that the parallel threads can't raise an exception? Can you add a short timeout here (~10sec) just to be safe? And also abort the test `if self.enter_barrier.broken` after.

kacf (approved):
Very nice solution!

alfrunes (changes requested):
Well done!
Maybe we should consider replacing (or merging the parallel implementation into) the reboot detector in testutils as well?
I just have one small change request.

tests/tests/test_compat.py
oleorhagen: As far as I understand, all exceptions are caught and put into the exception queue? And can the barrier be broken if no one has called `abort`?

oleorhagen:
> Well done!
> Maybe we should consider replacing (or merging the parallel implementation into) the reboot detector in testutils as well?
> I just have one small change request.

Yes I was playing with the idea. Need some more work then though, as this one only supports one reboot. How about we merge this, and I promise I will follow up!

tests/tests/test_compat.py
alfrunes: I believe the barrier "breaks" if you time out waiting for it.

tests/tests/test_compat.py
```diff
@@ -189,40 +189,68 @@ def assert_successful_deployment(api_deployments, deployment_id, timeout=TIMEOUT
...
+        self._timeout = timeout
+        self._exception_q = queue.Queue(maxsize=1)
+        self.detector = self._client.get_reboot_detector(self._host_ip)
+        self.barrier = barrier
+
+    def run(self):
+        try:
+            # Setup the individual reboot-detectors
+            with self.detector as rd:
+                # Possibly move this to after the reboot-detector accepts connections (?)
```
alfrunes, line 206: ... I say yes, and then time out if the client does not respond (ref. review comment).

tests/tests/test_compat.py
oleorhagen: > This method may raise a BrokenBarrierError exception if the barrier is broken or reset while a thread is waiting.

Is this not good enough? I mean, the test will fail with a broken barrier exception

tests/tests/test_compat.py
oleorhagen: I think I'm going to leave it as it is actually. This is easier, and cleaner. It fixes the recurring issue we have been seeing. The comment is a little misleading now, as it was made before some of the changes.

alfrunes (approved):

PR #1016, opened 2020-09-23
MEN-3076: update test_artifact to the latest mender-artifact

tranchitella:
Changelog: none


kacf (changes requested):
This should also be targeted towards the `feature-partial_updates` branch.

kacf (approved):

PR #1017, opened 2020-09-25
backend-tests: auditlogs for useradm

mchalski:
part 2 of:
https://tracker.mender.io/browse/MEN-3815

test log emission from useradm events.

needs the last bunch of fixes still:
https://github.com/mendersoftware/useradm-enterprise/pull/138

combined pipeline:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/194456139

### DIFF
--- a/backend-tests/tests/test_auditlogs.py
+++ b/backend-tests/tests/test_auditlogs.py
@@ -15,6 +15,7 @@
 import pytest
 import uuid
 import os
+import json
 import time
 from datetime import datetime
 import urllib.parse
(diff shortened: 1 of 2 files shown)


backend-tests/tests/test_auditlogs.py
```diff
@@ -95,29 +87,77 @@ def test_deployment_create(self, tenant_users):
...
+        user_change = tenant_users.users[1]
+
+        alogs = ApiClient(auditlogs.URL_MGMT)
+
+        roles = ["RBAC_ROLE_CI"]
+        change_role(user.token, user_change.id, roles)
+        expected = evt_change_role(user, user_change, roles)
+
+        time.sleep(0.5)
+        res = alogs.with_auth(user.token).call("GET", auditlogs.URL_LOGS)
```
alfrunes, line 134: Nitpick; just to be safe, maybe we should poll wait on this one? The response time may vary since the POST needs to bounce of workflows as well.
```suggestion
        for _ in range(20):
            time.sleep(0.5)
            res = alogs.with_auth(user.token).call("GET", auditlogs.URL_LOGS)
            if res.status_code == 200:
                break
```

backend-tests/tests/test_auditlogs.py
```diff
@@ -95,29 +87,77 @@ def test_deployment_create(self, tenant_users):
...
+        user = tenant_users.users[0]
+        user_del = tenant_users.users[1]
+
+        alogs = ApiClient(auditlogs.URL_MGMT)
+
+        delete_user(user, user_del.id)
+        expected = evt_user_delete(user, user_del)
+
+        time.sleep(0.5)
+        res = alogs.with_auth(user.token).call("GET", auditlogs.URL_LOGS)
```
alfrunes, line 116: Same here.
```suggestion
        for _ in range(20):
            time.sleep(0.5)
            res = alogs.with_auth(user.token).call("GET", auditlogs.URL_LOGS)
            if res.status_code == 200:
                break
```

backend-tests/tests/test_auditlogs.py
```diff
@@ -95,29 +87,77 @@ def test_deployment_create(self, tenant_users):
...
+    def test_user_create(self, tenant_users):
+        user = tenant_users.users[0]
+
+        alogs = ApiClient(auditlogs.URL_MGMT)
+
+        uid = make_user(user.token, "foo@acme.com", "secretsecret")
+        expected = evt_user_create(user, uid, "foo@acme.com")
+
+        time.sleep(0.5)
+        res = alogs.with_auth(user.token).call("GET", auditlogs.URL_LOGS)
```
alfrunes, line 99: ... And here
```suggestion
        for _ in range(20):
            time.sleep(0.5)
            res = alogs.with_auth(user.token).call("GET", auditlogs.URL_LOGS)
            if res.status_code == 200:
                break
```

alfrunes (approved):
Looks good to me.
I only have a couple of less important suggestions (take it or leave it :)

backend-tests/tests/test_auditlogs.py
mchalski: very good suggestion, fixed all the related issues.

actually, I bit the bullet and pulled in a retry library - some cool utils for retries, which we all need in multiple places.

I chose mozilla/redo for the time being, liked it the most. it can handle both function retries(decorators) and loose code blocks (wrap them in a retrier generator loop, like I do).

mchalski:
After @alfrunes suggestions, pipeline no. 2:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/194594294

PR #1028, opened 2020-10-01
[test_mtls] Fix hard-coded service name

lluiscampos:
So that the test can run on warrior or older yocto branches, where the
systemd service is "mender".

### DIFF
--- a/tests/tests/test_mtls.py
+++ b/tests/tests/test_mtls.py
@@ -126,13 +126,15 @@ pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin {pin} --write
             + '\nMODULE_PATH = /usr/lib/softhsm/libsofthsm2.so\ninit = 0\n" >> /etc/ssl/openssl.cnf'
         )
         device.run(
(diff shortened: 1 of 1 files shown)


tests/tests/test_mtls.py
```diff
@@ -126,13 +126,15 @@ def hsm_get_key_uri(self, pin, ssl_engine_id, device):
...
+            'sed -i.backup -e "/\\[Service\\]/ a Environment=SOFTHSM2_CONF=/softhsm/softhsm2.conf" /lib/systemd/system/%s.service'
+            % client_service_name
         )
         return key_uri
 
     def hsm_cleanup(self, device):
         device.run(
-            "mv /lib/systemd/system/mender-client.service.backup /lib/systemd/system/mender-client.service"
+            "mv /lib/systemd/system/%s.service.backup /lib/systemd/system/%s.service"
+            % (client_service_name, client_service_name)
```
kacf, line 137: Is the variable visible in these two locations?

kacf (approved):
The fix looks good, but there may be a typo, see below.

tests/tests/test_mtls.py
lluiscampos: No, the test failed. Too quick search+replace late at night, sorry. I'll fix ASAP

PR #1030, opened 2020-10-02
MEN-3075: Integration test for `clears_artifact_provides`. 

kacf:
We just need to modify and add some checks for the existing inventory
test to use the `clears_artifact_provides` defaults.

Note that deleting attributes does not work yet due to MEN-4001 and
MEN-4003, so the test is marked XFAIL at the moment. The rest appears
to work correctly.

Changelog: None

### DIFF
--- a/tests/tests/test_inventory.py
+++ b/tests/tests/test_inventory.py
@@ -38,32 +38,43 @@ exit 0
 @pytest.mark.usefixtures("standard_setup_one_client_bootstrapped")
 class TestInventory(MenderTesting):
     @MenderTesting.fast
+    @pytest.mark.xfail(reason="MEN-4001 and MEN-4003.", strict=True)
     def test_inventory(self):
         """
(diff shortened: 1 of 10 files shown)


tranchitella:
Depends on https://github.com/mendersoftware/inventory/pull/214

tests/tests/test_inventory.py
```diff
@@ -38,32 +38,43 @@ def make_script_artifact(artifact_name, device_type, output_path, extra_args):
 @pytest.mark.usefixtures("standard_setup_one_client_bootstrapped")
 class TestInventory(MenderTesting):
     @MenderTesting.fast
+    @pytest.mark.xfail(reason="MEN-4001 and MEN-4003.", strict=True)
```
tranchitella, line 41: I suppose we can get rid of the xfail, now

tranchitella (approved):
Looks good

PR #1034, opened 2020-10-05
MEN-3980: Add test for the rollback client error

oleorhagen:
Changelog: None

### DIFF
--- a/tests/tests/test_state_scripts.py
+++ b/tests/tests/test_state_scripts.py
@@ -438,7 +438,7 @@ REBOOT_TEST_SET = [
     (
         "simulate_powerloss_artifact_install_enter",
         {
-            "RebootScripts": ["ArtifactInstall_Enter_01",],
+            "RebootScripts": ["ArtifactInstall_Enter_01"],
             "ExpectedFinalPartition": ["OriginalPartition"],
             "ScriptOrder": [
                 "ArtifactInstall_Enter_01",
--- a/tests/tests/test_state_scripts.py
+++ b/tests/tests/test_state_scripts.py
@@ -394,6 +394,43 @@ TEST_SETS = [
             ],
         },
     ),
+    (
+        "No_roll_forward_on_ArtifactInstall_Leave_error",
+        {
+            "FailureScript": ["ArtifactInstall_Leave_03"],
+            "ExpectedStatus": "failure",
+            "SwapPartitionExpectation": False,  # We should not switch partition
+            "ScriptOrder": [
+                "Idle_Enter_08_testing",
+                "Idle_Enter_09",
+                "Idle_Leave_09",
+                "Idle_Leave_10",
+                "Sync_Enter_02",
+                "Sync_Enter_03",
+                "Sync_Leave_04",
+                "Sync_Leave_15",
+                "Download_Enter_12",
+                "Download_Enter_13",
+                "Download_Leave_14",
+                "Download_Leave_25",
+                "ArtifactInstall_Enter_01",
+                "ArtifactInstall_Enter_02",
+                "ArtifactInstall_Leave_01",
+                "ArtifactInstall_Leave_03",
+                "ArtifactInstall_Error_01",
+                "ArtifactInstall_Error_02",
+                "ArtifactInstall_Error_99",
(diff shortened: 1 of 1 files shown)


oleorhagen:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/198735414

tests/tests/test_state_scripts.py
```diff
@@ -873,6 +873,147 @@ def fetch_info(cmd_list):
...
+                "ArtifactInstall_Leave_01",
+                "ArtifactInstall_Leave_03",
+                "ArtifactRollback_Enter_01",
+                "ArtifactRollback_Enter_11",
+                "ArtifactRollback_Leave_01",
+                "ArtifactRollback_Leave_89",
+                "ArtifactFailure_Enter_01",
+                "ArtifactFailure_Leave_06",
+                "Idle_Enter_08_testing",  # Update attempt finished
+                "Idle_Enter_09",
```
kacf, line 904: Great!

But do you need a new test for this? Can't you just add this as one more case to the existing test?

tests/tests/test_state_scripts.py
```diff
@@ -873,6 +873,147 @@ def fetch_info(cmd_list):
                 % (client_service_name, client_service_name)
             )
 
+    TEST_SET_ROLLBACK = (
+        "Normal_success",
```
kacf, line 877: "Failure in ArtifactInstall_Leave"

tests/tests/test_state_scripts.py
```diff
@@ -873,6 +873,147 @@ def fetch_info(cmd_list):
...
+                    fd.write(script_failure_content)
+                else:
+                    fd.write(script_content)
+
+        # Create the artifact, and make the deployment.
+        device_id = Helpers.ip_to_device_id_map(
+            MenderDeviceGroup([mender_device.host_string])
+        )[mender_device.host_string]
+
+        host_ip = standard_setup_one_client_bootstrapped.get_virtual_network_host_ip()
```
lluiscampos, line 957: Unused

tests/tests/test_state_scripts.py
lluiscampos: I wonder the same, but it seems like the logic is a bit different.

In any case, for consistency I suggest we move `TEST_SET_ROLLBACK` to before the class definition, just after `REBOOT_TEST_SET`, or (better?) we move`TEST_SETS` and `REBOOT_TEST_SET` to just above the test methods that use them.

tests/tests/test_state_scripts.py
```diff
@@ -873,6 +873,147 @@ def fetch_info(cmd_list):
...
+            script_logs = ""
+            timeout = time.time() + 60 * 60
+            while timeout >= time.time():
+                time.sleep(3)
+                try:
+                    script_logs = mender_device.run("cat /data/test_state_scripts.log")
+                    if test_set.get("ExpectedScriptFlow")[-1] in script_logs:
+                        break
+                except Exception as e:
+                    pytest.fail("Received exception: %s".format(e))
```
lluiscampos, line 981:
```suggestion
                    pytest.fail("Received exception: %s".format(str(e)))
```

tests/tests/test_state_scripts.py
```diff
@@ -873,6 +873,147 @@ def fetch_info(cmd_list):
...
+            with open(os.path.join(artifact_script_dir, script), "w") as fd:
+                if script in test_set.get("FailureScript", []):
+                    fd.write(script_failure_content)
+                else:
+                    fd.write(script_content)
+
+        # Create the artifact, and make the deployment.
+        device_id = Helpers.ip_to_device_id_map(
+            MenderDeviceGroup([mender_device.host_string])
+        )[mender_device.host_string]
```
lluiscampos, line 955: I wonder why we need this. It is in all the three tests and probably just a leftover since before the Python 3 refactoring, but I don't think it is necessary.

Passing `devices=None` (default) to `common_update_procedure` should have the same effect, right? Do you dare to try? :)

lluiscampos (commented):
Some comments and suggestions below.

oleorhagen:
Moved to acceptance test instead

PR #1036, opened 2020-10-06
tenantadm mgmt /secret was moved to /billing/secret

merlin-northern:
ChangeLog:none


tranchitella (approved):

mzedel (approved):
and here I thought the gui would be the only consumer of the endpoint...

thanks for the fix!

PR #1037, opened 2020-10-07
Minor improvements HSM test

lluiscampos:
* [test_mtls] Separate p11tool call from output process
So that if there is an error invoking p11tool is caught by the test
framework.

* [test_mtls] Let the cleanup complete if there are no backup files


merlin-northern (approved):
lgtm

PR #1039, opened 2020-10-08
MEN-4029 Backend integration tests: deploy to group.

merlin-northern:
ChangeLog:none


merlin-northern:
pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/199667711

backend-tests/tests/test_deployments.py
```diff
@@ -552,7 +552,7 @@ def compare_phases_json(expected, response):
                 assert exp[k] == rsp[k]
 
 
-def setup_devices_and_management_st(nr_devices=100):
+def setup_devices_and_management_st(nr_devices=100, deploy_to_group=""):
```
oleorhagen, line 555: Suggestion:
I'm not a pythonista, but I do think it is common to:
```suggestion
def setup_devices_and_management_st(nr_devices=100, deploy_to_group=None):
```

And then the test can be:
```
if deploy_to_group:
    ...
```

backend-tests/tests/test_deployments.py
```diff
@@ -571,6 +571,14 @@ def setup_devices_and_management_st(nr_devices=100):
     devs = make_accepted_devices(devauthd, devauthm, utoken, "", nr_devices)
     # wait for devices to be provisioned
     time.sleep(3)
+    if len(deploy_to_group) > 0:
+        for device in devs[:-1]:
```
oleorhagen, line 575: Question:
I'm not into this, but why strip off one device?

oleorhagen (approved):
Just added some suggestions :)

backend-tests/tests/test_deployments.py
merlin-northern: thats part of the test logic. one of the tests check that last device was not part of the deployment.

oleorhagen: Makes sense :100:

tranchitella (approved):

backend-tests/tests/test_deployments.py
merlin-northern: sure! looks prettier:) I just pushed that and since that is cosmetic I will not re-run the pipelines.

PR #1043, opened 2020-10-13
MEN-4041: Adjust auditlog test to an updated API response model

alfrunes:


mzedel (approved):
~not quite sure what's behind the added imports, but the rest~ should align with the useradm PR 😄

PR #1057, opened 2020-10-28
test: upgrade the test_forced_inventory to wait for longer

oleorhagen:
The previous iteration slept for 5 seconds and only gave one shot at verifying
the inventory was properly updated.

This version tries 10 times with a 10 second period to see if the inventory was
indeed updated.

This should be ample in the case of slow test runners.

Changelog: None

### DIFF
--- a/tests/tests/test_basic_integration.py
+++ b/tests/tests/test_basic_integration.py
@@ -232,12 +232,12 @@ class TestBasicIntegration(MenderTesting):
         mender_device.run("mender -send-inventory")
         logger.info("mender client has forced an inventory update")
 
-        # Give the client some time to send the inventory.
-        time.sleep(5)
+        for i in range(10):
+            # Check that the updated inventory value is now present.
+            invJSON = inv.get_devices()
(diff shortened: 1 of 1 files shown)


merlin-northern (approved):
lgtm.
we can also check for the log message that the inventory update was sucessul.

oleorhagen:
> lgtm.
> we can also check for the log message that the inventory update was sucessul.

Yes, but in addition to this check you mean? Is that necessary?
Or does it add any improvements over this approach I'm not aware of?

kacf (approved):

kacf:
It's the test you are fixing which is failing.

@kjaskiewiczz: Do you think this will be fixed if the workflows job ordering is addressed?

oleorhagen:
> It's the test you are fixing which is failing.
> 
> @kjaskiewiczz: Do you think this will be fixed if the workflows job ordering is addressed?

Yes, but this was failing due to the service being down. There is no logic change in the code. I will re-run though, since the service seems to be back up.

PR #1066, opened 2020-11-03
QA-226: Add --feature-branches CLI option

lluiscampos:
* [release_tool] Add connect yaml file to the list of Docker-only
This is a new file added in feature-deviceconnect file for which
release_tool (master) is not aware of, hence failing with "More than one
container is using the image name 'deviceconnect'."

* [release_tool] QA-226: Add --feature-branches CLI option
Which in combination with --integration-versions-including option will
add the upstream remote branches to the candidates to check.

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -2609,6 +2609,8 @@ def do_integration_versions_including(args):
     ]
     if args.all:
         git_query += ["refs/heads/**"]
+    if args.feature_branches:
+        git_query += ["refs/remotes/%s/feature-*" % remote]
     output = execute_git(None, git_dir, git_query, capture=True)
     candidates = []
     for line in output.strip().split("\n"):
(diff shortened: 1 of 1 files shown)


extra/release_tool.py
```diff
@@ -3073,6 +3075,12 @@ def main():
         help="Find version(s) of integration repository that contain the given version of SERVICE, "
         + " where version is given with --version. Returned as a newline separated list",
     )
+    parser.add_argument(
+        "--feature-branches",
+        action="store_true",
+        default=False,
+        help="When used with -f, include upstream feature branches",
```
merlin-northern, line 3082: suggested rewording: "-f" -> "-f or --integration-versions-including"

merlin-northern (approved):
small nitpick. looks good to me! (ha, so I was so close to it) (missed the missing docker file)

extra/release_tool.py
lluiscampos: I was trying to keep in in sync with https://github.com/mendersoftware/integration/blob/981ea7f696c245358c6d824659568d0855f4034c/extra/release_tool.py#L3126

I would then need to modify both, and the other one would get too long IMO.

kacf (approved):

PR #1068, opened 2020-11-04
release_tool: Print HTTP body when Gitlab build fails.

kacf:
Changelog: None


kacf:
Cancelling pipeline, this does not affect it.

lluiscampos (approved):

PR #1070, opened 2020-11-10
MEN-4084: add mender-shell to integration yaml files

lluiscampos:


lluiscampos:
Requires https://github.com/mendersoftware/mender-qa/pull/448

lluiscampos:
Testing together with https://github.com/mendersoftware/mender-qa/pull/448 @ [Pipeline 214237931](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/214237931)

kacf (changes requested):
* I think you also need to add it [here](https://github.com/mendersoftware/integration/blob/master/extra/release_tool.py#L226)
* Shouldn't this be targeting a feature branch?

lluiscampos:
>     * I think you also need to add it [here](https://github.com/mendersoftware/integration/blob/master/extra/release_tool.py#L226)
> 
Correct, thanks

>     * Shouldn't this be targeting a feature branch?

(Related to a problem we are discussing in Slack with Fabio)

(Related to https://github.com/mendersoftware/integration/pull/1071)

I don't think so. We need to make the tool aware of the new repo in master, as integartion-test-bot and other pieces of CI/CD will always use integration/master to list repos or other tasks.

So what I suggest is we merge this to master, then we branch out to feature branch.

Does it make sense?

kacf:
Yes, I would have suggested setting `release_component: false` in this case (to avoid unintentionally releasing it), but I think this won't work either, because then it won't be listed by the release tool. So yeah, let's submit it to master, it's a small task to adjust it if we ever need to.

lluiscampos:
>     * I think you also need to add it [here](https://github.com/mendersoftware/integration/blob/master/extra/release_tool.py#L226)

Force-pushed fixing this and rebasing on top of latest master.

@kacf please have a second look, I don't think we need a new pipeline run

kacf (approved):

PR #1079, opened 2020-11-17
Update feature branch for deviceconnect

tranchitella:


tranchitella:
Cancelling the QA pipeline, as it runs with master branches.

lluiscampos (changes requested):
Looks good.

Additionally, I think the client needs to be set to `feature-client_auth_api`.

tranchitella:
@lluiscampos right, thanks!

alfrunes (approved):

PR #1084, opened 2020-11-18
Set HAVE_DEVICECONNECT for mender-api-gateway when deviceconnect is on

tranchitella:
Changelog: none


merlin-northern (approved):
lgtm

PR #1086, opened 2020-11-20
add deviceconnect to git-versions

kjaskiewiczz:


git-versions.yml
```diff
@@ -18,3 +18,6 @@ services:
 
     mender-create-artifact-worker:
         image: mendersoftware/create-artifact-worker:staging
+
+    mender-deviceconnect:
+        image: mendersoftware/deviceconnect:staging
```
lluiscampos, line 23: Could we put it before `mender-create-artifact-worker`? I think that will help to future conflicts resolution.

Or, even better, what if we get all the removed components (the OS ones) from master's `git-versions.yml` and just comment them out?

lluiscampos (approved):
Looks good. See nit-pick below.

lluiscampos (approved):
Thanks.

PR #1087, opened 2020-11-20
[release_tool] get_components_of_type: New param for independent comp

lluiscampos:
So that we don't need to hard-code them in the list for hosted release
procedure; which has proven to be difficult to maintain.


lluiscampos:
@alfrunes This should address the problem you had in last HM release.

alfrunes (approved):

kacf (approved):

PR #1088, opened 2020-11-20
backend test: test device status propagation

kjaskiewiczz:


backend-tests/run
```diff
@@ -15,6 +15,7 @@ COMPOSE_CMD_OPEN="docker-compose -p backend-tests \
         -f $INTEGRATION_PATH/docker-compose.demo.yml \
         -f $INTEGRATION_PATH/docker-compose.storage.minio.yml
         -f $INTEGRATION_PATH/backend-tests/docker/docker-compose.backend-tests.yml
+        -f $INTEGRATION_PATH/docker-compose.kj.yml \
```
tranchitella, line 18: Left-over here, I believe

tranchitella (changes requested):

backend-tests/run
kjaskiewiczz: yup

kjaskiewiczz:
the test only calls the command, but TBH, there not much more can be done here;
it's failing with deviceauth/master and passing with deviceauth/pull/423 as expected
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/219398268

tranchitella (approved):

PR #1092, opened 2020-11-23
[staging] Set mtls-ambassador version to 1.0.0

lluiscampos:


tranchitella (approved):

kacf (approved):
Why not `staging`?

lluiscampos:
> Why not `staging`?

@kacf It is not used inside Hosted Mender. There is an staging branch and saas tags, but as far as I know there are not in use (they are created automatically by the HM release process).

PR #1094, opened 2020-11-27
demo script: Fix misparsing of commands when using `-f` argument.

kacf:
If the `-f` argument is specified, we jump to the next parsing loop,
which doesn't contain parsing for docker commands. Move the parsing
there so it will be done correctly. This is fine because the first
loop will exit on unknown arguments, so we are guaranteed to get to
the second loop.

Changelog: None


kacf:
Using `Changelog: None` here, since even though the script is public, I don't think anyone uses `-f` besides us.

oleorhagen (approved):

PR #1097, opened 2020-11-30
Sync staging with master

tranchitella:


kacf (approved):

lluiscampos (approved):
The backend test failures are due to mender (client) version in use, 2.4.1, which does not support provides and depends.

We can upgrade mender (client) to master or just keep ignoring these errors.

tranchitella:
@lluiscampos I'll simply ignore those errors.

PR #1100, opened 2020-12-04
Use traefik

mchalski:
https://tracker.mender.io/browse/MEN-4060

needs:
https://github.com/mendersoftware/deviceauth/pull/432
https://github.com/mendersoftware/useradm/pull/205
https://github.com/mendersoftware/useradm-enterprise/pull/166
https://github.com/mendersoftware/inventory-enterprise/pull/130

example pipelines:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/231764028
(seems like spurious `context exceeded` error in state scripts tests - ugly new symptoms, old bug)

https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/232728722
(spurious MTLS error; state scripts error gone)

https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/232837838
(quite a few state scripts errors)

(all feature branches, including this one, were synced with master around Dec 17)

the first couple commits are the original proposal by Manuel, the rest are fixes that actually make the tests pass.

it's a lot, so to structure this here are the highlights:

**Conventional storage proxy is out**
normally traefik handles the storage domain (`s3...`) as well as the main gateway domain.

**New storage proxy is optional and external**. 
- It's just a HTTP 1.1 compat layer for legacy clients, used only in legacy tests (even 1.7.0 works then, of course)
- proxy is identified by `X-Forwarded-Host`
- updated my 'dirty design doc' with explanations here: https://docs.google.com/document/d/1M7DsyGQovNd4NUFh68emq-5G_IjoR9ptEa6nvDkuKOQ/edit#heading=h.5s6vg8fg34bb

**New test isolation mechanism**
Previously it was enough to isolate xdist setups via a network suffix. Traefik however ignores these, and is happy to pull any and all containers running at the same time into the routing pool, i.e. any containers reported by the docker provider. Result: cross talk and cross load balancing between containers from unrelated setups - extremely confusing.
Now: containers in each setup have a common `mender.testprefix` label, and traefik in this setup only allows the matching containers.

**More timing/syncing mechanisms**
Before, it was clear when a service is available - if it's up and healthy, nginx is ready to serve.

Traefik needs some time to register a service even if it's up and healthy - so don't just wait for services to be up, wait until all the expected api routers are registered with traefik. Use the traefik API to monitor that.

**Trickier depends_on**
Gateway can't depend on deployments anymore - deployments pings the storage domain at startup, so in fact deployments needs the gateway first. depends_on are relaxed, dependencies are really managed by the 'wait for routers' routine described above.

**Misc**
Lots of boring stuff in between: file encoding issues, tuning release components, minor bugfixes, fixes to specific tests.

I'm leaving it un-squashed so every little change has a comment for reference. 

TODOs:
bring production/enterprise envs up to speed, user-facing docs, etc. - followups will be identified later.


oleorhagen (approved):
I trust you on this!

kjaskiewiczz (approved):
+1 from me
it would be great if someone who worked with traefik in the past could take a look too

docker-compose.yml
```diff
@@ -22,6 +46,53 @@ services:
...
+            - traefik.enable=true
+            - traefik.http.routers.gui.entrypoints=https
+            - traefik.http.routers.gui.middlewares=ensure-ui-path,signup-redirect,ui-stripprefix,sec-headers,json-error-responder1,json-error-responder2,json-error-responder3,json-error-responder4
+            - traefik.http.routers.gui.rule=PathPrefix(`/`)
+            - traefik.http.routers.gui.service=gui
+            - traefik.http.routers.gui.tls=true
+            - traefik.http.services.gui.loadbalancer.server.port=80
+            
+            - traefik.http.routers.error-responder.entrypoints=https,http
+            # - traefik.http.routers.error-responder.rule=HostRegexp(`{host:.+}`)
```
mzedel, line 59: if I remember correctly that's the error catch-all - based on the passing tests I think the error responses seem to work just fine, so I don't think we need that line to be in 🤷‍♂️

mzedel (approved):
LGTM - especially the neat handling of the legacy support layer 🥰 and the waiting functionality for the tests... 💪!

mchalski:
feature branches were behind yet again (causing esp. failures around mender-connect/shell), so after another rebase: 
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/237546516

PR #1110, opened 2020-12-11
Merge feature branch deviceconnect

tranchitella:


docker-compose.connect.yml
```diff
@@ -1,27 +1,33 @@
...
+            - mender
+
+    mender-workflows-worker:
+        environment:
+            DEVICECONNECT_ADDR: "mender-deviceconnect:8080"
+            HAVE_DEVICECONNECT: 1
+
+    mender-api-gateway:
+        environment:
+            HAVE_DEVICECONNECT: 1
```
merlin-northern, line 33: ahy are adding that here? (asking since I do not know)

docker-compose.connect.yml
```diff
@@ -1,27 +1,33 @@
...
-      environment:
-          HAVE_DEVICECONNECT: 1
+    mender-nats:
+        image: nats:2.1-alpine3.11
+        networks:
+            - mender
+
+    mender-workflows-worker:
+        environment:
+            DEVICECONNECT_ADDR: "mender-deviceconnect:8080"
```
merlin-northern, line 28: consider adding `http://` ?

docker-compose.connect.yml
```diff
@@ -1,27 +1,33 @@
...
+    mender-deviceconnect:
+        image: mendersoftware/deviceconnect:mender-master
+        command: server --automigrate
+        extends:
+            file: common.yml
+            service: mender-base
+        networks:
+            - mender
+        depends_on:
+            - mender-mongo
```
merlin-northern, line 16: technically it depends on nets as well.

merlin-northern (changes requested):
3 comments only. I also checked with [the PR created from UI](https://github.com/mendersoftware/integration/pull/1111/commits) and reviewed the new commits.

docker-compose.connect.yml
tranchitella: Because we need to include the nginx configuration file for deviceconnect

docker-compose.connect.yml
tranchitella: All workflows have the http:// hard-coded, see for example:
https://github.com/mendersoftware/workflows/blob/master/worker/update_device_inventory.json#L11

We can change it if we think it is useful, but not in this PR because it is out of scope.

docker-compose.connect.yml
tranchitella: Right: https://github.com/mendersoftware/integration/pull/1110/commits/7922d1162cc0de988beba6353e513761dd1d1c46

docker-compose.connect.yml
merlin-northern: right. thanks. during the review I noticed that workflows differs in that from the rest, where we pass http:// in the env was for URL. right. perhaps a low priority follow up on this one.

merlin-northern (approved):
lgtm!

PR #1119, opened 2020-12-16
Add env variable HAVE_DEVICECONNECT to gui

tranchitella:
Changelog: none


tranchitella:
Cancel the QA pipeline, as there is no value in running it.

mzedel (approved):

PR #1129, opened 2021-01-01
MEN-4088: Add integration tests for mender-connect

tranchitella:


tranchitella:
Pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/236455939

tests/tests/test_mender_connect.py
```diff
@@ -33,12 +31,10 @@ def test_regular_protocol_commands(
...
-        if WRONG_BEHAVIOR:
-            # MEN-4273.
-            device.run(
-                """sed -i -e '2i"SkipVerify": true,' /etc/mender/mender-connect.conf"""
-            )
-            device.run("systemctl restart mender-connect")
+        device.run(
+            """sed -i -e '2i"SkipVerify": true,' /etc/mender/mender-connect.conf"""
+        )
+        device.run("systemctl restart mender-connect")
```
kacf, line 37: This was under a `if WRONG_BEHAVIOR`, and hence this is what should not happen. There are more of these below.

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -145,7 +145,10 @@ def _wait_for_containers(self, expected_containers):
                 "docker-compose -p %s %s ps -q" % (self.name, files_args), shell=True
             )
             running_countainers_count = len(out.split())
-            if running_countainers_count == expected_containers:
+            if running_countainers_count in (
+                expected_containers,
+                expected_containers + 1,
+            ):
```
kacf, line 151: Why not?

kacf (changes requested):

tests/tests/test_mender_connect.py
```diff
@@ -17,14 +17,12 @@
...
 from ..common_setup import class_persistent_standard_setup_one_client_bootstrapped
 from ..MenderAPI import devconnect
 
-if WRONG_BEHAVIOR:
-    logger = logging.getLogger("websockets")
-    logger.setLevel(logging.DEBUG)
-    logger.addHandler(logging.StreamHandler())
+logger = logging.getLogger("websockets")
+logger.setLevel(logging.DEBUG)
+logger.addHandler(logging.StreamHandler())
```
kacf, line 25: Oh, and this too, btw. It is *extremely* verbose, so I think it is too much for regular logs after this development has finished.

testutils/infra/container_manager/docker_compose_manager.py
tranchitella: because you break other tests (migrate os to ent) because of an extra container running (I suppose it is your session scope client).

tests/tests/test_mender_connect.py
tranchitella: ack

tests/tests/test_mender_connect.py
tranchitella: ack

testutils/infra/container_manager/docker_compose_manager.py
kacf: I don't understand why this would break other tests. The `NUM_CONTAINERS` variables have been updated, and the service is added to `BASE_FILES` so it should be included in all tests. I think the lines above are the wrong fix, because we are just relaxing conditions, but the number of containers should be exact.

tranchitella: @kacf it did break those tests, though. Look here: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/236766853

kacf: Hmm, ok. I'll investigate a bit more.

kacf: Doh, of course. I need to adjust `NUM_CONTAINERS` by 2, not 1, since `docker-compose.connect.yml` also contains the nats service.

I'll fix that and update it in my original https://github.com/mendersoftware/integration/pull/1120, along with your new commits. Closing this.

kacf:
Superseded by #1120.

PR #1130, opened 2021-01-03
Sync staging with master

tranchitella:


tranchitella:
Back-end integration tests fail because they depend on the new, not-yet-released Mender client.
Full integration tests (open source flavour) fail because they depend on the new, not-yet-released Mender client.

kacf (approved):

PR #1133, opened 2021-01-06
tests/run.sh: Fix inability to pass filenames into run.sh.

kacf:
Because the last argument is in quotes, it always gets included, even
if empty, and acts as a silent path wildcard which always includes all
files. Fix it by using ":+" notation to make the quoting conditional.
If the variable is empty it will not be included as an argument.

Changelog: None


kacf:
Wtf..? I cannot request @lluiscampos and @oleorhagen as reviewers, apparently...

kacf:
In addition to the [normal pipeline](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/238118740), I'm also running [a pipeline](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/238125228) where `SPECIFIC_INTEGRATION_TEST` is set, so that we verify it still works.

tests/run.sh
```diff
@@ -168,13 +168,19 @@ if ! python3 -m pip show pytest-html >/dev/null; then
...
+        "$@"
 fi
 
-python3 -m pytest \
-    $EXTRA_TEST_ARGS \
-    --verbose \
-    --junitxml=results.xml \
-    $HTML_REPORT \
-    "$@" \
-    $SPECIFIC_INTEGRATION_TEST_FLAG "$SPECIFIC_INTEGRATION_TEST"
```
oleorhagen, line 180: I know it's fugly, but we could do:
```
    $SPECIFIC_INTEGRATION_TEST_FLAG ${SPECIFIC_INTEGRATION_TEST_FLAG:+"${SPECIFIC_INTEGRATION_TEST}"}
```
instead of the if thingy

kacf: Oh interesting! I can't say I'm confident the quoting will work correctly there, but I agree it is much nicer. Trying...

lluiscampos:
> Wtf..? I cannot request @lluiscampos and @oleorhagen as reviewers, apparently...

It happened to me yesterday too...

lluiscampos (approved):
LGTM as-is.

No objections with using the fancy bash conditional suggested by @oleorhagen  neither.

tests/run.sh
kacf: That worked, nice one!

kacf:
Again running two pipelines, [one original](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/238469554) and one with [a specific integration test](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/238470298).

oleorhagen (approved):

PR #1135, opened 2021-01-06
Speed up test_state_scripts by using class-scoped setup fixture.

kacf:
Changelog: None


kacf:
@lluiscampos: This saves between 7-12 min of execution time compared to recent master branch runs!

kacf:
Failure is definitely unrelated, this does not affect backend integration tests at all.

lluiscampos (approved):
Nice optimization, thanks :+1: 

I check for possible side effects. The does clean-ups so it should be okey.

PR #1137, opened 2021-01-08
DEBUG: break legacy client

mchalski:
by removing the storagy proxy temporarily -
for development only.

changelog: none

### DIFF
--- a/tests/legacy-v1-client.yml
+++ b/tests/legacy-v1-client.yml
@@ -5,4 +5,4 @@ services:
     # mender-legacy-client 
     #
     mender-client:
-        image: mendersoftware/mender-client-qemu:1.7.0
+        image: mendersoftware/mender-client-qemu:2.0.0
--- a/testutils/infra/container_manager/docker_compose_manager.py
+++ b/testutils/infra/container_manager/docker_compose_manager.py
@@ -52,8 +52,6 @@ class DockerComposeNamespace(DockerNamespace):
     LEGACY_CLIENT_FILES = [
         COMPOSE_FILES_PATH + "/docker-compose.client.yml",
         COMPOSE_FILES_PATH + "/tests/legacy-v1-client.yml",
-        COMPOSE_FILES_PATH + "/storage-proxy/docker-compose.storage-proxy.yml",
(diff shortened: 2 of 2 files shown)


kacf:
And, it failed! Now I'm really confused... Let me retry on my branch as well, maybe there is some subtle difference we're not seeing.

mchalski:
@kacf however - note that it failed much less catastrophically - no `panic`s, nothing blew up:
```
Jan 11 14:19:23 qemux86-64 mender[243]: time="2021-01-11T14:19:23Z" level=info msg="Running Mender version 2.0.0" module="deployment_logger"
Jan 11 14:19:23 qemux86-64 mender[243]: time="2021-01-11T14:19:23Z" level=info msg="State transition: update-fetch [Download_Enter] -> update-store [Download_Enter]" module=mender
Jan 11 14:19:23 qemux86-64 mender[243]: time="2021-01-11T14:19:23Z" level=info msg="no public key was provided for authenticating the artifact" module=installer
Jan 11 14:19:23 qemux86-64 mender[243]: time="2021-01-11T14:19:23Z" level=error msg="update (234881024 bytes) is larger than the size of device /dev/hda3 (226492416 bytes)" module="dual_rootfs_device"
Jan 11 14:19:23 qemux86-64 mender[243]: time="2021-01-11T14:19:23Z" level=error msg="Artifact install failed: Payload: can not install Payload: core-image-full-cmdline-qemux86-64.ext4: no space left on device" module=state
```
ps. sorry - I generally trust your research more than mine, but I just had to check wth I was seeing back when I was testing 2.0...

kacf:
Good observation, @mchalski! This is because even though 2.0.0 was built at (almost) the same Mender client commit as 2.0.x, it was built with a different Yocto branch, which had smaller partition sizes. Hence this version cannot be used for this test, and the failure is indeed because the partition is too small.

The few differences in commits between 2.0.0 and 2.0.x do not change the source code in any way that could affect this (only [one line has changed and it is definitely not related](https://github.com/mendersoftware/mender/compare/2.0.0..0a07c40)). So I believe we are good, and my results in https://tracker.mender.io/browse/MEN-4244 still stand.

Unless you need it for other work, you can probably close this PR.

PR #1138, opened 2021-01-09
Sync staging with master

tranchitella:


tranchitella:
All the failures are related to the old Mender client.

PR #1139, opened 2021-01-11
docker-compose.auditlogs: enable auditlogs for the deviceconnect service

kjaskiewiczz:


tranchitella (approved):
This requires the inclusion of deviceconnect if you enable auditlogs.
Not a big deal, but we have to be aware of it.

PR #1142, opened 2021-01-13
Sync staging with master

tranchitella:


tranchitella:
Failures are due to the usage of the current stable Mender client instead of the release candidate one.

alfrunes (approved):

PR #1146, opened 2021-01-15
release_tool: Handle patch versions that have been released independently.

kacf:
Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -1145,22 +1145,32 @@ def find_prev_version(tag_list, version):
     return None
 
 
-def next_patch_version(prev_version, next_beta=None):
-    """Returns the next patch version is a series, based on the given version.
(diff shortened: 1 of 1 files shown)


lluiscampos (approved):
To be honest, I don't completely follow the logic here. But overall makes sense and I know you are releasing exactly under these circumstances right now, so it should be okey :)

kacf:
It was not entirely straightforward to figure it out either! The key takeaway is that before, it just blindly incremented the version by one, but now it actually checks whether this really is the next one. This results in two things: The correct new version is suggested (very important, I almost ended up not releasing a new client!), and it warns you when this happens, so you can do additional checking.

kacf:
For some reason this failed the Black test after it was merged. I suspect the pipeline didn't run, and I failed to notice. I pushed the fix straight to master, since it will block everyone and there is no functionality change.

PR #1147, opened 2021-01-17
MEN-4357: demo script, include docker-compose.connect.yml by default

tranchitella:
Changelog: title

### DIFF
--- a/demo
+++ b/demo
@@ -7,7 +7,7 @@
 # that aren't allowed.
 DOCKER_COMPOSE_PROJECT_NAME=${DOCKER_COMPOSE_PROJECT_NAME:-$(tr -d ".-" <<<${PWD##*/})}
 
-EXTRA_FILES=
+EXTRA_FILES=""
 
 CLIENT_FILES="-f docker-compose.client.yml -f docker-compose.client.demo.yml"
 ENTERPRISE_FILES="-f docker-compose.enterprise.yml"
@@ -180,6 +180,7 @@ run_non_up_commands() {
         # exit 1 will never happen.
         exec docker-compose \
                 -f docker-compose.yml \
+                -f docker-compose.connect.yml \
                 -f docker-compose.storage.minio.yml \
                 -f docker-compose.demo.yml \
                 -p ${DOCKER_COMPOSE_PROJECT_NAME} \
@@ -193,6 +194,7 @@ run_non_up_commands() {
 exitfunc() {
     retval=$(docker-compose \
             -f docker-compose.yml \
+            -f docker-compose.connect.yml \
             -f docker-compose.storage.minio.yml \
             -f docker-compose.demo.yml \
             -p ${DOCKER_COMPOSE_PROJECT_NAME} \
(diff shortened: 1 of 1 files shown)


demo
```diff
@@ -7,7 +7,7 @@
 # that aren't allowed.
 DOCKER_COMPOSE_PROJECT_NAME=${DOCKER_COMPOSE_PROJECT_NAME:-$(tr -d ".-" <<<${PWD##*/})}
 
-EXTRA_FILES=
+EXTRA_FILES="-f docker-compose.connect.yml"
```
kacf, line 10: The other default files are specified inline in several places. That's actually quite bad, but I think mixing the two approaches is worse. So I suggest either:

1. Clean up the whole approach, and specify all files here.
2. Or specify this together with the inline files instead, so that we keep them together.

kacf (changes requested):

demo
lluiscampos: I agree with this, sorry. The `EXTRA_FILES` has a clear functionality and we shouldn't mix it.

lluiscampos (commented):
Why have all integration tests passed? I was expecting almost all to fail due to the number of expected running containers, which should not match any more... I'll have a look at the code.

lluiscampos (changes requested):
> Why have all integration tests passed? I was expecting almost all to fail due to the number of expected running containers, which should not match any more... I'll have a look at the code.

Well, of course: the test framework does not use `demo` directly (except for the specific tests for this script) but they maintain [their own list of files](https://github.com/mendersoftware/integration/blob/master/testutils/infra/container_manager/docker_compose_manager.py#L37-L41). However, I would argue that such list should be updated adding `docker-compose.connect.yml` to run our tests with what is our now default environment.

This is open for discussion, what do you think?

demo
tranchitella: @lluiscampos @kacf ack, reworked as per your suggestions.

tranchitella:
> > Why have all integration tests passed? I was expecting almost all to fail due to the number of expected running containers, which should not match any more... I'll have a look at the code.
> 
> Well, of course: the test framework does not use `demo` directly (except for the specific tests for this script) but they maintain [their own list of files](https://github.com/mendersoftware/integration/blob/master/testutils/infra/container_manager/docker_compose_manager.py#L37-L41). However, I would argue that such list should be updated adding `docker-compose.connect.yml` to run our tests with what is our now default environment.
> 
> This is open for discussion, what do you think?

Don't know... as you can run Mender backend without connect, and it should work, I see value in testing it **without** connect and then, when needed, include it. But I don't have a strong opinion, if you believe it's better the other way around, let's do it.

kacf:
> Well, of course: the test framework does not use `demo` directly (except for the specific tests for this script) but they maintain [their own list of files](https://github.com/mendersoftware/integration/blob/master/testutils/infra/container_manager/docker_compose_manager.py#L37-L41). However, I would argue that such list should be updated adding `docker-compose.connect.yml` to run our tests with what is our now default environment.
> 
> This is open for discussion, what do you think?

I agree, but I think this is a generic cleanup that should not block this PR.

> Don't know... as you can run Mender backend without connect, and it should work, I see value in testing it **without** connect and then, when needed, include it. But I don't have a strong opinion, if you believe it's better the other way around, let's do it.

Although it's far from exhaustive, we have some dedicated tests for the production environment, which runs without connect.

kacf (approved):

lluiscampos (approved):

lluiscampos:
> I agree, but I think this is a generic cleanup that should not block this PR.

I'll create a task for it, probably after a topic I want to discuss in next week Grooming meetings

PR #1149, opened 2021-01-18
Add auditlogs and deviceconnect to production templates

alfrunes:
I noticed that auditlogs and deviceconnect were never added to the production templates, and no compose file enables audit logging for deviceconnect so I figured it might be best to put it in the enterprise template.

### DIFF
--- a/production/config/enterprise.yml.template
+++ b/production/config/enterprise.yml.template
@@ -14,3 +14,7 @@ services:
             # don't supply a tenant token. If empty, clients without a tenant
             # token will be rejected.
             DEVICEAUTH_DEFAULT_TENANT_TOKEN: ''
+
+    mender-deviceconnect:
+        environment:
+            DEVICECONNECT_ENABLE_AUDIT: "1"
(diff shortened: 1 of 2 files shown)


mchalski (approved):

alfrunes:
I guess we could add our own command-line options to the `run` script. I don't think it's very user friendly to let the user have to include the correct compose files and reconfigure the necessary env variables manually.

kacf:
This made me think: Why is deviceconnect optional at all? I see the need for the client part to be optional, but making it optional on the server side just seems like unnecessary hurdles (more docs, questions on Mender Hub, etc). We could still make it opt-out, of course, by keeping it in a separate file, but otherwise, why not just accept this PR and include it everywhere?

lluiscampos (approved):
> [...] why not just accept this PR and include it everywhere?

Okey. Makes sense.

lluiscampos:
(I'll take care of the cherry-pick to 2.6.x together with other fixes)

PR #1151, opened 2021-01-19
Make the number of reboots configureable

oleorhagen:
Should be pretty harmless imo


kacf (approved):
LGTM. But what do you need it for?

oleorhagen:
> LGTM. But what do you need it for?

I need it for the `python-client` failure tests. The reboot detector is only triggered once.

PR #1160, opened 2021-01-21
production template: configure mender-api-gateway as a storage proxy

kjaskiewiczz:


mzedel (commented):
my understanding was that in an environment without self-hosted storage the `DEPLOYMENTS_AWS_URI` would just point to that external storage, without proxying through the api gateway... looking at the prod template this seems to contain ?both? options (by including minio and the external/ s3 alias)...
to clearly isolate this I would include the alias in the `minio` service (as in the `docker-compose.storage.minio.yml`) - but maybe I misunderstood the prod setup in that area...

mchalski:
so I had a good look through our setups and docs, and here's my take:

@mzedel you raise a technically correct point, but for practical purposes I would not pursue the conclusion.

you're right that in general - gateway needs the storage alias only if minio is included. that suggests that the alias should live in the minio layer. otherwise, in the 'real s3' case, the external s3 url would end up in the gateway section, so deployments could not reach it I think; gateway would 'steal' the domain.

I have to say though imo it's safe to have this alias here in prod:
- `prod.yml` is pretty much only for setups with internal minio; the prod doc leaves really no room for substituting real s3. my understanding is `prod.yml == a safer, prod ready demo with minio`:
https://docs.mender.io/2.6/server-installation/production-installation

- OTOH: as is the case of many customizations we have a more dedicated section about s3 with some pointers only (no automation, less handholding):
https://docs.mender.io/2.6/server-installation/storage

so my final proposal is to:
- leave the alias in prod, it really plays nicely with the current instructions+automation
- clarify in the `storage` section that the alias should be removed (give a sed example, etc.)

mchalski (approved):

PR #1178, opened 2021-02-02
Fix: warning about MENDER_TESTPREFIX

tranchitella:
Export the MENDER_TESTPREFIX env variable to its current value, or empty string
if not set, to avoid the warning message below:

WARNING: The MENDER_TESTPREFIX variable is not set. Defaulting to a blank
string.

### DIFF
--- a/docker-compose.auditlogs.yml
+++ b/docker-compose.auditlogs.yml
@@ -22,7 +22,7 @@ services:
             - traefik.http.routers.auditlogs.tls=true
             - traefik.http.routers.auditlogs.service=auditlogs
             - traefik.http.services.auditlogs.loadbalancer.server.port=8080
-            - mender.testprefix=${MENDER_TESTPREFIX}
+            - mender.testprefix=${MENDER_TESTPREFIX:-""}
 
     mender-api-gateway:
         environment:
--- a/docker-compose.config.yml
+++ b/docker-compose.config.yml
@@ -19,7 +19,7 @@ services:
             - traefik.http.routers.deviceconfig.tls=true
             - traefik.http.routers.deviceconfig.service=deviceconfig
             - traefik.http.services.deviceconfig.loadbalancer.server.port=8080
-            - mender.testprefix=${MENDER_TESTPREFIX}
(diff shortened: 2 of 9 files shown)


lluiscampos (commented):
I think this fix belongs to all `docker-compose` files that expect this env variable.

tranchitella:
> I think this fix belongs to all `docker-compose` files that expect this env variable.

Not sure I get what you mean. What I'm trying to fix is the warning when I use `./demo`.
Do you think we can fix it in other ways?

mchalski (approved):

tranchitella:
> > I think this fix belongs to all `docker-compose` files that expect this env variable.
> 
> Not sure I get what you mean. What I'm trying to fix is the warning when I use `./demo`.
> Do you think we can fix it in other ways?

@lluiscampos can you please elaborate? Is this fix good to go, or we need to rework it somehow?

lluiscampos:
> > > I think this fix belongs to all `docker-compose` files that expect this env variable.
> > 
> > 
> > Not sure I get what you mean. What I'm trying to fix is the warning when I use `./demo`.
> > Do you think we can fix it in other ways?
> 
> @lluiscampos can you please elaborate?

I think it is a cleaner solution set a default value (empty, in this case) in all `docker-compose.*.yml` that use this `MENDER_TESTPREFIX` env variable, as we do for all the other env variables used in docker-compose files (as an example, [STORAGE_URL in docker-compose.storage.minio.yml](https://github.com/mendersoftware/integration/blob/master/docker-compose.storage.minio.yml#L19
).

This will achieve the same for `demo` script, and also remove the warnings if I manually launch `docker-compose <the files I want> up`.

> Is this fix good to go, or we need to rework it somehow?

It is good to go regardless of my comment: I did not block it and @mchalski approved it.

tranchitella:
@lluiscampos ack, I prefer your approach, I reworked the PR to set a default directly in the docker-compose files

lluiscampos (approved):

tranchitella:
Failure is unrelated (test_demo_artifact).

PR #1181, opened 2021-02-03
Add docker-compose.config.yml

oleorhagen:
This is required as to the deviceconfig service.

Changelog: None


lluiscampos (approved):
The fix is correct.

The commit message is somewhat misleading, IMO, as what this is doing is basically making release_tool aware of a file that can potentially be found in other branches, note really caring about the contents of the file.

Anyway, thank for for fixing!

PR #1200, opened 2021-02-11
Expose deviceconfig devices API

tranchitella:


kacf (approved):
Looks good to me.

The only thing I didn't understand was the `middlewares` line, but that's definitely my lacking understanding of traefik.

PR #1209, opened 2021-02-15
QA-268: test_state_scripts to use a lighweight update module

lluiscampos:
* [test_state_scripts] Modify suite to use dummy update module
The update module does nothing, just pretend to reboot by restarting
mender-client systemd service.
With this idea, we cannot check anymore that the device would end up in
the correct partition, but we do verify that the correct state scripts
are run (thus the state machine goes through the expected states).
Commenting out the last test case, which needs an actual full rootfs
update.

* [test_state_scripts] Modify test_reboot_recovery for update module
Similarly to previous commit, modify the reboot recovery cases to use
the dummy update module. Verify that the device reboots once
(spontaneous reboot).

### DIFF
--- a/tests/module-state-scripts-test
+++ b/tests/module-state-scripts-test
@@ -1,0 +1,27 @@
+#!/bin/sh
+
+set -e
+
+STATE="$1"
+
+case "$STATE" in
+
+    NeedsArtifactReboot)
+        echo "Yes"
+        ;;
+
+    SupportsRollback)
+        echo "Yes"
+        ;;
+
+    Download|ArtifactInstall|ArtifactRollback)
+        ;;
+
+    ArtifactReboot|ArtifactRollbackReboot)
+        systemctl restart mender-client
+        ;;
+
+esac
+
+exit 0
+
--- a/tests/tests/test_state_scripts.py
+++ b/tests/tests/test_state_scripts.py
@@ -418,10 +418,11 @@ REBOOT_TEST_SET = [
     (
         "simulate_powerloss_artifact_install_enter",
         {
-            "RebootScripts": ["ArtifactInstall_Enter_01",],
+            "RebootScripts": ["ArtifactInstall_Enter_02",],
             "ExpectedFinalPartition": ["OriginalPartition"],
             "ScriptOrder": [
                 "ArtifactInstall_Enter_01",
+                "ArtifactInstall_Enter_02",
                 "ArtifactInstall_Leave_01",
                 "ArtifactReboot_Enter_01",
                 "ArtifactReboot_Leave_01",
@@ -429,7 +430,8 @@ REBOOT_TEST_SET = [
                 "ArtifactFailure_Leave_89",
             ],
             "ExpectedScriptFlow": [
-                "ArtifactInstall_Enter_01",  # kill!
+                "ArtifactInstall_Enter_01",  # run one script to init log
+                "ArtifactInstall_Enter_02",  # kill!
                 "ArtifactFailure_Enter_01",  # run failure scripts
                 "ArtifactFailure_Leave_89",
             ],
@@ -439,7 +441,6 @@ REBOOT_TEST_SET = [
         "simulate_powerloss_in_commit_enter",
         {
             "RebootScripts": ["ArtifactCommit_Enter_89"],
-            "DoubleReboot": [True],
             "ExpectedFinalPartition": ["OriginalPartition"],
             "ScriptOrder": [
                 "ArtifactInstall_Enter_01",
@@ -469,7 +470,6 @@ REBOOT_TEST_SET = [
         "simulate_powerloss_in_artifact_commit_leave",
         {
             "RebootOnceScripts": ["ArtifactCommit_Leave_01"],
-            "DoubleReboot": [True],
             "ExpectedFinalPartition": ["OtherPartition"],
             "ScriptOrder": [
                 "ArtifactInstall_Enter_01",
@@ -554,13 +554,12 @@ class TestStateScripts(MenderTesting):
     @pytest.mark.parametrize("description,test_set", REBOOT_TEST_SET)
     def test_reboot_recovery(
         self,
-        class_persistent_standard_setup_one_client_bootstrapped,
+        class_persistent_setup_client_state_scripts_update_module,
         description,
         test_set,
-        valid_image,
     ):
 
-        mender_device = class_persistent_standard_setup_one_client_bootstrapped.device
+        mender_device = class_persistent_setup_client_state_scripts_update_module.device
         work_dir = "test_state_scripts.%s" % mender_device.host_string
 
         script_content = (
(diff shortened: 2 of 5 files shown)


tests/tests/test_state_scripts.py
```diff
@@ -348,71 +363,74 @@
...
-                "ArtifactRollbackReboot_Leave_01",
-                "ArtifactRollbackReboot_Leave_99",
-                "ArtifactFailure_Enter_22",
-                "ArtifactFailure_Enter_33",
-                "ArtifactFailure_Leave_44",
-                "ArtifactFailure_Leave_55",
-            ],
-        },
-    ),
+    # TODO: Move this test case out of the rest, requires full rootfs update
```
lluiscampos, line 366: @kacf What do you think about this test case?

It can't be tested with an update module, as the purpose is to boot into a broken image. We have other tests verifying broken images and the rollback mechanics, so the question here is: how breaking scripts version is different from other kinds of corruptions? Is it worth testing?

tests/tests/test_state_scripts.py
```diff
@@ -348,71 +363,74 @@
...
+                "ArtifactInstall_Enter_02",
                 "ArtifactInstall_Leave_01",
                 "ArtifactReboot_Enter_01",
                 "ArtifactReboot_Leave_01",
                 "ArtifactFailure_Enter_01",
                 "ArtifactFailure_Leave_89",
             ],
             "ExpectedScriptFlow": [
-                "ArtifactInstall_Enter_01",  # kill!
+                "ArtifactInstall_Enter_01",  # run one script to init log
```
lluiscampos, line 432: TBH I didn't get to the bottom of this one, but without initializing the `/data/test_state_scripts.log` before the reboot, the test hangs forever (until timeout, really). I think my workaround does not modify the test intention, so it should be okey.

tests/tests/test_state_scripts.py
```diff
@@ -23,10 +23,25 @@
...
 
+
+@pytest.fixture(scope="class")
+def setup_client_state_scripts_update_module(
+    class_persistent_standard_setup_one_client_bootstrapped,
+):
+    device = class_persistent_standard_setup_one_client_bootstrapped.device
+    device.put(
+        "module-state-scripts-test",
+        local_path="/home/lluis/northern.work/integration/tests",
```
lluiscampos, line 38: That is value-add quality code xD

tests/tests/test_state_scripts.py
```diff
@@ -23,10 +23,25 @@
...
 from ..helpers import Helpers
-from ..MenderAPI import deploy, logger
+from ..MenderAPI import deploy, logger, image
 from .mendertesting import MenderTesting
 from testutils.infra.device import MenderDeviceGroup
 
+
+@pytest.fixture(scope="class")
+def setup_client_state_scripts_update_module(
+    class_persistent_standard_setup_one_client_bootstrapped,
```
kacf, line 33: If you are basing it on `class_persistent`, then it should also be called `class_persistent` I think. Yes, it's long, but it's important when reading that this property is clearly communicated.

tests/tests/test_state_scripts.py
kacf: Lol!

tests/tests/test_state_scripts.py
kacf: It is different, because we have no other test which verifies that the client rolls back if the `/etc/mender/scripts/version` file is incorrect. Fortunately, this one is easy to solve: Just add an `ArtifactReboot_Leave` state script which corrupts the version, and an `ArtifactRollbackReboot_Leave` script which restores it.

tests/tests/test_state_scripts.py
```diff
@@ -597,31 +601,28 @@ def test_reboot_recovery(
...
-                # handle case where the client has not finished the update
-                # path on the committed partition, but new partition is installed,
-                # thus we will not get a valid entrypoint into the uncommitted parition(reboot_leave)
-                # and the client will thus reboot straight after starting, and u-boot will
-                # fall back to the committed partition
-                if test_set.get("DoubleReboot", False):
-                    reboot_detector.verify_reboot_performed(number_of_reboots=2)
-                else:
-                    reboot_detector.verify_reboot_performed()
+                reboot_detector.verify_reboot_performed()
```
kacf, line 625: I'm not quite clear on why this is still here? This only detects real reboots, not `systemctl restart` reboots, AFAIK.

tests/tests/test_state_scripts.py
```diff
@@ -422,7 +440,6 @@
         "simulate_powerloss_in_commit_enter",
         {
             "RebootScripts": ["ArtifactCommit_Enter_89"],
-            "DoubleReboot": [True],
```
kacf, line 425: If you add handling of `ArtifactRollbackReboot` (as mentioned in an earlier comment), you may be able to put this back. However, it is also pending the discussion about `verify_reboot_performed`, which I didn't quite follow.

kacf (changes requested):
Nice work! Not that much changed, after all!

tests/module-state-scripts-test
```diff
@@ -0,0 +1,33 @@
...
+
+    NeedsArtifactReboot)
+        echo "Yes"
+        ;;
+
+    SupportsRollback)
+        echo "Yes"
+        ;;
+
+    ArtifactDownload)
```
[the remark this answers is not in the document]
lluiscampos: Yes! Tthanks

tests/tests/test_state_scripts.py
lluiscampos: :+1:

tests/tests/test_state_scripts.py
lluiscampos: I suspected we had to test this, I wanted to make the skeptical question first :)

Thanks for the good suggestion! I am reworking this test case.

tests/tests/test_state_scripts.py
lluiscampos: Yes, in only detects that the spontaneous reboot (via `sysrq-trigger`) happened. Note that such check is completely removed on `test_state_scripts` below. So it is rather a validation of the test logic than the device behavior.

Do you suggest removing it completely, then?

tests/tests/test_state_scripts.py
lluiscampos: It won't, see reply below.

Originally I thought about having extra logic in `ArtifactRollback` to increment a counter in a `/data` log file or something to note the number of reboots and be able to verify reboots in all test cases. But I considered to be not that important: we verify that the correct state script is called via the existing log file, so no point in double verifying it for this specific state (reboot)

tests/module-state-scripts-test
```diff
@@ -0,0 +1,33 @@
...
+        ;;
+
+    ArtifactInstall)
+        ;;
+
+    ArtifactReboot)
```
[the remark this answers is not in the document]
lluiscampos: You are right, thanks.

lluiscampos (commented):
@kacf Thank you for your review! I am reworking it today ASAP (we still have a change to make it in this sprint).

lluiscampos:
@kacf Please have a second look now. I reworked following your comments and re-enabled `Corrupted_script_version_in_etc` case (last commit).

tests/tests/test_state_scripts.py
kacf: Oh, right, it is needed for the spontaneous reboots. No, then you're right, it should stay. It must stay actually, because otherwise you risk just reconnecting to the previous boot's ssh (this is why we introduced the reboot detector).

tests/tests/test_state_scripts.py
kacf: You're right, this is not important, it can be removed.

kacf (approved):
Good work!

lluiscampos:
Rebasing on latest master to get latest fixes on `backend-integration` tests.

lluiscampos:
`test_demo_artifact` seems to fail consistently, but it is not related to this PR. Merging!

oleorhagen:
I'm sorry I'm late to the party, but I have been perusing this a little. Don't have time to go in depth, but the general idea here is really good! We should at the next demo have some stats for how much faster our pipes will be after this and removing the client builds I think :)

PR #1210, opened 2021-02-15
Upgrade minio+traefik

mzedel:
just since I came across the minio version (while investigating storage problems that turned out to be a full disk 🙈) - unfortunately the newer ones don't contain a healthcheck, requiring us to maintain our own image if we want to stay up-to-date

and traefik for good measure - as the release is not too long ago + we are still steeped into finding issues 😅


tranchitella (approved):
LGTM!

PR #1211, opened 2021-02-15
aligned demo server location with the default demo server 

mzedel:
...that's used in other parts of the setup (client/ connect/ ui)

I hope that's not problematic in any other area

Changelog: None

### DIFF
--- a/demo
+++ b/demo
@@ -163,11 +163,11 @@ env_setup() {
     export MENDER_VERSION=$(awk -F':' '/mendersoftware\/mender-client/ {print $3}' docker-compose.client.yml)
     export MENDER_DEB_PACKAGE_VERSION=$MENDER_VERSION
 
-    MENDER_SERVER_URI="https://localhost"
+    MENDER_SERVER_URI="https://docker.mender.io"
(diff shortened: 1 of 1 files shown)


kacf (approved):

tranchitella (approved):

lluiscampos (commented):
I think this is incorrect: `MENDER_SERVER_URI` variable is only used within the script (iow local workstation!) to create the user or upload the artifact. If we change it to docker.mender.io, then we would need to ask users to modify their `/etc/hosts` to set this url to `localhost`. (does this make sense?)

lluiscampos:
Both @kacf and @tranchitella approving makes me believe I am missing something...

tranchitella (changes requested):
Actually @lluiscampos is right, and I didn't notice it because I have docker.mender.io in my /etc/hosts. Without it, the demo server, OS version doesn't start.

mzedel:
hm too bad, I too forgot about the hosts file change... sadly it seems to only be possible to get rid of the requirement in the demo script but not running the under the correct location... much disappoint 😞

PR #1214, opened 2021-02-16
tests/test_configuration: verify update was triggered by check-update

mchalski:


tranchitella (approved):
The test looks good!

mchalski:
hmm. consistent failures - the expected lines are not in the output...
the configuration is being downloaded, but simply by regular polling... not sure what's happening, will investigate.

lluiscampos (approved):

mchalski:
in the end it seems to work with additional wait logic.
the problem was with the device connecting too late to deviceconnect (409 from check-update)

PR #1215, opened 2021-02-18
gitlab: trigger doc tests on build

mchalski:
Issues: https://tracker.mender.io/browse/QA-266

ok, @lluiscampos @kacf please re-review.
I spent a short while trying to get rid of the tiangolo image to reuse python:3... for some reason it looked unnecessary(and a long docker pull) but I was wrong.
I did however remove version checks, I think it's mender-doc build specific, correct?

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -64,6 +64,33 @@ test:extra-tools:release-tool:
     # Run release-tool unit tests.
     - python3 -m pytest extra/test_release_tool.py
 
+test:docs:
+  image: tiangolo/docker-with-compose
+  services:
+    - docker:19.03.5-dind
+  except:
+    variables:
+      - $RUN_TESTS_STAGING == "true"
+
+  before_script:
+    - apk add bash git openssl pwgen python3 jq
+    - git config --global user.name "user"
+    - git config --global user.email "user@example.com"
+    - git clone --depth=1 https://github.com/mendersoftware/mender-docs.git mender-docs
+
+  script:
+    - cd mender-docs
+    - env TEST_OPEN_SOURCE=1 ./test_docs.py 07.Server-installation/03.Production-installation/docs.md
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -64,6 +64,31 @@ test:extra-tools:release-tool:
...
+
+  except:
+    variables:
+      - $RUN_TESTS_STAGING == "true"
+
+  only:
+    refs:
+      - merge_requests
+    changes:
+      - docker-compose*
```
kacf, line 89: I would suggest putting `demo` and `production/*` also in this list.

mchalski: (I'm changing this pr a lot and playing around still, so this might disappear):

I think it's a good idea but I'd like like to get an ACK from @lluiscampos

lluiscampos: This is now obsolete; but if we were to keep the `changes`, `production` should have been here (`demo` I think would have been irrelevant, even though harmless).

.gitlab-ci.yml
```diff
@@ -64,6 +64,33 @@ test:extra-tools:release-tool:
     # Run release-tool unit tests.
     - python3 -m pytest extra/test_release_tool.py
 
+test:docs:
+  image: tiangolo/docker-with-compose
+  services:
+    - docker:19.03.5-dind
+  except:
+    variables:
+      - $RUN_TESTS_STAGING == "true"
```
lluiscampos, line 73: Good thinking :+1:

lluiscampos (changes requested):
It looks just great, thanks!

We should not have a `changelog` tag here, as this is a pure ci or test change (nothing to be concerned from a user point of view). Please amend your commit (but no need to rerun the full pipeline, ping me and I'll force-merge).

mchalski:
@lluiscampos changelog removed

lluiscampos (approved):

PR #1216, opened 2021-02-19
Add mender-client-docker-addons Docker image

lluiscampos:
This will be a lightweight Docker image containing all Mender device
components (client + add-ons) but without dual rootfs (nor QEMU). To be
used in tests.

Integrated it also in release_tool as an image to be published from
integration repo itself.

### DIFF
--- a/component-maps.yml
+++ b/component-maps.yml
@@ -38,8 +38,10 @@ git:
     release_component: true
 
   integration:
-    docker_image: []
-    docker_container: []
+    docker_image:
+    - mender-client-docker-addons
+    docker_container:
+    - mender-client
     release_component: true
     independent_component: true
 
@@ -59,8 +61,8 @@ git:
 
   mender:
     docker_image:
-    - mender-client-qemu
     - mender-client-docker
+    - mender-client-qemu
     - mender-client-qemu-rofs
     docker_container:
     - mender-client
@@ -230,6 +232,13 @@ docker_image:
   mender-client-docker:
     git:
     - mender
+    docker_container:
+    - mender-client
+    release_component: true
+
+  mender-client-docker-addons:
+    git:
+    - integration
     docker_container:
     - mender-client
     release_component: true
@@ -404,8 +413,9 @@ docker_container:
     git:
     - mender
     docker_image:
-    - mender-client-qemu
     - mender-client-docker
+    - mender-client-docker-addons
+    - mender-client-qemu
     - mender-client-qemu-rofs
     release_component: true
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,9 +6,13 @@ include:
     file: '.gitlab-ci-check-python3-format.yml'
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-github-status-updates.yml'
+  - project: 'Northern.tech/Mender/mendertesting'
+    file: '.gitlab-ci-check-docker-build.yml'
 
 stages:
   - test
+  - build
+  - publish
 
 test:check-commits:
   except:
@@ -111,3 +115,13 @@ test:staging:backend-tests:
       - backend-tests/report_backend_integration_*.html
     reports:
       junit: backend-tests/results_backend_integration_*.xml
+
+build:docker:
+  variables:
+    DOCKER_REPOSITORY: mendersoftware/mender-client-docker-addons
+    DOCKER_DIR: extra/mender-client-docker-addons
+
+publish:image:
+  variables:
+    DOCKER_REPOSITORY: mendersoftware/mender-client-docker-addons
+    DOCKER_DIR: extra/mender-client-docker-addons
--- a/docker-compose.docker-client.addons.yml
+++ b/docker-compose.docker-client.addons.yml
@@ -1,0 +1,8 @@
+version: '2.1'
+services:
+
+    mender-client:
+        # Needs to be built in extra/mender-client-docker-addons
+        image: mendersoftware/mender-client-docker-addons:master
+        networks:
+            - mender
--- a/extra/mender-client-docker-addons/Dockerfile
+++ b/extra/mender-client-docker-addons/Dockerfile
@@ -1,0 +1,63 @@
+# Creates a container which acts as a bare bones non-VM based Mender
+# installation, for use in tests.
+FROM ubuntu:20.04 AS build
+
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt update && apt install -y make git build-essential golang liblzma-dev \
+    jq libssl-dev libglib2.0-dev
+
+ARG MENDER_CLIENT_VERSION=master
+ARG MENDER_CONNECT_VERSION=master
+
+RUN git clone https://github.com/mendersoftware/mender /src/mender
+RUN (cd /src/mender && git checkout $MENDER_CLIENT_VERSION)
+
+RUN git clone https://github.com/mendersoftware/mender-connect /src/mender-connect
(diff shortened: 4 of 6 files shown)


extra/mender-device-docker/entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+set -e
+
+/etc/init.d/ssh start
+cp /usr/share/dbus-1/system.d/io.mender.AuthenticationManager.conf /etc/dbus-1/system-local.conf
```
lluiscampos, line 5: I don't know why we need this. I took it from https://github.com/mendersoftware/gui/blob/master/tests/e2e_tests/dockerClient/clientEntrypoint.sh#L4

lluiscampos:
Also, it doesn't really work. Here the `docker run` output:
```
$ docker run dummy
 * Starting OpenBSD Secure Shell server sshd
   ...done.
dbus-daemon[17]: Failed to start message bus: Failed to bind socket "/var/run/dbus/system_bus_socket": No such file or directory
time="2021-02-19T09:25:41Z" level=warning msg="Could not connect to syslog daemon: Unix syslog delivery error. (use -no-syslog to disable completely)"
time="2021-02-19T09:25:41Z" level=info msg="Loaded configuration file: /etc/mender/mender.conf"
time="2021-02-19T09:25:41Z" level=info msg="No dual rootfs configuration present"
time="2021-02-19T09:25:42Z" level=info msg="State transition: init [none] -> init [none]"
time="2021-02-19T09:25:42Z" level=info msg="State transition: init [none] -> idle [Idle]"
time="2021-02-19T09:25:42Z" level=info msg="State transition: idle [Idle] -> authorize-wait [Idle]"
time="2021-02-19T09:25:42Z" level=info msg="State transition: authorize-wait [Idle] -> authorize [Sync]"
time="2021-02-19T09:25:42Z" level=info msg="Device keys not present or bootstrap forced, generating"
time="2021-02-19T09:25:42Z" level=error msg="Failure occurred while executing authorization request: Method: Post, URL: https://docker.mender.io/api/devices/v1/authentication/auth_requests"
time="2021-02-19T09:25:42Z" level=error msg="Authorize failed: transient error: authorization request failed: transient error: authorization request failed: Unknown url.Error type: dial tcp: lookup docker.mender.io on 84.208.20.110:53: no such host"

(process:19): GLib-GIO-CRITICAL **: 09:25:42.528: g_dbus_connection_emit_signal: assertion 'G_IS_DBUS_CONNECTION (connection)' failed
time="2021-02-19T09:25:42Z" level=info msg="State transition: authorize [Sync] -> authorize-wait [Idle]"
time="2021-02-19T09:25:49Z" level=info msg="Loaded configuration file: /etc/mender/mender-connect.conf"
time="2021-02-19T09:25:49Z" level=error msg="mender-shall dbus failed to connect, error: Could not connect: No such file or directory"
time="2021-02-19T09:25:49Z" level=error msg="Could not connect: No such file or directory"
```

I am working on it.

component-maps.yml
```diff
@@ -234,6 +236,13 @@ docker_image:
     - mender-client
     release_component: true
 
+  mender-device-docker:
+    git:
+    - integration
+    docker_container:
+    - mender-client
+    release_component: true
```
kacf, line 244: Ah, hehe, now I understand why you had concerns about the release_tool! But I *think* it should work. Logically it makes sense, at least.

But should it have entries in `docker_container` and `git` sections as well..?

extra/mender-device-docker/entrypoint.sh
mzedel: dbus didn't start without the file being present at that location... 🤷‍♂️ - I guess @kacf knows a lot more about dbus 😅

mzedel: that might depend on the base image `mendersoftware/mender-client-docker:master` uses - but the image building for the client images is opaque to me...

lluiscampos:
> Also, it doesn't really work. [...]
> I am working on it.

It "works" now, it was missing creating `/run/dbus` dir.

component-maps.yml
lluiscampos: > Ah, hehe, now I understand why you had concerns about the release_tool! But I think it should work. Logically it makes sense, at least.

I also *think* it should work, but let's be aware during next release...

> But should it have entries in docker_container and git sections as well..?

Yes. Aren't they there? For `git` in under `integration`, for `docker_container` is under `mender-client`. I think it is correct, but please check again.

extra/mender-device-docker/entrypoint.sh
lluiscampos: I'll leave this workaround (at least for now) unless Kristian knows the right fix.

extra/mender-device-docker/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# installation, for use in tests.
+FROM ubuntu:20.04 AS build
+
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt update && apt install -y make git build-essential golang liblzma-dev \
+    jq libssl-dev libglib2.0-dev
+
+ARG MENDER_CLIENT_VERSION=master
+ARG MENDER_CONNECT_VERSION=master
+ARG MENDER_CONFIGURE_VERSION=master
```
lluiscampos, line 11: @kacf This one is challenging in combination with the build job I am creating in `mender-qa`.

As you know, mender-configure will use versions like "latest", which then we need to translate here to master. Assuming that this is going to be a very stable repo, I am thinking of not taking this as an argument and just add logic in this Dockerfile to checkout latest git tag (or master, if no tags).

This will be correct most of the times, and we could still rework later if we really need to pass a version from mender-qa to here. What do you think?

lluiscampos: I reworked following my above suggestion. Let me know if you disagree.

lluiscampos:
I renamed the image to `mender-client-docker-addons`.

The reason is that `mender-qa` pipeline and scripts have many expression of the sort `if ... | grep mender-client` so it is very convenient that all four "client" images have the same prefix. (I started modifying the pipeline and scripts but realized half way that the readability was getting bad).

So, let me know what you think! We can modify the suffix for something else, but `mender-client-docker-` has to stay.

lluiscampos:
> Hello smile_cat I created a pipeline for you here: [Pipeline-259124958](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/259124958)
> Build Configuration Matrix

No, thank you. Verifying it https://github.com/mendersoftware/mender-qa/pull/486 with @ [Pipeline 259126676](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/259126676)

tranchitella (approved):

docker-compose.docker-device.yml
```diff
@@ -0,0 +1,8 @@
+version: '2.1'
```
mzedel, line 1: for consistencies sake I think renaming this file would be good, but might not be strictly necessary

mzedel (approved):

docker-compose.docker-device.yml
lluiscampos: You are right, thanks!

This file is actually not in use for now - but I plan to use it in integ tests soon :)

lluiscampos:
Last force-push contained a rename of a file not in use. Merging!

extra/mender-device-docker/Dockerfile
kacf: Sorry, a bit late to the party, but yes, I agree, better to leave it out if we can.

PR #1218, opened 2021-02-20
[test_mender_connect.py] Support running test for older yocto releases

lluiscampos:
Where systemd service is still called "mender".


lluiscampos:
See https://github.com/mendersoftware/meta-mender/pull/1304

tranchitella (approved):

PR #1219, opened 2021-02-22
Sync staging with master

tranchitella:


tranchitella:
Failures:
- configuration: expected (depends on unreleased client)
- demo_artifact: known issue

PR #1226, opened 2021-02-22
QA-269: run backend tests in k8s, make the EKS cluster name configurable

tranchitella:
Changelog: none


tranchitella:
Stopped the pipeline, because it doesn't add any value to this PR.

tranchitella:
In the context of this PR, only backend tests are important, and both are green.
We can ignore full integration test failures.

lluiscampos (approved):
All these look good to me :+1:

tranchitella:
@lluiscampos pipeline is green (check the first comment for the link).

PR #1231, opened 2021-02-23
extended the traefik security headers & limited tls settings to non-weak options

mzedel:
not sure if this could break client interactions... this should however give us a better security rating for HM as we are currently still supporting old/ weak TLS versions

Changelog: None


mzedel:

tranchitella:
@mzedel doesn't seem to make the tests happy...

mzedel:

mzedel:
based on the test failures it looks like this is mainly due to the client using a TLS 1.1 certificate 😟 if that's common we won't gain anything, otherwise we should upgrade the client certificate... - maybe @kacf, @oleorhagen or @lluiscampos can chime in there...

kacf:
You mean the demo certificate? Well, it is almost five years old so it's maybe not surprising. I think we can upgrade it without too many problems, but there are a few different places to cover, at least integration, meta-mender and mender repositories, off the top of my head.

mzedel:

lluiscampos:
@mzedel Is this PR worth a refresh?

mzedel:
> @mzedel Is this PR worth a refresh?

This would get rid of the changes to the `sec-headers` middleware, but unfortunately the other changes would require a more detailed look into the client tests + the tls variants that interact with HM (which likely prevent a strict alignment to whatever we set here I'm afraid) => closing

PR #1233, opened 2021-02-25
[MEN-4393] Integration tests for session logging

alfrunes:


tests/MenderAPI/deviceconnect.py
```diff
@@ -48,3 +48,22 @@ def get_websocket(self):
         )
 
         return ws
+
+    def get_playback_url(self, session_id, sleep_ms=None):
+        url = "wss://%s/api/management/%s/deviceconnect/sessions/%s/playback" % (
```
merlin-northern, line 53: arent we putting the urls in a global variables somewhere usually? (as far as I remember, but I maybe wrong)

tests/tests/test_mender_connect.py
```diff
@@ -180,3 +181,70 @@ def test_bogus_proto_message(
...
+                    return body
+
+        session_id = ""
+        session_bytes = b""
+        with devconnect.get_websocket() as ws:
+            # Start shell.
+            shell = proto_shell.ProtoShell(ws)
+            body = shell.startShell()
+            assert shell.protomsg.props["status"] == protomsg.PROP_STATUS_NORMAL
+            assert body == b"Shell started"
```
merlin-northern, line 208: can we have "Shell started" as a const/var?

merlin-northern (commented):
lovely! minor comment only.
one general comment: do you think we should make another test scenario where we send a couple of Mb and see that the session gets closed?

tests/MenderAPI/deviceconnect.py
alfrunes: This was a copy/paste/modify of the other class member function.

tests/tests/test_mender_connect.py
alfrunes: Sure

alfrunes:

tests/tests/test_mender_connect.py
```diff
@@ -180,3 +181,70 @@ def test_bogus_proto_message(
             assert prot.props["status"] == protomsg.PROP_STATUS_ERROR
             assert prot.protoType == 12345
             assert prot.typ == proto_shell.MSG_TYPE_SPAWN_SHELL
+
+    def test_session_recording(
+        self, class_persistent_standard_setup_one_client_bootstrapped
+    ):
+        def get_cmd(ws, timeout=1):
+            pmsg = protomsg.ProtoMsg(1)
```
kacf, line 189: Should use `proto_shell.PROTO_TYPE_SHELL` instead of 1.

tests/tests/test_mender_connect.py
```diff
@@ -180,3 +181,70 @@ def test_bogus_proto_message(
...
+            body = b""
+            while True:
```
kacf, line 191: The outer `while` is redundant, no? The inner `while` has only one exit condition, the exception, which immediately returns.

tests/tests/test_mender_connect.py
```diff
@@ -180,3 +181,70 @@ def test_bogus_proto_message(
             assert prot.props["status"] == protomsg.PROP_STATUS_ERROR
             assert prot.protoType == 12345
             assert prot.typ == proto_shell.MSG_TYPE_SPAWN_SHELL
+
+    def test_session_recording(
+        self, class_persistent_standard_setup_one_client_bootstrapped
+    ):
+        def get_cmd(ws, timeout=1):
```
kacf, line 188: I think `get_cmd()` does the same thing as the already existing `shell.recvOutput()`, doesn't it? Get output with a small wait?

merlin-northern (approved):
lgtm. do not forget to answer Kristian's comments :) and you can also comment on the general one from me ;-)

tests/tests/test_mender_connect.py
alfrunes: The only difference is that it filters all buffered messages by message type (`== command`) instead of making assertions.

alfrunes:

PR #1246, opened 2021-03-04
Restore staging to before the unintentional revert.

kacf:
1. Merges master into staging right before the unintentional commit.
2. Merges the unintentional commit using merge strategy "ours" (to keep the staging content)
3. Merges the rest of the changes since the unintentional commit using normal merge.

### DIFF
--- a/tests/tests/test_update_modules.py
+++ b/tests/tests/test_update_modules.py
@@ -1,10 +1,10 @@
-# Copyright 2020 Northern.tech AS
+# Copyright 2021 Northern.tech AS
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
 #    you may not use this file except in compliance with the License.
 #    You may obtain a copy of the License at
 #
-#        https://www.apache.org/licenses/LICENSE-2.0
+#        http://www.apache.org/licenses/LICENSE-2.0
 #
 #    Unless required by applicable law or agreed to in writing, software
 #    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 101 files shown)


tranchitella (approved):
Diff looks good!

tests/tests/test_update_modules.py
```diff
@@ -65,7 +65,7 @@ def make_artifact(artifact_file, artifact_id):
             deploy.check_expected_statistics(deployment_id, "failure", 1)
             deploy.check_expected_status("finished", deployment_id)
 
-            output = mender_device.run("mender -show-artifact").strip()
+            output = mender_device.run("mender -no-syslog -show-artifact").strip()
```
lluiscampos, line 68: Changes on this file I think are the result of a bad conflict resolution at some point. (It was incorrect before my mess, so I'll fix independently later on)

lluiscampos (approved):
Looks good. Thank you, Kristian.

kacf:
Regarding testing, do I use the regular "start pipeline" approach, or does this need special treatment?

tranchitella:
You can ask @mender-test-bot to start pipeline, right :)

lluiscampos:
> Regarding testing, do I use the regular "start pipeline" approach, or does this need special treatment?

Regular pipeline, as Fabio has triggered. The only special thing is that we expect some tests to fail (related to unreleased features from the client).

lluiscampos:
Looks good. Merging

PR #1248, opened 2021-03-05
[pipeline] Override staging specific jobs to no-op

lluiscampos:
integration is an special repo, which has an "staging" branch but not
with the same meaning as other repos (iow it does not build a Docker
image meant to be deployed in stating).

Kill this functionality from the templates by overriding the jobs.


lluiscampos:
The pipeline syntax is invalid because it requires first https://github.com/mendersoftware/mendertesting/pull/138

mchalski (approved):
another review won't hurt, but I think I can give an approve here

tranchitella (approved):

PR #1251, opened 2021-03-08
[tests] Adjust test_mender_connect.py to expect error msg on error

alfrunes:
`test_bogus_proto_message` was affected by the changes introduced with mendersoftware/mender-connect#17. Messages with a `proto` header that is not associated with a known protocol will now receive an error control message.

### DIFF
--- a/tests/tests/test_mender_connect.py
+++ b/tests/tests/test_mender_connect.py
@@ -176,11 +176,12 @@ class TestMenderConnect:
             msg = prot.encode(b"")
             ws.send(msg)
 
-            msg = ws.recv()
-            body = prot.decode(msg)
-            assert prot.props["status"] == protomsg.PROP_STATUS_ERROR
-            assert prot.protoType == 12345
-            assert prot.typ == proto_shell.MSG_TYPE_SPAWN_SHELL
+            data = ws.recv()
+            rsp = protomsg.ProtoMsg(0xFFFF)
+            rsp.decode(data)
+            assert rsp.typ == "error"
+            body = rsp.body
+            assert isinstance(body.get("err"), str) and len(body.get("err")) > 0
 
     def test_session_recording(
         self, class_persistent_standard_setup_one_client_bootstrapped
--- a/testutils/api/protomsg.py
+++ b/testutils/api/protomsg.py
@@ -64,6 +64,7 @@ class ProtoMsg:
 
     # Returns body, attributes can be fetched from the ProtoMsg object.
     def decode(self, buf):
+
         obj = msgpack.unpackb(buf)
         if type(obj.get("hdr")) is not dict:
             raise TypeError("Malformed protomsg received.")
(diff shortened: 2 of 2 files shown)


tests/tests/test_mender_connect.py
```diff
@@ -176,11 +176,12 @@ def test_bogus_proto_message(
...
-            body = prot.decode(msg)
-            assert prot.props["status"] == protomsg.PROP_STATUS_ERROR
-            assert prot.protoType == 12345
-            assert prot.typ == proto_shell.MSG_TYPE_SPAWN_SHELL
+            data = ws.recv()
+            rsp = protomsg.ProtoMsg(0xFFFF)
+            rsp.decode(data)
+            assert rsp.typ == "error"
+            body = rsp.body
+            assert isinstance(body.get("err"), str) and len(body.get("err")) > 0
```
kacf, line 184: Aren't errors expressed through the `status` field anymore? I'm a bit worried about changing basic protocol properties, mender-connect is already released, and works the way that the old test worked. Will this impact that?

alfrunes: The `shell` protocol is [not affected](https://github.com/mendersoftware/mender-connect/blob/b5db81a6e8025fb3ecd453a63785c603ab1b6dca/app/daemon.go#L508). However, this test case is sending a message with an unregistered protocol header (malformed message), for which the behavior hasn't really been specified on the client side ([until now](https://github.com/mendersoftware/deviceconnect/pull/45)).

kacf: Maybe it has not been specified in text, but it has been in code. The reason I insisted on writing this test in the first place was to have this unregistered protocol header case covered. If you send an unregistered protocol message to an existing client in the field, you will receive what is expressed in the removed code above.

So we cannot remove this. If we want to change the way that this is reported, we have to make sure that clients are prepared to deal with both error responses. But is it worth having that complexity?

What is the reason that the original error responses can not be used?

alfrunes: > So we cannot remove this. If we want to change the way that this is reported, we have to make sure that clients are prepared to deal with both error responses. But is it worth having that complexity?

I don't think the UI nor `mender-cli` actually accounts for this case.

> What is the reason that the original error responses can not be used?

These error responses are used as long as the `proto` header is equal to 1 (shell). I introduced a new reserved protocol for control messages that are only intended for managing sessions, and in this protocol I added an error message for expressing internal errors (such as session expiry, no handler registered for protocol, etc.). The problem with using the shell logic is that is going to break any message that has an actual schema. Why? Because you don't know if what you get is a binary blob of error text or an encoded message (e.g. msgpack or JSON object).

alfrunes: Based on our discussions in meetings, I've added a handshake for detecting the protocol version used by `mender-connect`. As a follow up we should add a test case using `mender-connect v1.0` and the old test case.

kacf: Ok, I created https://tracker.mender.io/browse/MEN-4533.

kacf (approved):

PR #1253, opened 2021-03-09
keygen: Change script and update demo certificate to use SAN

alfrunes:
Golang 1.15 is phasing out [1] relying on common name when validating the
certificates and in go1.17 the option to trust legacy certificates is
removed entirely [2].
The keygen script is updated to produce a single demo certificate for
both api gateway and storage proxy, relying on SAN to specify the
correct alternative hostname. Likewise the two environment variables
CERT_(API|STORAGE)_CN has been shorthanded to one CERT_CN variable. If
not specified, the script will prompt for a value for CERT_CN. A new env
variable CERT_SAN can be used to manually set the openssl subjectAltName
setting, otherwise a reasonable default is derived from the common name.

[1] https://golang.org/doc/go1.15#commonname
[2] https://golang.org/doc/go1.16#crypto/x509

changelog: Renewed demo certificate


docker-compose.demo.yml
```diff
@@ -42,15 +42,15 @@ services:
...
         volumes:
             - ./tls.toml:/config/tls.toml
-            - ./certs/api-gateway/cert.crt:/certs/cert.crt
-            - ./certs/api-gateway/private.key:/certs/private.key
-            - ./certs/storage-proxy/cert.crt:/certs/s3.docker.mender.io.crt
-            - ./certs/storage-proxy/private.key:/certs/s3.docker.mender.io.key
+            - ./cert/cert.crt:/certs/cert.crt
+            - ./cert/private.key:/certs/private.key
+            - ./cert/cert.crt:/certs/s3.docker.mender.io.crt
+            - ./cert/private.key:/certs/s3.docker.mender.io.key
```
kacf, line 48: This also needs to be changed [here](https://github.com/mendersoftware/integration/blob/master/production/config/prod.yml.template#L85-L99).

kacf (changes requested):
A few things:

* Does this need migration, or will it still work if you have the old key set? I think the latter, judging from `prod.yml.template`, but should be confirmed.

* This will need a lot of documentation changes under https://docs.mender.io/2.6/server-installation.

* Also needs to be changed in [the client](https://github.com/mendersoftware/mender/blob/master/support/demo.crt)

* Also needs to be changed [in meta-mender](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-demo/recipes-mender/mender-server-certificate/files/server.crt). Here it's important to append, not replace, because we still need to be able to build old clients.

mzedel (commented):
the keygen looks very comfortable now 🙂
with the SAN functionality we could possibly remove the requirement for multiple certificates in the api-gateway (looking at the certificate & SAN list from stackoverflow.com this should be perfectly fine)

as @kacf already noted, the production changes are required too - maybe limiting the certificate to a single one saves _some_ work 😅

alfrunes:

alfrunes:
@kacf @mzedel I've rebased to master and updated the PR to use the configuration files. The only thing missing is updating `mender-docs` which is conflicting with [MEN-4679](https://tracker.mender.io/browse/MEN-4679) which is in progress by @mchalski. I'm not entirely sure how to resolve the integration test failures though. Some tests relies on specific versions of the `mender-client-qemu` docker image which have the old demo certificate. Any suggestions?

extra/integration-testing/docker-compose.compat.yml
```diff
@@ -12,6 +12,9 @@ services:
     privileged: true
     environment:
       TENANT_TOKEN: ""
+    volumes:
+      # Provision demo certificate
+      - ./cert/cert.crt:/mnt/config/server.crt
```
kacf, line 17: Nice solution!

kacf (changes requested):
Looks good!

However, a change with significant ramifications for production like this one deserves a better changelog, I think. If upgrading, and both certificates exist, only one of the two certificates are used, correct? I think this change should be explained in the changelog, perhaps suggesting that the storage certificate be retired, and the specific lines containing "s3.docker.mender.io.crt" in `prod.yml` should be removed (since it is a template, changes we make to it won't be propagated to users automatically). If you can think of other necessary steps, this should also go in the changelog, I think.

If it's too detailed, we can of course direct people to our documentation instead, but then there should at least be a brief message about what has changed, and why it's necessary to consult the documentation.

alfrunes:
> Looks good!
> 
> However, a change with significant ramifications for production like this one deserves a better changelog, I think. If upgrading, and both certificates exist, only one of the two certificates are used, correct? I think this change should be explained in the changelog, perhaps suggesting that the storage certificate be retired, and the specific lines containing "s3.docker.mender.io.crt" in `prod.yml` should be removed (since it is a template, changes we make to it won't be propagated to users automatically). If you can think of other necessary steps, this should also go in the changelog, I think.
> 
> If it's too detailed, we can of course direct people to our documentation instead, but then there should at least be a brief message about what has changed, and why it's necessary to consult the documentation.

I agree, I will update the changelog as soon as soon as I've updated the docs to keep things aligned in case something changes along the way.

mchalski:
@alfrunes @kacf @mzedel late to the party, I'll chime in on a couple points because I'm the one gatekeeping both the production install and the new production docs. indeed, a lot of changes are ahead and I'm not sure how much of this will transfer to the helm chart install (not much).

> Golang 1.15 is phasing out [1] relying on common name  ...
good catch, the SAN issue bit me when working on the new production install path.
another issue is: the cert(s) must not be 0-depth, there must be a root CA on the way.

so, as a matter of principle I'd be all for this PR but I'll be honest: I wanted to get rid of the keygen completely. 

you didn't have a chance to look at any PRs yet but let me summarize some critical changes I'm working on:
- we'll move away from the prod template and all the git machinery towards a simple `helm install ...`
- we won't be using 2 domains in prod anymore; 1 is enough by default to serve both the gateway functionality and the 'storage proxy' (the old proxy is indeed out); users are confused and don't want extra domains anyway
- I'd strongly suggest to reduce the amount of handholding we do in general (point the user to standard methods of generating certs and keys, but don't automate it for them; maybe have these guidelines as a footnote)

so, the keygen has some usefulness (for the demo maybe?), the new demo cert is certainly useful, but this won't be in the front and center of the new prod install. I wouldn't propagate these changes beyond the demo layer.

alfrunes:
> @alfrunes @kacf @mzedel late to the party, I'll chime in on a couple points because I'm the one gatekeeping both the production install and the new production docs. indeed, a lot of changes are ahead and I'm not sure how much of this will transfer to the helm chart install (not much).
> 
> > Golang 1.15 is phasing out [1] relying on common name  ...
> > good catch, the SAN issue bit me when working on the new production install path.
> > another issue is: the cert(s) must not be 0-depth, there must be a root CA on the way.
> 
> so, as a matter of principle I'd be all for this PR but I'll be honest: I wanted to get rid of the keygen completely.
>
> you didn't have a chance to look at any PRs yet but let me summarize some critical changes I'm working on:
> 
>     * we'll move away from the prod template and all the git machinery towards a simple `helm install ...`
> 
>     * we won't be using 2 domains in prod anymore; 1 is enough by default to serve both the gateway functionality and the 'storage proxy' (the old proxy is indeed out); users are confused and don't want extra domains anyway

I think it could be valuable to support multiple sub-domains under one common name (as in `*.docker.mender.io`) so that we don't clutter the Traefik routing configuration (or more specifically the UI redirect logic), but also to allow [virtual-host-style bucket URLs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAPI.html) for minio.

>     * I'd strongly suggest to reduce the amount of handholding we do in general (point the user to standard methods of generating certs and keys, but don't automate it for them; maybe have these guidelines as a footnote)

+1

> so, the keygen has some usefulness (for the demo maybe?), the new demo cert is certainly useful, but this won't be in the front and center of the new prod install. I wouldn't propagate these changes beyond the demo layer.

I generally agree with your comments, but I'd suggest at least keeping the demo certificate. But I definitely don't think we should commend the usage of self-signed certificates in a production setup.

The reason I created this PR in the first place is because of the dependency on the demo certificate in the integration test environment. This in turn keeps the Deployment service's Go version pinned at `1.14` because `1.15 >=` is rejecting any certificate that relies on the common name field by default.

alfrunes:

keygen
```diff
@@ -25,40 +46,51 @@ if [ "$OPENSSL_VERSION_MAJOR" -lt 1 ]; then
...
+fi
+
+CERT_VALID_DAYS="3650"
+FILE_NAME_PRIVATE_KEY="private.key"
+FILE_NAME_CERT="cert.crt"
 
 ROOTDIR=$(pwd)/keys-generated
 
-CERTDIR=$ROOTDIR/certs
+CERTDIR=$ROOTDIR/cert
```
kacf, line 70: Any particular reason this needs to be changed? Especially since træfik uses `certs` in its path.

kacf (changes requested):
I think my comment about changelog and the production template [still stands](https://github.com/mendersoftware/integration/pull/1253#pullrequestreview-676004343), but otherwise this looks in pretty good shape now!

What is the status of the certificates in mender client and meta-mender? It might be enough to simply replace them after this is merged, or?

tests/legacy-v1-client.yml
```diff
@@ -6,3 +6,6 @@ services:
     #
     mender-client:
         image: mendersoftware/mender-client-qemu:1.7.0
+        volumes:
+          # Provision demo certificate
+          - ./cert/cert.crt:/mnt/config/server.crt
```
mzedel, line 11: does this not also need the `      - ./extra/integration-testing/setup-mender-configuration.py:/setup-mender-configuration.py` mount?
`

keygen
```diff
@@ -25,40 +46,51 @@ if [ "$OPENSSL_VERSION_MAJOR" -lt 1 ]; then
...
+    echo -n "Enter a certificate common name [docker.mender.io]:"
+    IFS='' read -r CERT_CN
+    if [ -z "$CERT_CN" ]; then
+        CERT_CN="docker.mender.io"
+    fi
+fi
+
+if [ -z "$CERT_SAN" ]; then
+    CERT_SAN="DNS:${CERT_CN},DNS:*.${CERT_CN}"
+    echo "Subject alternative name (CERT_SAN) not set using default:"
```
mzedel, line 60: just since it confused me for a second 🙈:
```suggestion
    echo "Subject alternative name (CERT_SAN) not set, using default:"
```

mzedel (approved):
Thank you so much for sticking with this!
For the image manipulation I will have to defer to others for the final verdict, but it looks ok to me.
The rest (so mainly the keygen) looks very good 👌!

keygen
alfrunes: Guess I got a little too enthusiastic about merging the demo certificates into a common one :smiley_cat: 
I don't really want to revert it though (unless there is a strong preference) as it's already consistent with all the compose files.

tests/legacy-v1-client.yml
alfrunes: This version of the client does not use `mender-connect` or any other services that needs the certificate, so the default entrypoint script is enough to update `/etc/mender/mender.conf` to point to this mount.

keygen
alfrunes: That is confusing indeed. Good catch :+1:

alfrunes:
> I think my comment about changelog and the production template [still stands](https://github.com/mendersoftware/integration/pull/1253#pullrequestreview-676004343), but otherwise this looks in pretty good shape now!

I'll update the changelog tag for the main commit.

> What is the status of the certificates in mender client and meta-mender? It might be enough to simply replace them after this is merged, or?

The new certificate is appended to the chain so the old certificates are still trusted when built for the demo environment (the [mender-qa pipeline](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/363057884) also requires these PRs). I'll merge those PRs simultaneously as this one. As for remaining work, I still need to make some adjustments to the documentation on "production setup using Docker Compose" to reflect the new paths and parameters to the `keygen` script.

alfrunes:
Backend integration-test failure is caused by race with parallel work. :checkered_flag:

kacf (approved):

PR #1254, opened 2021-03-10
Men 4518: Integration test

oleorhagen:


.gitignore
```diff
@@ -31,3 +31,5 @@ docker-compose.override.yml
 backend-tests/acceptance.*
 backend-tests/report_backend_integration_enterprise.html
 backend-tests/results_backend_integration_enterprise.xml
+
+.gitattributes
```
kacf, line 35: Why? This is a very common file to have in a repository. There is a separate place for "your own" gitattributes, they should be put in `$GIT_DIR/info/attributes`.

kacf (approved):
I would omit the gitignore entry (see comment), but otherwise looks beautiful!

oleorhagen:
Pipe passed last time around, only difference is the `.gitignore` commit has been removed. 

Merging

lluiscampos:
:+1:

PR #1193, opened 2021-02-09
[python-requirements.txt] Add cryptography

lluiscampos:
And set it to the previous to last version, not introducing Rust
dependency. We'll update it in a different PR.

PR #1255, opened 2021-03-12
pip freeze the requirements.txt file(s)

lluiscampos:
For the integration one, also removing the Python3 specification, as we
don't require Python 2 any more.

---------------------------------

This is a request for comments, what do you guys think?
* On the one hand this is the "right way" to manage pypi dependencies so that we have stricter control on what gets updated (remember that the cryptography incident happen because we didn't had this indirect dependency in the reqs file, see #1193)
* while on the other hand it will explode the number of dependabot PRs on Mondays (note that I usually cherry-pick all of them in a single PR and run tests once, but still takes my time to cherry-pick all and resolve conflicts).


oleorhagen:
Makes sense to me. I don't think dependabot should dictate what we do here. 
One possibility could be to have mender-test-bot batch them for us?

oleorhagen (approved):

tranchitella (approved):
I agree with @oleorhagen

lluiscampos:
Edited the title, not an RFC any more :)

@mender-test-bot please start pipeline :robot:

PR #1261, opened 2021-03-18
[release_tool] Update yml name for integration special logic

lluiscampos:
The "yaml" version of integration is now the name of the Docker image
that publishes. In order for the special handling in `version_of` to
kick in, it needs to use the Docker image name


lluiscampos:
> The "yaml" version of integration is now the name of the Docker image
> that publishes. In order for the special handling in `version_of` to
> kick in, it needs to use the Docker image name

It used to work because of [this line](https://github.com/mendersoftware/integration/blob/master/extra/release_tool.py#L189).

kacf (approved):

PR #1264, opened 2021-03-19
MEN-4533: Integration checking API compatibility with mender-connect v1.0

alfrunes:

PR #1275, opened 2021-03-23
Mark test_mtls_enterprise_hsm and test_demo_artifact with xfail

alfrunes:
Allow these tests (yet still keep track of their failure statistics) until the issues have been resolved.
I also cleaned up the duplicate variable introduced in #1264.

### DIFF
--- a/tests/tests/test_demo_artifact.py
+++ b/tests/tests/test_demo_artifact.py
@@ -103,6 +103,7 @@ class TestDemoArtifact(MenderTesting):
     # Give the test a timeframe, as the script might run forever,
     # if something goes awry, or the script is not brought down properly.
     @pytest.mark.timeout(3000)
+    @pytest.mark.xfail(reason="Test is known to be unstable: FIXME")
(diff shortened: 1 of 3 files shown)


alfrunes:

alfrunes:

Sorry, code was formatted using the newer black version. I also added xfail to `test_mtls_enterprise` as this one also happened to fail for the previous run.

lluiscampos (commented):
Thank you for the initiative, but I think we should go for @oleorhagen suggestion in https://github.com/mendersoftware/integration/pull/1276 What do you think?

alfrunes:
> Thank you for the initiative, but I think we should go for @oleorhagen suggestion in #1276 What do you think?

I'm just tired of running the integration tests is turning into a dice game. Although flaky will up the odds, there's still a significant chance that the jobs will fail due to the same (known) errors. IMHO opinion I think we should temporarily ignore the result of the erroneous tests (but still keep track of them) until they're fixed.

alfrunes:
Decided to start with flaky and see how it goes.

PR #1284, opened 2021-03-25
limited minio api to only serve as mender storage

mzedel:
Changelog: None


mzedel:

tranchitella:
@mzedel do we still need this PR? If yes, we'll have to move your changes to the YAML files now that we moved away from labels.

mzedel:
@tranchitella you're right - the minio access is now no longer possible 👌 closing here and the ticket

PR #1285, opened 2021-03-25
Update the gitdm tool

oleorhagen:
This removes the subfolder containing the old gitdm sources, and adds a submodule, set to latest master. This enables the use of `Python3` with the statistics generator.


lluiscampos (commented):
I am completely in favor of this, but I'll let the floor to @kacf to review in case there is a good historical reason for this.

oleorhagen:
Whether or not it should be a submodule is also up for grabs

kacf (changes requested):
I agree with everything, it's great both to update it, keep it updated, and get rid of Python 2.

However, I do think that this requires us to write a test which tests the formatting done by the `post_process` stage. This is very sensitive to small changes in the output, and probably will break with new updates to this tool. A suggested test would be one that takes a hard coded range of history, and feeds it to gitdm and the `post_process` filter, and makes sure the output looks the same as the prerendered output.

lluiscampos (approved):
Nice test, great job!

extra/test_statistics_generator.py
```diff
@@ -0,0 +1,102 @@
...
+import pytest
+
+
+def test_post_process():
+    """
+    Test and verify that the output of the post_process step in the statistics
+    generator stays consistent through updates to the gitdm module
+    """
+    # ./statistics-generator --repo 2.5.0..2.6.0
+    prerendered_output = """Processed 102 csets from 8 developers
```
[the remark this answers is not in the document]
oleorhagen: Generated from an `Ubuntu:18.04` container with a cloned `integration:master` repo

.gitlab-ci.yml
```diff
@@ -68,10 +68,14 @@ test:extra-tools:release-tool:
...
     # Fetch master branch for tests using --in-integration-version
     - git fetch origin master:master
+    - git fetch origin --tags
+    - git submodule update --init --recursive
 
   script:
     # Run release-tool unit tests.
     - python3 -m pytest extra/test_release_tool.py
+    - git config --global mailmap.file $(pwd)/extra/gitdm/mailmap
+    - python3 -m pytest extra/test_statistics_generator.py
```
kacf, line 78: There is a different job defined above which tests the changelog generator; I think it fits better there.

But on the other hand, I think it's overkill to have separate jobs for these tests, they execute in fractions of seconds. Maybe you can just squash this job with the previous job and call it `test:extra-tools`?

kacf (approved):
Nice test! I only have a suggestion, can be merged as is too!

.gitlab-ci.yml
oleorhagen: Sure, @lluiscampos unless you object

lluiscampos: Makes sense to group them :+1:

oleorhagen:
Merged the tests, and all green :green_circle: 
No need to run the full `QA` pipeline here - Merging.

PR #1286, opened 2021-03-25
[Træfik] Change dynamic config from label- to file-based

alfrunes:
This will allow us to share a greater part of the configuration with the
production Kubernetes setup.

### DIFF
--- a/docker-compose.enterprise.yml
+++ b/docker-compose.enterprise.yml
@@ -22,21 +22,6 @@ services:
         extends:
             file: common.yml
             service: mender-base
-        labels:
-            - traefik.enable=true
-            - traefik.http.routers.tenantadm.entrypoints=https
-            - traefik.http.routers.tenantadm.middlewares=circuitbreaker,userauth,sec-headers,compression
-            - traefik.http.routers.tenantadm.rule=PathPrefix(`/api/management/{v[0-9]+}/tenantadm`)
-            - traefik.http.routers.tenantadm.tls=true
-            - traefik.http.routers.tenantadm.service=tenantadm
-            - traefik.http.services.tenantadm.loadbalancer.server.port=8080
-
-            - traefik.http.routers.tenantadmSignup.entrypoints=https
-            - traefik.http.routers.tenantadmSignup.middlewares=circuitbreaker,sec-headers,compression
-            - traefik.http.routers.tenantadmSignup.rule=Path(`/api/management/{v[0-9]+}/tenantadm/tenants`) || Path(`/api/management/{v[0-9]+}/tenantadm/tenants/trial`) && Method(`OPTIONS`,`POST`)
-            - traefik.http.routers.tenantadmSignup.tls=true
-            - traefik.http.routers.tenantadmSignup.service=tenantadm
-            - mender.testprefix=${MENDER_TESTPREFIX:-""}
         networks:
             - mender
         depends_on:
@@ -55,8 +40,8 @@ services:
             USERADM_HAVE_ADDONS: 1
 
     mender-api-gateway:
-        environment:
-            HAVE_MULTITENANT: 1
+      volumes:
+        - ./config/traefik/traefik.enterprise.yaml:/etc/traefik/config/traefik.enterprise.yaml
 
     mender-gui:
         environment:
--- a/testutils/common.py
+++ b/testutils/common.py
@@ -19,9 +19,11 @@ import string
 import tempfile
 import os
 import subprocess
+from contextlib import contextmanager
+
+import docker
 import redo
 import requests
-from contextlib import contextmanager
 
 import testutils.api.deviceauth as deviceauth
 import testutils.api.tenantadm as tenantadm
@@ -44,7 +46,6 @@ def clean_mongo(mongo):
     pymongo.MongoClient connected to the DB."""
     mongo_cleanup(mongo)
     yield mongo.client
-    mongo_cleanup(mongo)
 
 
 def mongo_cleanup(mongo):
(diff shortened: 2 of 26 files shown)


alfrunes:

alfrunes:

alfrunes:

alfrunes:

mzedel (commented):
the docker provider to file provider changes look good to me, but I would expect the omission of the `MENDER_TESTPREFIX` handling will lead to failing tests with multiple compositions running in parallel... looking at the træfik docs I'm not sure if there is a way to constrain/ tie the services defined in the files to the specific containers...

alfrunes:
> the docker provider to file provider changes look good to me, but I would expect the omission of the `MENDER_TESTPREFIX` handling will lead to failing tests with multiple compositions running in parallel... looking at the træfik docs I'm not sure if there is a way to constrain/ tie the services defined in the files to the specific containers...

I thought `MENDER_TESTPREFIX` was introduced because the label based configuration makes routing (& load balancing) based on Docker labels. As the parallel compositions will run on separate docker networks I don't think this issue is present with file-based configuration. Which is why I removed the env variable.

mzedel:
ah right, I was expecting the checking of the docker socket to still get the træfik services associated to the containers based on their docker service names... so:

> ~@alfrunes~ @mzedel, Let me know if you want to start the integration pipeline by mentioning me and the command "start pipeline".

🤡

mzedel:
ah right, I was expecting the checking of the docker socket to still get the træfik services associated to the containers based on their docker service names... so:

> @alfrunes @mzedel, Let me know if you want to start the integration pipeline by mentioning me and the command "start pipeline".

@mender-test-bot 🤡

mchalski:
@alfrunes I'll finish the review later, just a comment on 

> Which is why I removed the env variable.

double and triple check please, but I think you might be correct!

thinking out loud: 
the docker provider supplied container events to traefik disregarding the network - so, across network boundaries(=setups). containers from diff. networks would be added to traefik's load balancing pool this way. then a round robin would sometimes hit them, and that's a 502 and heisenbugs all around.

I think the file provider will have no other choice but to respect network boundaries. let's just make sure this is really what happens with >1 setups.

tranchitella:
@alfrunes what about the failures in the backend integration tests?

alfrunes:
@mender-test-bot :do_not_litter:

alfrunes:
> @alfrunes, Let me know if you want to start the integration pipeline by mentioning me and the command "start pipeline".

@mender-test-bot :do_not_litter:

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -166,7 +166,7 @@ def _wait_for_containers(self, expected_containers):
             )
             running_countainers_count = len(out.split())
             if running_countainers_count == expected_containers:
-                wait_for_traefik(self.get_mender_gateway())
+                wait_util_healthy(self.name)
```
mchalski, line 169: until

testutils/common.py
```diff
@@ -329,18 +329,19 @@ def wait_for_traefik(gateway_host, routers=[]):
...
-            "deviceauth@docker",
-            "deviceauthMgmt@docker",
-            "inventoryMgmt@docker",
-            "inventoryMgmtV1@docker",
-            "useradm@docker",
-            "useradmNoAuth@docker",
-            "deviceauth@docker",
-            "deviceauthMgmt@docker",
-            "inventoryV1@docker",
+            "deployments@file",
```
mchalski, line 332: killing my most precious helpers, huh? :)

so do I understand correctly that we don't need it because the file provider will register the routes immediately when the containers are healthy? - hence - waiting for 'healthy' is now our primary concern?

if yes, I'm ok with this.

it's just important to know if indeed there is no window where everything is healthy and routes are not fully registered yet. otherwise - it's a nightmare to debug, so please make sure!

mchalski: ...ok, so I misunderstood, but indeed wait_for_traefik is not used anymore.

so the question of the possibly small time window when routes are not registered still stands. sorry for the confusion :)

alfrunes: > killing my most precious helpers, huh? :)

Sorry! :see_no_evil: 

But seriously though, I figured that the tests were initially failing because of another setup race condition where the `backend-test-runner` would start before the entire backend was up, so I created this health-check function to wait for all services to become healthy. However, as I did not find a robust way to check if the entire docker-composition has been created I ended up modifying the `run` and entrypoint for the test-runner itself after already replacing the `wait_for_traefik` with the health check. As the file configuration is loaded immediately after Traefik starts, I was hoping that the `/ping` endpoint would not return successfully until the config is set up, but I did not find any documentation to back this up and your probably right.

alfrunes:
> @alfrunes, Let me know if you want to start the integration pipeline by mentioning me and the command "start pipeline".

@mender-test-bot :do_not_litter: :pray:

alfrunes:

alfrunes:

alfrunes:
@mchalski, @mzedel, @tranchitella the PR is ready for review now. All the tests pass except for `test_demo_artifact` which seems to fail pretty consistently for this PR.

docker-compose.enterprise.yml
```diff
@@ -55,8 +40,8 @@ services:
             USERADM_HAVE_ADDONS: 1
 
     mender-api-gateway:
-        environment:
-            HAVE_MULTITENANT: 1
+      volumes:
+        - ./config/traefik/traefik.enterprise.yaml:/etc/traefik/config/traefik.enterprise.yaml
```
mzedel, line 44: for consistency maybe:
```suggestion
        - ./config/traefik/traefik.enterprise.yaml:/etc/traefik/config/traefik.enterprise.yaml:ro
```

mzedel (approved):
LGTM! especially the DRY templates for the circuit breaker 🙂

tranchitella (approved):

PR #1299, opened 2021-03-31
Make sure version_of for integration returns proper versions.

kacf:
Previously if we specified HEAD, it would return HEAD. While not
exactly wrong, this name is not very useful for pushing docker tags.
Fix by looking for tag matches and symbolic refs to branch names, only
returning HEAD if we can find neither of those.

Changelog: None


kacf:
@mender-test-bot: start pipeline

tranchitella (approved):
LGTM.

You probably don't need the `output` variable at all, and you could simply return execute_git(...), but doesn't really matter.

kacf:
Indeed, you're right. I will fix that, because I have to change it slightly. There is one case not covered: If you check out `master~0` instead of `master` ("detached mode", which I think Gitlab uses), then it will still return HEAD.

PR #1312, opened 2021-04-12
Sync staging with master

tranchitella:


tranchitella:

tranchitella:
Failure in the main pipeline is because of `mender-docs`.
Failure in opensource full integration tests is due to the old mender-cli, we depend on the (unreleased) one.

PR #1317, opened 2021-04-12
compose: fix useradm verification

mchalski:
https://tracker.mender.io/browse/MEN-4623

useradm's /verify is not called for its own mgmt endpoints.
useradmLogin 'hijacks' all mgmt requests. that router doesn't enforce verification
of course; we intended to use the 'useradm' router instead.

the reason is malformed regex paths. traefik wants regex expressions
formatted in a very specific way:

... to use regular expressions with Host and Path expressions, you
must declare an arbitrarily named variable followed by the
colon-separated regular expression, all enclosed in curly braces. Any
pattern supported by Go's regexp package may be used (example:
`/posts/{id:[0-9]+}`).

iow the var name is mandatory. without it, the regex is not even parsed,
and the captured path segment can be anything.

that's why the useradmLogin router path means in fact:
`/api/management/{something}/useradm/{something}`

so it happily captures `/api/management/v1/useradm/users`, etc.
our regexes our understood as just var names/path segments.

changelog: fix auth verification on useradm APIs

---

for reference: 
https://doc.traefik.io/traefik/routing/routers/ (Regexp Syntax section)

### DIFF
--- a/docker-compose.auditlogs.yml
+++ b/docker-compose.auditlogs.yml
@@ -18,7 +18,7 @@ services:
             - traefik.enable=true
             - traefik.http.routers.auditlogs.entrypoints=https
             - traefik.http.routers.auditlogs.middlewares=userauth,sec-headers,compression,json-error-responder1,json-error-responder4
-            - traefik.http.routers.auditlogs.rule=PathPrefix(`/api/management/{ver:(v[0-9]+)}/auditlogs`)
+            - traefik.http.routers.auditlogs.rule=PathPrefix(`/api/management/{ver:v[0-9]+}/auditlogs`)
             - traefik.http.routers.auditlogs.tls=true
             - traefik.http.routers.auditlogs.service=auditlogs
             - traefik.http.services.auditlogs.loadbalancer.server.port=8080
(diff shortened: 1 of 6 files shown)


mchalski:

mchalski:

alfrunes (approved):

docker-compose.auditlogs.yml
```diff
@@ -18,7 +18,7 @@ services:
             - traefik.enable=true
             - traefik.http.routers.auditlogs.entrypoints=https
             - traefik.http.routers.auditlogs.middlewares=userauth,sec-headers,compression,json-error-responder1,json-error-responder4
-            - traefik.http.routers.auditlogs.rule=PathPrefix(`/api/management/{(v[0-9]+)}/auditlogs`)
+            - traefik.http.routers.auditlogs.rule=PathPrefix(`/api/management/{ver:(v[0-9]+)}/auditlogs`)
```
tranchitella, line 21: Can we get rid of the parenthesis?
```suggestion
            - traefik.http.routers.auditlogs.rule=PathPrefix(`/api/management/{ver:v[0-9]+}/auditlogs`)
```

mchalski:

mchalski:

mchalski:
a followup PR is in useradm-enterprise: https://github.com/mendersoftware/useradm-enterprise/pull/270
but it's not strictly necessary - consider it cleanup so no need to rush it.

tranchitella (approved):

PR #1321, opened 2021-04-19
Skip test_contact_support in staging

tranchitella:
Changelog: none

### DIFF
--- a/backend-tests/tests/test_tenantadm_support.py
+++ b/backend-tests/tests/test_tenantadm_support.py
@@ -34,9 +34,6 @@ api_useradm = ApiClient(useradm.URL_MGMT)
 
 
 class TestContactSupportEnterprise:
-    @pytest.mark.skipif(
-        isK8S(), reason="not testable in a staging or production environment"
-    )
     def test_contact_support_bad_request(self, clean_mongo):
         uuidv4 = str(uuid.uuid4())
         tenant, email, password = (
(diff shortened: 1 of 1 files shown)


backend-tests/tests/test_tenantadm_support.py
```diff
@@ -34,9 +34,6 @@
 
 
 class TestContactSupportEnterprise:
-    @pytest.mark.skipif(
-        isK8S(), reason="not testable in a staging or production environment"
-    )
     def test_contact_support_bad_request(self, clean_mongo):
```
lluiscampos, line 37: Shouldn't we skip also this one?

lluiscampos (approved):

backend-tests/tests/test_tenantadm_support.py
tranchitella: @lluiscampos as it returns "bad request", no email is sent, thus the test can run. Am I wrong?
The reason why we cannot run this test is because the staging env will send real emails, and we won't catch them from the mocked SMTP server.

tranchitella:
Manually running the staging tests:
https://gitlab.com/Northern.tech/Mender/integration/-/pipelines/288471997

backend-tests/tests/test_tenantadm_support.py
lluiscampos: Yeah, I was only wondering if it had value at all. But with your reply I understand it does: it will verify the `tenantadm` logic :+1:

mchalski (approved):

PR #1323, opened 2021-04-19
Fix statistics generator

oleorhagen:
* Do not process empty CHANGES
* Fix simple weirs bash quoting

### DIFF
--- a/extra/statistics-generator
+++ b/extra/statistics-generator
@@ -57,7 +57,7 @@ if [ -z "$BASE_DIR" -a "$REPO_ONLY" != 1 ]; then
 fi
 
 collect_changes() {
-    RELEASE_TOOL="$(dirname "$0")/release_tool.py"
+    RELEASE_TOOL="$(dirname $0)/release_tool.py"
     if [ "$REPO_ONLY" = 1 ]; then
         REPO_LIST=.
     else
--- a/extra/statistics-generator
+++ b/extra/statistics-generator
@@ -78,6 +78,9 @@ collect_changes() {
             pushd .
         else
             CHANGES="$("$RELEASE_TOOL" --version-of $repo --in-integration-version "$given_range")"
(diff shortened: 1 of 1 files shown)


lluiscampos:
While I the fixes themselves look good, this smells like you were running the tool from master while generating statistics for an old release which might have undesirable side effects. Am I wrong?

This fix prevents one side effect (getting stats from an non existing repo) but you could also get the opposite case (missing a repo!). Can you please double-check that generating the stats from their respective branches produce the same numbers?

extra/statistics-generator
```diff
@@ -78,6 +78,9 @@ collect_changes() {
             pushd .
         else
             CHANGES="$("$RELEASE_TOOL" --version-of $repo --in-integration-version "$given_range")"
+            if [[ "${CHANGES}" = "" ]]; then
+                continue
+            fi
```
kacf, line 83: How can this happen? This seems more like a user error or bug, which I'd want to know about.

extra/statistics-generator
```diff
@@ -57,7 +57,7 @@ if [ -z "$BASE_DIR" -a "$REPO_ONLY" != 1 ]; then
 fi
 
 collect_changes() {
-    RELEASE_TOOL="$(dirname "$0")/release_tool.py"
+    RELEASE_TOOL="$(dirname $0)/release_tool.py"
```
kacf, line 60: How do you figure? The old one is correct, AFAIK.

extra/statistics-generator
lluiscampos: That is also my suspicion :wink:

oleorhagen: Like @lluiscampos figured out, it was because I was using master for `2.4`

extra/statistics-generator
oleorhagen: Why do you unquote the `$0` argument?

PR #1329, opened 2021-04-22
[test_state_scripts] Support builds with systemctl service "mender"

lluiscampos:
By falling back to `systemctl restart mender` when the default service
name does not exist. This is the case for meta-mender/warrior.


lluiscampos:
To be tested together with https://github.com/mendersoftware/meta-mender/pull/1349

kacf (approved):
You're right, I did smile! :smile: 

I assume you targeted 2.7.x because that's what you're testing, but don't forget to cherry-pick to master.

PR #1338, opened 2021-04-28
QA-284: Add a Python requirements file for the release tool

oleorhagen:
Changelog: None


lluiscampos (changes requested):
We are missing to add this dir into the [dependabot configuration](https://github.com/mendersoftware/integration/blob/master/.github/dependabot.yml).

.github/dependabot.yml
```diff
@@ -18,3 +18,9 @@ updates:
     package-ecosystem: docker
     schedule:
       interval: weekly
+  - commit-message:
+      prefix: "Changelog:All"
+    directory: /extra/requirements.txt
```
lluiscampos, line 23: Is this correct? It is not a directory...

lluiscampos (approved):

.github/dependabot.yml
oleorhagen: Wow, quick! :p
Fixed

PR #1339, opened 2021-04-28
[release tools] MEN-4513: Changelog and statistics for Mender server

lluiscampos:
From now on we need to group the "Mender server" for changelog and
statistics, leaving everything else as independent repos.

`release_tool` supports now a new flag `--only-backend` that, when used
in combination with `--list` skips the independent components. The other
tools `changelog-generator` and `statistics-generator` are updated to
use this tool (note that they need to add `integration` repo).

Added unit tests to verify backwards compatibility.

### DIFF
--- a/extra/changelog-generator/changelog-generator
+++ b/extra/changelog-generator/changelog-generator
@@ -114,12 +114,17 @@ else:
 if args.all:
     repo_list = (
         subprocess.check_output(
-            [os.path.join(base_dir, "integration/extra/release_tool.py"), "--list"]
+            [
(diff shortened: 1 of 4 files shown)


oleorhagen (approved):
I'm not too familiar with this code, but on the surface looks good :)

lluiscampos:
> I'm not too familiar with this code, but on the surface looks good :)

I created the tests before I modified the code to make sure I wouldn't mess up with the output of that delicate function.

I'll merge as is. If @kacf spots and error later I can quickly amend.

PR #1345, opened 2021-05-04
[requirements.txt]: Downgrade idna to < 3.0

oleorhagen:
This is needed, since the requests library does not yet support `idna > 2.x`.

See this for information:

https://github.com/psf/requests/issues/5710

This shows up in Alpine containers during install as:

```
ERROR: Cannot install -r tests/integration/mender_integration/tests/requirements/python-requirements.txt (line 54) and idna==3.1 because these package versions have conflicting dependencies.
The conflict is caused by:
    The user requested idna==3.1
    requests 2.25.1 depends on idna<3 and >=2.5
```

But not in containers with these dependencies already installed.

Like this is from a `mender-qa` pipeline from a container running on the
`docker` image:

```
Requirement already satisfied: idna==3.1 in /usr/lib/python3.8/site-packages (from -r /builds/Northern.tech/Mender/integration/tests/requirements/python-requirements.txt (line 23)) (3.1)
Requirement already satisfied: requests==2.25.1 in /usr/lib/python3.8/site-packages (from -r /builds/Northern.tech/Mender/integration/tests/requirements/python-requirements.txt (line 54)) (2.25.1)
```

This is due to the installation in the `requirements/apk-requirements.txt` file,
already installing this, in some cases:

```
[32;1m$ apk add $(cat ${WORKSPACE}/integration/tests/requirements/apk-requirements.txt)[0;m
(1/73) Installing readline (8.1.0-r0)
  .
  .
  .
(55/73) Installing py3-idna (3.1-r0)
```

Never the less, the requirements file is still wrong.

Changelog: None
Signed-off-by: Ole Petter <ole.orhagen@northern.tech>

PR #1346, opened 2021-05-04
[tests requirements] Lock idna version to 2.10

lluiscampos:
As `requests` require idna<3.0 which cannot be satisfied with the
current requirements.txt list.

The issue seems to be side tracked from our CI due to some previously
installed packages (I honestly don't know exactly what).

Introduced by mistake at 099539ab6.


lluiscampos:

lluiscampos:
Already submitted at #1345

PR #1361, opened 2021-05-11
Traefik file cleanups

alfrunes:
I missed the `devicemonitor` configuration in #1286. I also did some minor cleanup suggested in that PR.

### DIFF
--- a/config/traefik/traefik.monitor.yml
+++ b/config/traefik/traefik.monitor.yml
@@ -1,9 +1,9 @@
 http:
   routers:
     #
-    # devicecmonitor
+    # devicemonitor
     #
-    devicecmonitor:
+    devicemonitor:
       entrypoints: https
       middlewares: #{{- block "circuit-breaker" . -}}{{- if not (env "TESTING") }}
       - circuit-breaker #{{- end}}{{end}}
@@ -15,10 +15,10 @@ http:
       - json-error-responder3
       - json-error-responder4
       rule: "PathPrefix(`/api/devices/{version:v[0-9]+}/devicemonitor`)"
-      service: devicecmonitor
+      service: devicemonitor
       tls: true
 
-    devicecmonitorMgmt:
+    devicemonitorMgmt:
       entrypoints: https
       middlewares: #{{- template "circuit-breaker"}}
       - userauth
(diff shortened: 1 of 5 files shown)


tranchitella (approved):

docker-compose.demo.yml
```diff
@@ -76,9 +76,6 @@ services:
         environment:
             # enable demo mode for UI ["true"/"false"]
             DEMO: "true"
-        labels:
-            - traefik.http.middlewares.sec-headers.headers.stsSeconds=0
-            - traefik.http.middlewares.sec-headers.headers.isDevelopment=true
```
mzedel, line 81: Although not critical, it would be nice to keep these to not have to clear the HSTS browser settings - I would guess this would mean an extra middleware in the `traefik.demo.yaml` and adding it to the gui router in there too...

config/traefik/traefik.monitor.yml
```diff
@@ -0,0 +1,39 @@
...
+      - json-error-responder4
+      rule: "PathPrefix(`/api/management/{version:v[0-9]+}/devicemonitor`)"
+      service: devicecmonitor
+      tls: true
+
+  services:
+    devicecmonitor:
+      loadBalancer:
+        servers:
+        - url: "http://mender-devicecmonitor:8080"
```
mzedel, line 39:
```suggestion
        - url: "http://mender-devicemonitor:8080"
```

mzedel (changes requested):
Thanks for the clean- & follow up @alfrunes!
There might be a reason that I'm not aware of, but `devicecmonitor` seems a bit confusing to me (maybe it's for monitoring device c - or it's written in C?)...
Apart from that, just an optional request for the STS functionality...

merlin-northern:
no other comments I have, looks good.

config/traefik/traefik.monitor.yml
alfrunes: oops, bad regex/replace string...

docker-compose.demo.yml
alfrunes: Will do, I just assumed they were already present in the file configuration and that I forgot to clear them previously.

alfrunes:

PR #1363, opened 2021-05-12
README: update backend diagram

kjaskiewiczz:
I was originally thinking about replacing the diagram with the link to the docs, but the relation between docs and integration is, let's say, complicated (maybe the docs should have a link to this diagram, since we are referencing to the integration repo there?).
There is ongoing discussion about the server installation docs. We have tasks to restructure this part of the documentation, so I decided that having the diagram updated is enough for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,27 +44,37 @@ A provided `docker-compose.yml` file will provision the following set of
 services:
 
 ```
-       |
-  port |        +-----------------------+         +------------------------+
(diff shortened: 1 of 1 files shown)


tranchitella (approved):
Thanks!

tranchitella:
@kjaskiewiczz no need to run the pipeline, I'm merging as it is.

PR #1368, opened 2021-05-14
backend tests: fix 2fa test

kjaskiewiczz:
tests has to be updated because settings obtained with GET /settings endpoint does not include 2fa status anymore;
the 2fa status comes with the user object (GET /user/me)


kjaskiewiczz:

tranchitella (approved):
Thanks!

PR #1381, opened 2021-05-20
[test_release_tool] Fix tests for staging branch

lluiscampos:
Adding a new fixture that can deduce if we are running in staging or
not, and then using it to add some exceptions and quirks on the test
logic.

### DIFF
--- a/extra/test_release_tool.py
+++ b/extra/test_release_tool.py
@@ -13,21 +13,38 @@
 #    limitations under the License.
 
 import os
-import sys
-import shutil
-import re
 import pathlib
+import re
+import shutil
+import sys
 from unittest.mock import patch
 
 import pytest
-
-from release_tool import main
-from release_tool import docker_compose_files_list
-from release_tool import Component
+import yaml
+from release_tool import Component, docker_compose_files_list, main
 
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
 RELEASE_TOOL = os.path.join(THIS_DIR, "release_tool.py")
 INTEGRATION_DIR = os.path.normpath(os.path.join(THIS_DIR, ".."))
+
+
+@pytest.fixture(scope="session")
+def is_staging():
+    """Fixture to figure out if we are running the tests in staging branch
+
+    Inspect git-versions.yml and ensure that the core OS repositories that have
+    enterprise forks (i.e. deployments, inventory, and useradm) do not exist; as
+    they shall exist only in git-versions-enterprise.yml.
+    """
+
+    non_staging_components = [
+        "mender-deployments",
+        "mender-inventory",
+        "mender-useradm",
+    ]
+    with open(os.path.join(INTEGRATION_DIR, "git-versions.yml")) as fd:
+        content = yaml.safe_load(fd)
+        return not any(c in non_staging_components for c in content["services"])
 
 
 @pytest.fixture(scope="function", autouse=True)
@@ -238,7 +255,7 @@ def test_version_of_with_in_integration_version(capsys):
     )
 
 
-def test_set_version_of(capsys):
+def test_set_version_of(capsys, is_staging):
     # Using --set-version-of modifies both versions, regardless of using the repo name
     run_main_assert_result(
         capsys, ["--set-version-of", "deviceauth", "--version", "1.2.3-test"]
@@ -267,8 +284,7 @@ def test_set_version_of(capsys):
         "4.5.6-test",
     )
     # NOTE: skip check for OS flavor for branches without it (namely staging)
-    list_repos = run_main_assert_result(capsys, ["--list", "git"], None)
-    if "deployments" in list_repos.split("\n"):
+    if not is_staging:
         run_main_assert_result(capsys, ["--version-of", "deployments"], "4.5.6-test")
         run_main_assert_result(
             capsys,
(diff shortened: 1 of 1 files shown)


extra/test_release_tool.py
```diff
@@ -419,32 +436,38 @@ def test_get_components_of_type(integration_dir_func):
...
     # should only include deprecated repos
     assert "deviceadm" in repos_name
     assert "mender-api-gateway-docker" in repos_name
     assert "mender-conductor" in repos_name
     assert "mender-conductor-enterprise" in repos_name
+    if is_staging:
+        assert "deployments" in repos_name
+        assert "inventory" in repos_name
+        assert "useradm" in repos_name
+        assert "workflows" in repos_name
```
mzedel, line 470: is this right? I think the location is rather misleading...

extra/test_release_tool.py
```diff
@@ -419,32 +436,38 @@ def test_get_components_of_type(integration_dir_func):
...
     assert "auditlogs" not in repos_name
     assert "mtls-ambassador" not in repos_name
     assert "deviceconnect" not in repos_name
     assert "mender-connect" not in repos_name
     assert "deviceconfig" not in repos_name
+    if not is_staging:
+        assert "deployments" not in repos_name
+        assert "inventory" not in repos_name
+        assert "useradm" not in repos_name
+        assert "workflows" not in repos_name
```
kacf, line 460: This approach does not seem very maintainable. Isn't it better to deduce this from looking at the "-enterprise" suffix? It will only happen to services that have both OS and ENT components, right?

extra/test_release_tool.py
```diff
@@ -13,23 +13,40 @@
...
+
+    Inspect git-versions.yml and ensure that the core OS repositories that have
+    enterprise forks (i.e. deployments, inventory, and useradm) do not exist; as
+    they shall exist only in git-versions-enterprise.yml.
+    """
+
+    non_staging_components = [
+        "mender-deployments",
+        "mender-inventory",
+        "mender-useradm",
```
kacf, line 43: Shouldn't workflows also be here?

extra/test_release_tool.py
lluiscampos: > It will only happen to services that have both OS and ENT components, right?

Correct



> This approach does not seem very maintainable.

The whole test is not very maintainable, really. Adding a new repo will still make the this test pass, but won't really verify that the repo is there... (it will however break the next test, `test_list_repos`).

Maybe I should just remove these test? I added them when I implementing the `--only-backend` flag to make sure I didn't break existing functionality.

extra/test_release_tool.py
lluiscampos: It _could_ be there. But there are two `workflows` Docker images (server and worker), so for simplicity I made the fixture to only look for the core ones. It doesn't really matter, it could look for only one of them...

extra/test_release_tool.py
lluiscampos: It is "right" from the `release_tool` point of view: these repos are marked as non-releasable for staging branch (iow we don't tag them when making a HM release)

mzedel: just since it is right underneath the `# should only include deprecated repos` comment - 
```suggestion
    if not is_staging:
        assert "deployments" not in repos_name
        assert "inventory" not in repos_name
        assert "useradm" not in repos_name
        assert "workflows" not in repos_name
    else:
        assert "deployments" in repos_name
        assert "inventory" in repos_name
        assert "useradm" in repos_name
        assert "workflows" in repos_name
    # should only include deprecated repos
    assert "deviceadm" in repos_name
    assert "mender-api-gateway-docker" in repos_name
    assert "mender-conductor" in repos_name
    assert "mender-conductor-enterprise" in repos_name
```

lluiscampos: Okey, but my point was that deployments, inventory, useradm and workflows are "deprectated" (*) in `staging` branch. I think they should stay together with "deviceadm", "mender-api-gateway-docker", etc, because they are treated the same way.

(*) Instead of "deprecated" I can write "non-releasable repos" or "repos not included in a release", if that will help.

mzedel: but you have the "not include" almost right above in [line 456](https://github.com/mendersoftware/integration/pull/1381/files/d42229c70af12a9c8d153980550a092ebd549f00#diff-a4f8e500e8139d190a4257978c6b52b60639d5423572a8d97fad2d03861bf9ecR456), apart from that I think adding a comment for these would clarify things, since they are very much _not_ deprecated + the entire file looks rather painful to maintain already...

extra/test_release_tool.py
kacf: Right, ok, it doesn't really matter as long as there is at least one positive match. Never mind then.

extra/test_release_tool.py
kacf: Hmm yes, tricky... Maybe we can just test some known repositories, without testing the whole list?

I don't really have a strong preference, you're right that it already a bit unmaintainable, but I don't see another way to test this without hardcoding.

lluiscampos: Okey: I'll rewrite to verify _some_ repos, making sure I catch the different "types" we have. Then is obvious we are not testing all but still verifying the gist of the functionality.

extra/test_release_tool.py
lluiscampos: Related to @kacf comment above, I will simplify the tests.

lluiscampos:
@kacf @mzedel I introduce these samples of the different "types" of repositories and simplify the tests. I think the test code is readable enough now :)

extra/test_release_tool.py
```diff
@@ -13,22 +13,46 @@
...
-from release_tool import Component
+import yaml
+from release_tool import Component, docker_compose_files_list, main
 
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
 RELEASE_TOOL = os.path.join(THIS_DIR, "release_tool.py")
 INTEGRATION_DIR = os.path.normpath(os.path.join(THIS_DIR, ".."))
 
+# Samples of the different "types" of repos for listing tests
+SAMPLE_REPOS_BACKEND_BASE = ["deviceauth", "gui", "deviceauth", "tenantadm"]
```
mzedel, line 31:
```suggestion
SAMPLE_REPOS_BACKEND_BASE = ["deviceauth", "gui", "tenantadm"]
```

mzedel (commented):
that looks so much better! the one thing I noted _should_ still work with the double deviceauth, but you could remove it just to be sure...

extra/test_release_tool.py
lluiscampos: Oh. This probably means that I had some other repo in mind... :thinking:

mzedel: `workflows` maybe?

mzedel (approved):

PR #1392, opened 2021-05-27
[staging] backend-tests: produce report and result when skip

lluiscampos:
This is a rather hacky approach, but the goal is to let pytest run as
expected, so that it produces the report and results files for CI to
collect. Otherwise CI fails and does not report back to the PR.

### DIFF
--- a/backend-tests/run
+++ b/backend-tests/run
@@ -209,7 +209,12 @@ for suite in "${TEST_SUITES[@]}"; do
     case "$suite" in
         open)
             echo "Skipping Open Source test suite for staging testing"
-            continue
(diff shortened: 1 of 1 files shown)


lluiscampos:
@mender-test-bot start pipeline amigo

lluiscampos:
> @mender-test-bot start pipeline amigo

:womans_hat:

lluiscampos:
`pytest` returns 5 when no tests are collected, see [documentation here](https://docs.pytest.org/en/6.2.x/usage.html#possible-exit-codes). So this won't work.

PR #1403, opened 2021-06-04
[tests] Add client version fixture and mark File Transfer limits tests

lluiscampos:
The fixture uses release_tool to figure out which mender client is
being used. Tests marked with min_version higher that the one being used
will be skip.

This fixture is specifically designed for staging branch, for which we
use the "latest stable" mender client and some tests might be at times
incompatible. For the rest of the branches it won't really have any
effect, as the client version will be a branch and validate as more
recent that the required minimum version.

Mark File Transfer limits tests to be skip on current client version.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,19 +12,25 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+import distutils.spawn
 import logging
-import requests
-import filelock
-import uuid
-import subprocess
 import os
 import re
+import subprocess
+import uuid
+from distutils.version import LooseVersion
+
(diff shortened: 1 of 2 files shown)


lluiscampos:
> @lluiscampos, Let me know if you want to start the integration pipeline by mentioning me and the command "start pipeline".

lluiscampos:
@kacf @tranchitella I know that nobody has asked for this, but I just don't like having PRs against integration/staging red by default...

tranchitella (approved):
Looks good, thanks a lot!

kacf (approved):
:+1:

PR #1414, opened 2021-06-11
QA-316: release_tool: Add release related hotkeys to build menu.

kacf:
Add hotkeys to enable/disable tests, platforms and release publishing,
as well as a way to reset options to their default values.

Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -1569,18 +1569,30 @@ def get_extra_buildparams_from_yaml():
 
 
 def trigger_build(state, tag_avail):
-    extra_buildparams = get_extra_buildparams()
-
-    for param in extra_buildparams.keys():
-        if state_value(state, ["extra_buildparams", param]) is None:
-            update_state(
-                state, ["extra_buildparams", param], extra_buildparams[param].value
-            )
-
+    extra_buildparams = None
     params = None
+
+    def set_param(name, value):
+        params[name] = value
+        if extra_buildparams.get(name) is not None:
+            # Extra build parameters, that are not part of the build tags for
+            # each repository, should be saved persistently in the state file so
+            # that they can be repeated in subsequent builds.
+            update_state(state, ["extra_buildparams", name], params[name])
 
     # Allow changing of build parameters.
     while True:
+        if extra_buildparams is None:
+            extra_buildparams = get_extra_buildparams()
+
+            for param in extra_buildparams.keys():
+                if state_value(state, ["extra_buildparams", param]) is None:
+                    update_state(
+                        state,
+                        ["extra_buildparams", param],
+                        extra_buildparams[param].value,
+                    )
+
         if params is None:
             # We'll be adding parameters here that shouldn't be in 'state', so make a
             # copy.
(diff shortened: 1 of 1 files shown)


kacf:
The hard coding isn't super nice, but I couldn't think of a better way. I also considered adding tests, but it doesn't seem very sensible to test this, because there are so many things that can go wrong which a test won't catch. Ultimately it depends on the build system. I'm open to suggestions though.

kacf:

extra/release_tool.py
```diff
@@ -1576,12 +1588,47 @@ def trigger_build(state, tag_avail):
...
+Y = Yes
+N = No (Quit)
+E = Open in editor
+R = Reset all to default
+ET = Enable all tests
+DT = Disable all tests
+EP = Enable all platforms
+DP = Disable all platforms
+ER = Enable publishing of the release
+DR = Disable publishing of the release
```
lluiscampos, line 1608: I would be explicit here adding "automatic" somewhere. The publishing of the release is always enabled, just manually done.
```suggestion
ER = Enable automatic publishing of the release
DR = Disable automatic publishing of the release
```

extra/release_tool.py
```diff
@@ -1576,12 +1588,47 @@ def trigger_build(state, tag_avail):
...
+                trigger_jenkins_build(params, extra_buildparams)
+            return
 
         reply = ask(
-            "Do you want to change any of the parameters (Y/N/open in Editor)? "
+            """Do you want to change any of the parameters?
+Y = Yes
+N = No (Quit)
+E = Open in editor
+R = Reset all to default
```
lluiscampos, line 1602: This is a great addition, thanks!

I used to do something like `rm -f ~/.release-state.yaml && release_tool -b ...`

extra/release_tool.py
```diff
@@ -1592,39 +1639,47 @@ def trigger_build(state, tag_avail):
...
+                    if action == "true" and params.get(build_param) is not None:
+                        set_param(build_param, action)
+
+        elif reply.upper() == "EP" or reply.upper() == "DP":
+            if reply.upper().startswith("E"):
+                action = "true"
+            else:
+                action = "false"
+            set_param("BUILD_CLIENT", action)
+            set_param("BUILD_SERVERS", action)
```
lluiscampos, line 1669: I don't think that "disable platforms" should also disable `BUILD_CLIENT` and `BUILD_SERVERS`. Which is the use case?

I would say that "disable platforms" refers to "disable build on everything not strictly needed for a release", leaving `BUILD_CLIENT` and `BUILD_SERVERS`, which are required for publishing the release. In other words, this code here could look something like:

```suggestion
        elif reply.upper() == "EP" or reply.upper() == "DP":
            if reply.upper().startswith("E"):
                action = "true"
                set_param("BUILD_CLIENT", action)
                set_param("BUILD_SERVERS", action)
            else:
                action = "false"
```

Thinking about it from another angle, this way would complement the "enable tests" case above, where "disable tests" does not touch BUILD_CLIENT, BUILD_SERVERS

lluiscampos (changes requested):
One thing that requires discussion/rework.

But overall looks good.

I also don't think we need to test this. In my vision, the main focus on the tests should be on release_tool options and flows used automatically by other tools or by CI piepelines, where regressions are expensive. This is a manual option so we will catch errors and fix accordingly, if any.

extra/release_tool.py
kacf: > I don't think that "disable platforms" should also disable `BUILD_CLIENT` and `BUILD_SERVERS`. Which is the use case?

I guess I'm looking at it more from the test perspective. It's quite a common use case for me. If I need to debug something on a runner: Disable all platforms and then enable the one I want. In this scenario, `BUILD_CLIENT` and `BUILD_SERVERS` are unneeded.

But you are right that from a release perspective it doesn't make sense. How about renaming it to "build combinations" instead of "platforms"? And then when enabling the release publishing, or all tests, it will turn those two back on again?

kacf: Resolved according to discussion in Slack.

kacf:
@lluiscampos: Updated!

kacf:

lluiscampos (approved):
I _think_ it is correct :rocket: 

Thanks for the good discussion. Looking forward to use this feature :)

kacf:

kacf:
> I _think_ it is correct rocket

I also _think_ it is correct! :smile: I tested back and forth quite a lot, but it's easy to go blind, and it's also difficult to judge what feels intuitive and what doesn't.

PR #1422, opened 2021-06-17
2.7.x: QA-316: release_tool: Add release related hotkeys to build menu.

kacf:
Add hotkeys to enable/disable tests, platforms and release publishing,
as well as a way to reset options to their default values.

Changelog: None

(cherry picked from commit 651b4c65c76f4da44d1be24bf00a3f3846ecd135)


kacf:
I'll merge without mender-qa pipeline. It passed on master, and all of this is inside a function which is only used when triggering builds.

PR #1424, opened 2021-06-20
Regenerate crypto materials for mTLS testing using a script

tranchitella:
Changelog: none


tranchitella:

tranchitella:

tranchitella:

extra/mtls/certs/README.md
```diff
@@ -1,3 +1,24 @@
...
-- `client` - example client certs, using different algorithms, signed by `tenant-ca`
+# Example certificates for the mTLS setup
+
+This folder contains the example certificates for the mTLS setup:
+
+* `/server` - example self-signed server cert
+* `tenant-ca` - example self-signed tenant certificate authority ( key pass: root)
+* `client` - example client certs, using different algorithms, signed by `tenant-ca`
+
+You can create your own certificates following the [Mutual TLS documentation](https://docs.mender.io/2.7/server-integration/mutual-tls-authentication).
```
lluiscampos, line 9: This link to documentation is tricky. It cannot point to hosted (we don't have this doc there) but pointing it to 2.7 will make it deprecated soon. I don't have a solution... :thinking:

extra/mtls/certs/client/cert.conf
```diff
@@ -0,0 +1,12 @@
+[req]
```
lluiscampos, line 1: File committed by mistake? Is it also generated in the script

extra/mtls/certs/tenant-ca/tenant.ca.srl
```diff
@@ -0,0 +1 @@
+0D48525F87BC581B1DD7ECB336E53F2A8C73ADA0
```
lluiscampos, line 1: What are these `srl` finger prints?

lluiscampos (approved):
I am not the most openssl knowledgeable from the team, but given that we had no documentation on how to generate these previously... they look good to me :)

Thank you very much for the extra documentation/script effort.

extra/mtls/certs/client/cert.conf
tranchitella: Actually, I wanted it to be in thee commit to have the right configuration used to generate the certs together with the certs. Is it a bad idea?

extra/mtls/certs/tenant-ca/tenant.ca.srl
tranchitella: Registry of signatures from the CA

extra/mtls/certs/README.md
tranchitella: Same here, no idea on how to fix it.

extra/mtls/certs/client/cert.conf
lluiscampos: I don't know. This file is overridden three times by the script, so why save only one of them? And why not to save also the server conf file?

These kind of questions makes me thing that it is best not to have it committed.

tranchitella: ok, let's remove them!

tranchitella:
The pipeline was green, I force-pushed to remove openssl x509 req config files which are not used anywhere.
I won't run again the pipeline, merging it as it is.

PR #1429, opened 2021-06-23
release_tool.py: Misc fixes

lluiscampos:


extra/release_tool.py
```diff
@@ -670,7 +670,10 @@ def do_list_repos(args, optional_too, only_backend):
     repos_versions_dict = {}
     for repo in repos:
         if repo.name == "integration":
-            repos_versions_dict[repo.name] = args.in_integration_version
+            if args.in_integration_version:
+                repos_versions_dict[repo.name] = args.in_integration_version
+            else:
+                repos_versions_dict[repo.name] = "HEAD"
```
kacf, line 676: Actually, I think a better approach here is to remove the `if repo.name == "integration"` entirely, and let "integration" follow the path into `version_of` which already deals with this in a smarter way. The algorithm is quite slow, but it figures out the real branch identity of HEAD instead of just printing it outright.

I just tried it locally, it works.

extra/release_tool.py
```diff
@@ -675,12 +675,16 @@ def do_list_repos(args, optional_too, only_backend):
...
+            try:
+                repos_versions_dict[repo.name] = version_of(
+                    integration_dir(),
+                    repo.yml_components()[0],
+                    args.in_integration_version,
+                    git_version=(args.list == "git"),
+                )
+            except KeyError:
+                # This repo doesn't exist in the given integration version
+                repos_versions_dict[repo.name] = "UNRELEASED"
```
kacf, line 687: :+1:

kacf (changes requested):

extra/release_tool.py
lluiscampos: Ok, thanks.

I'll try that a bit more in with different combinations because originally it did break, and then I added the special case handling fro integration.

lluiscampos:
Reworked

kacf (approved):

PR #1430, opened 2021-06-23
Some smaller fixes for feature-update_control_maps branch.

kacf:


tranchitella (approved):

lluiscampos (approved):
Nice fix on the test, I bet we can use this logic in more places when we find flaky tests.

PR #1431, opened 2021-06-24
Fix spurious failure in test_rootfs_image_rejected (and maybe others).

kacf:
The test fails if the client is too quick, because it looks for the
"inprogress" deployment status. However, by the time it gets to do
that check, the client may already be past that and the deployment is
"finished".

What we really want to look for is whether the deployment has left the
"pending" status, which is before any clients have picked it up.

Changelog: None

### DIFF
--- a/tests/MenderAPI/deployments.py
+++ b/tests/MenderAPI/deployments.py
@@ -167,11 +167,34 @@ class Deployments:
                 time.sleep(polling_frequency)
                 continue
 
-        if time.time() > timeout:
-            pytest.fail(
-                "Never found status: %s for %s after %d seconds"
-                % (expected_status, deployment_id, max_wait)
-            )
+        pytest.fail(
+            "Never found status: %s for %s after %d seconds"
+            % (expected_status, deployment_id, max_wait)
+        )
+
+    def check_not_in_status(
+        self, expected_status, deployment_id, max_wait=60 * 60, polling_frequency=0.2
+    ):
+        timeout = time.time() + max_wait
+
+        while time.time() <= timeout:
+            data = self.get_status(status=expected_status)
+
+            for deployment in data:
+                if deployment["id"] == deployment_id:
+                    time.sleep(polling_frequency)
+                    continue
+            else:
+                logger.info(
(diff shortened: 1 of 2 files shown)


tests/MenderAPI/deployments.py
```diff
@@ -173,6 +173,31 @@ def check_expected_status(
...
+                    time.sleep(polling_frequency)
+                    continue
+            else:
+                logger.info(
+                    "left deployment status (%s) as expected for: %s"
+                    % (expected_status, deployment_id)
+                )
+                return
+
+        if time.time() > timeout:
```
tranchitella, line 195: Do you really need the if here?
I think you can only reach this point if `time.time() > timeout`.

kacf: It is preexisting, actually, I copied this from the inverse function. But you're right, I believe I can remove it from both.

lluiscampos (approved):

tests/MenderAPI/deployments.py
lluiscampos: I agree with this observation, the `if` condition could be removed from both functions.

kacf: @tranchitella, @lluiscampos: Fixed!

tests/MenderAPI/deployments.py
```diff
@@ -167,11 +167,34 @@ def check_expected_status(
...
+                    continue
+            else:
+                logger.info(
+                    "left deployment status (%s) as expected for: %s"
+                    % (expected_status, deployment_id)
+                )
+                return
+
+        pytest.fail(
+            "Never found status: %s for %s after %d seconds"
```
lluiscampos, line 195: Nitpick: the log message here should be something like:

```suggestion
            "Never left status: %s for %s after %d seconds"
```

lluiscampos (approved):

tests/MenderAPI/deployments.py
kacf: Good point, fixed!

tranchitella (approved):

kacf:
Only known/spurious failures.

PR #1442, opened 2021-06-30
MEN-4790: fix ad improve the the production template

tranchitella:
Improve the comments to make it clear the deployments service must
contact the storage layer using the same DNS name the devices will use.
Set the correct value for the CREATE_ARTIFACT_GATEWAY_URL env variable.

We cannot use the domain name in the rule to route requests to minio,
because you can use the same domain name for both APIs/UI and minio. The
X-Amz-Date header is always present when performing AWS S3 requests.

### DIFF
--- a/docker-compose.storage.minio.yml
+++ b/docker-compose.storage.minio.yml
@@ -16,7 +16,7 @@ services:
         labels:
             - "traefik.enable=true"
             - "traefik.http.routers.minio.entrypoints=https"
-            - "traefik.http.routers.minio.rule=Host(`${STORAGE_URL:-s3.docker.mender.io}`)||Headers(`X-Forwarded-Host`,`${STORAGE_URL:-s3.docker.mender.io}`)||PathPrefix(`/mender-artifact-storage`)"
+            - "traefik.http.routers.minio.rule=PathPrefix(`/mender-artifact-storage`)||HeadersRegexp(`X-Amz-Date`, `.+`)"
             - "traefik.http.routers.minio.tls=true"
             - "traefik.http.services.minio.loadbalancer.server.port=9000"
             - mender.testprefix=${MENDER_TESTPREFIX:-""}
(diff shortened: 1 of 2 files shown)


tranchitella:
@lluiscampos what's up here with devicemonitor?

lluiscampos:
@tranchitella Your PR is not based on latest mendersoftware/2.7.x, but rather at an old point in history where devicemonitor did not yet exist in the eyes of the release_tool.

Please rebase and the error will go away.

tranchitella:
@lluiscampos thanks, and sorry, I didn't notice it.

tranchitella:

tranchitella:
Failures in mtls are expected

docker-compose.storage.minio.yml
```diff
@@ -16,7 +16,7 @@ services:
         labels:
             - "traefik.enable=true"
             - "traefik.http.routers.minio.entrypoints=https"
-            - "traefik.http.routers.minio.rule=Host(`${STORAGE_URL:-s3.docker.mender.io}`)||Headers(`X-Forwarded-Host`,`${STORAGE_URL:-s3.docker.mender.io}`)||PathPrefix(`/mender-artifact-storage`)"
+            - "traefik.http.routers.minio.rule=PathPrefix(`/mender-artifact-storage`)||HeadersRegexp(`X-Amz-Date`, `.+`)"
```
merlin-northern, line 19: just wodnering (I am not an expert here) but does logical or here mean that every request with X-Amz-Date will match? does && make sense?

merlin-northern:
after this is merged, we should update the two threads on Mender Hub.

docker-compose.storage.minio.yml
tranchitella: yes, requests with `X-Amz-Date` will match. OR is correct, because when you download the artifact with a presigned link, you do not provide the `X-Amz-Date` header.

merlin-northern: ok then thank you

merlin-northern (approved):

kjaskiewiczz (approved):

PR #1443, opened 2021-06-30
MEN-4790: rely on the X-Amz-Date header to route requests to minio

tranchitella:
We cannot use the domain name in the rule to route requests to minio,
because you can use the same domain name for both APIs/UI and minio. The
X-Amz-Date header is always present when performing AWS S3 requests.

Changelog: none


tranchitella:

merlin-northern (approved):
lgtm (not an expert on traefik)

alfrunes (approved):

lluiscampos:
Do we need this one in 3.0.0 release?

kjaskiewiczz (approved):

PR #1436, opened 2021-06-28
Fix: mtls tests, mount the CA certificate in the right location:

tranchitella:
Adapt the CA certificate mount point to the new location in the
mtls-ambassador updated in the commit
7c1429946266ef12113c9dc2d61e6b2f59599dfb.

Changelog: none

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>

PR #1445, opened 2021-07-01
Set the correct tenant CA certificate path for mtls integration tests

alfrunes:


alfrunes:
Started a pipeline with mendersoftware/mtls-ambassador#58 here: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/330119388

tranchitella (approved):

lluiscampos:
Why aren't we strictly reverting this? https://github.com/mendersoftware/integration/pull/1436/files

lluiscampos:
(the build error I think is my half-way branched out release - can be ignored)

lluiscampos:
> (the build error I think is my half-way branched out release - can be ignored)

It should be fine now, I re-started the job

PR #1446, opened 2021-07-01
release_tool: Update extra compose files used in integration tests

lluiscampos:
When using option I) to update branch names on docker-compose files,
update also the two files in "extra" that are using mender images.

This will prevent the current situation where integration/2.7.x was
testing mtls-ambassador/master. It has manually been fixed for 2.7.x,
2.6.x and 3.0.x.

Renamed the mtls test file to have the docker-compose prefix, to avoid
having yet another exception in the release_tool code.


lluiscampos:

tranchitella (approved):

oleorhagen (approved):

PR #1456, opened 2021-07-06
Test that client rejects an Update Control map which is broken.

kacf:
I forgot to submit this one earlier.

### DIFF
--- a/tests/tests/test_update_control.py
+++ b/tests/tests/test_update_control.py
@@ -127,3 +127,67 @@ class TestUpdateControlEnterprise:
         )
 
         deploy.check_expected_status("finished", deployment_id)
+        deploy.check_expected_statistics(deployment_id, "success", 1)
+
+    def test_update_control_with_broken_map(
+        self, enterprise_no_client, valid_image,
+    ):
+        """
+        Schedule an update with an invalid map, which should fail.
+        """
+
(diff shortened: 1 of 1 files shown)


kacf:

tests/tests/test_update_control.py
```diff
@@ -127,3 +127,67 @@ def test_update_control(
...
+        devauth.accept_devices(1)
+
+        deploy = Deployments(auth, devauth)
+
+        devices = list(
+            set([device["id"] for device in devauth.get_devices_status("accepted")])
+        )
+        assert 1 == len(devices)
+
+        wait_for_connect(auth, devices[0])
```
tranchitella, line 162: Out of curiosity, why do you need to wait for connect?

tranchitella (approved):

tests/tests/test_update_control.py
kacf: I don't know, actually. All this boiler plate code was copied from the previous test, which @oleorhagen wrote. I wouldn't be surprised if this was again copied from somewhere else that actually needs it. I'll merge this as is, but I can try removing them in a followup.

kacf: https://github.com/mendersoftware/integration/pull/1460

PR #1457, opened 2021-07-06
test_image_download_retry_timeout: Rework update start detection

lluiscampos:
Previous code relied on a single call to MenderDevice.run, with retries
    the command until it succeeds using an exponential back-off time between
    tries. This test is test time sensitive, it needs to break the
    connectivity after it starts the download but before it completes, and
    sometimes it is missed between different SSH connection attempts.

    The new code basically tries with a while loop in a single SSH
    connection, which should be much more accurate.

### DIFF
--- a/tests/tests/test_fault_tolerance.py
+++ b/tests/tests/test_fault_tolerance.py
@@ -224,7 +224,19 @@ class TestFaultTolerance(MenderTesting):
             if test_set["blockAfterStart"]:
                 # Block after we start the download.
                 deployment_id, new_yocto_id = common_update_procedure(valid_image)
-                mender_device.run("fuser -mv %s" % (inactive_part))
+                mender_device.run(
+                    "start=$(date -u +%s);"
+                    + 'output="";'
+                    + 'while [ -z "$output" ]; do '
+                    + "sleep 0.1;"
+                    + 'output="$(fuser -mv %s)";' % inactive_part
+                    + "now=$(date -u +%s);"
+                    + "if [ $(($now - $start)) -gt 600 ]; then "
+                    + "exit 1;"
+                    + "fi;"
+                    + "done",
+                    wait=10 * 60,
+                )
 
                 # storage must be blocked by ip to kill an ongoing connection
                 # so block the whole gateway
--- a/tests/tests/test_fault_tolerance.py
+++ b/tests/tests/test_fault_tolerance.py
@@ -116,7 +116,7 @@ class TestFaultTolerance(MenderTesting):
                 hide=True,
             )
             time.sleep(2)
-            if int(output) >= 2:  # check that some retries have occured
+            if int(output) >= 2:  # check that some retries have occurred
                 logger.info(
                     "Looks like the download was retried 2 times, restoring download functionality"
                 )
@@ -195,7 +195,7 @@ class TestFaultTolerance(MenderTesting):
     ):
         """
             Install an update, and block storage connection when we detect it's
-            being copied over to the inactive parition.
+            being copied over to the inactive partition.
 
             The test should result in a successful download retry.
             
--- a/testutils/infra/device.py
+++ b/testutils/infra/device.py
@@ -76,7 +76,7 @@ class MenderDevice:
         """Run given cmd in remote SSH host
 
         Argument:
-        cmd - sring with the command to execute remotely
+        cmd - string with the command to execute remotely
 
         Recognized keyword arguments:
         hide - do not print stdout nor stderr, and do not fail on errors


lluiscampos:
@mender-test-bot start pipeline :basketball:

tests/tests/test_fault_tolerance.py
```diff
@@ -224,7 +224,19 @@ def test_image_download_retry_timeout(
...
+                starttime = time.time()
+                while starttime + 60 * 10 >= time.time():
+                    output = mender_device.run(
+                        "fuser -mv %s" % (inactive_part), warn_only=True
+                    )
+                    if output.rstrip() != "":
+                        # If it succeeds, stop.
+                        break
+                    else:
+                        time.sleep(0.1)
```
kacf, line 236: Hmm, I'm a little bit worried about this change. I know that the reason for the exponential backoff is that for emulated QEMU CPUs, if the system is bogged down, it can get so slow that short timeouts like this essentially lead to no other work being performed, because SSH takes up all the processing. I recently saw this while developing the update control tests, so this is definitely still an issue. The backoff handles these situations by slowing down and leaving larger and larger gaps for other work to happen.

I have a suggestion for a fix which is more "backoff tolerant": Add an `ArtifactInstall_Enter_00` state script to the artifact, which creates a file, and then sleeps until that file is removed. You can get look for that file with `run()`, when it is present you manipulate the network settings, and then remove it for the client to continue.

How does that sound?

lluiscampos: That sounds like a much better approach, thanks! I'll rework this a a later time

lluiscampos:
@kacf I reworked the logic following your idea. Thanks again!

lluiscampos:
For some reason it really never returns from the script:
```
Jul 07 19:22:07 qemux86-64 mender[237]: time="2021-07-07T19:22:07Z" level=info msg="Executing script: ArtifactInstall_Enter_00"
Jul 07 19:22:07 qemux86-64 mender[237]: time="2021-07-07T19:22:07Z" level=info msg="Output (stdout) from command \"/var/lib/mender/scripts/ArtifactInstall_Enter_00\": Waiting..."
Jul 07 19:22:17 qemux86-64 mender[237]: time="2021-07-07T19:22:17Z" level=info msg="Output (stdout) from command \"/var/lib/mender/scripts/ArtifactInstall_Enter_00\": Waiting..."
```

At that point the file is removed but the mender-client does not advance form there. I'll look at it tomorrow with fresh eyes :sweat_smile:

tests/tests/test_fault_tolerance.py
```diff
@@ -222,9 +222,30 @@ def test_image_download_retry_timeout(
...
+touch /tmp/wait-for-install
+while [ -f /tmp/wait-for-install ]; do
+    echo Waiting...
+    sleep 10
+done
+exit 0"""
+                try:
+                    f = open("ArtifactInstall_Enter_00", "w")
+                    f.write(script_content)
+                    f.close()
```
kacf, line 239: A more idiomatic pattern would be:

```python
with open("ArtifactInstall_Enter_00", "w") as f:
    f.write(script_content)
```

tests/tests/test_fault_tolerance.py
```diff
@@ -222,9 +222,30 @@ def test_image_download_retry_timeout(
...
+
+                    deployment_id, new_yocto_id = common_update_procedure(
+                        valid_image, scripts=["ArtifactInstall_Enter_00"]
+                    )
+                finally:
+                    os.unlink(f.name)
+
+                # Wait until the update is in progress
+                mender_device.run("test -f /tmp/wait-for-install")
```
kacf, line 248: Are you really waiting though? `run()` will retry failed connections, but not failed commands, or?

kacf (commented):
Definitely a good approach, but I also cannot see why it hangs. Maybe related to the waiting I mentioned?

tests/tests/test_fault_tolerance.py
lluiscampos: But how do I then get the `finally` catch to remove the file regardless or errors?

tests/tests/test_fault_tolerance.py
lluiscampos: Yes, it "waits" in the sense that it retries the command until returns 0. See output below:
```
[gw0] TestFaultTolerance__test_image_download_retry_timeout[test_set0] -- 2021-07-08 09:56:20,925 [INFO]: >> Got UnexpectedExit while executing command in host 172.18.0.2: Encountered a bad command exit code!

Command: 'test -f /tmp/wait-for-install'

Exit code: 1

Stdout: already printed

Stderr: already printed


[gw0] TestFaultTolerance__test_image_download_retry_timeout[test_set0] -- 2021-07-08 09:56:22,942 [INFO]: >> Got UnexpectedExit while executing command in host 172.18.0.2: Encountered a bad command exit code!

Command: 'test -f /tmp/wait-for-install'

Exit code: 1

Stdout: already printed

Stderr: already printed


[gw0] TestFaultTolerance__test_image_download_retry_timeout[test_set0] -- 2021-07-08 09:56:27,061 [INFO]: >> Got UnexpectedExit while executing command in host 172.18.0.2: Encountered a bad command exit code!

Command: 'test -f /tmp/wait-for-install'

Exit code: 1

Stdout: already printed

Stderr: already printed


[gw0] TestFaultTolerance__test_image_download_retry_timeout[test_set0] -- 2021-07-08 09:56:36,187 [INFO]: >> Disallowing network communication to docker.mender.io
...
```

tests/tests/test_fault_tolerance.py
kacf: Well, you know the name, so you don't need `f.name`, do you? But it's not important, your way is fine too.

tests/tests/test_fault_tolerance.py
kacf: Oh, I didn't know this, actually. It's a little... unexpected, I would say. Is it clearer to use a while loop in the shell command instead? This also saves you from having to redo the SSH handshake, and produces less scary output.

Whether it solves the problem, you had, well that's another matter. Probably not.

tests/tests/test_fault_tolerance.py
lluiscampos: Ah, sorry I got confused. You just meant adding the `with` context manager but of course keeping the try/finally there. That I can do :)

lluiscampos:
> At that point the file is removed but the mender-client does not advance form there. I'll look at it tomorrow with fresh eyes sweat_smile

So I enabled debug logs in the client, last thing that happens after the state script returns is the client contacting the server... which does not timeout for at least 10 minutes (when the test gives up):

Jul 08 09:11:44 qemux86-64 mender[537]: time="2021-07-08T09:11:44Z" level=debug msg="Connecting to server docker.mender.io" func="github.com/mendersoftware/mender/client.(*ApiRequest).Do" file="/builds/Northern.tech/Mender/build-qemux86-64-uefi-grub/tmp/work/core2-64-poky-linux/mender-client/master-git999-r0/mender-client-master-git999/src/github.com/mendersoftware/mender/client/client.go:224"

kacf:
I guess it's trying to report its status. Dammit, perhaps this won't work the way I suggested. It looks like it depends on the connection failure happening during the download specifically. And we can't detect that with a state script (`Download_Enter` won't work either because it is before the download has started).

The only foolproof method I can see is using an Update Module to do the download in, and then halt it halfway, create the file, which results in manipulation of the network settings, and then resume by removing the file. But this raises the complexity considerably.

We could also just try your initially suggested fix, and if we do see additional failures, we can revisit this.

lluiscampos:
Actually, you gave me the solution in the other (unrelated) commit.

What we should do is to do the `fuse` method, but instead of relying on the SSH re-connections to repeat the command until it gives output, just have a simple shell script that waits. With high likelihood it will return while the download is still in progress and we have time to break the network. Maybe not bullet proof, but better that what we had.

lluiscampos:
@mender-test-bot start pipeline :goal_net:

lluiscampos:
The test failure is unrelated (old branch, I guess).

@kacf What do you think about the last rework?

kacf:
> Actually, you gave me the solution in the other (unrelated) commit.
> 
> What we should do is to do the `fuse` method, but instead of relying on the SSH re-connections to repeat the command until it gives output, just have a simple shell script that waits. With high likelihood it will return while the download is still in progress and we have time to break the network. Maybe not bullet proof, but better that what we had.

Yes! Perfect! I'm merging this. And cherry-picking it too.

PR #1463, opened 2021-07-07
MEN-4670: add test_api_endpoints to test auth on all the end points

tranchitella:
Changelog: none

Depends on:
* https://github.com/mendersoftware/deviceauth/pull/513
* https://github.com/mendersoftware/tenantadm/pull/304
* https://github.com/mendersoftware/useradm-enterprise/pull/308
* https://github.com/mendersoftware/inventory/pull/281
* https://github.com/mendersoftware/inventory-enterprise/pull/238
* https://github.com/mendersoftware/auditlogs/pull/79
* https://github.com/mendersoftware/deviceconnect/pull/123
* https://github.com/mendersoftware/deviceconfig/pull/60


tranchitella:
Pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/333360811

kjaskiewiczz (approved):
nice job!
I was looking for some tools which will generate these tests for us and I was surprised I couldn't find anything.

tranchitella:
> nice job!
> I was looking for some tools which will generate these tests for us and I was surprised I couldn't find anything.

All in all, reading the YAMLs directly is easy and requires less machinery.
Not super generic, but it will be enough for our tests.

tranchitella:
@lluiscampos should I wait for your review? I'm especially interested in your feedback on the dynamic pytest fixture.

backend-tests/tests/test_api_endpoints.py
```diff
@@ -0,0 +1,185 @@
...
+    "deviceconfig": "DEVICECONFIG_REV",
+    "deviceconnect": "DEVICECONNECT_REV",
+    "inventory": "INVENTORY_REV",
+    "inventory-enterprise": "INVENTORY_ENTERPRISE_REV",
+    "tenantadm": "TENANTADM_REV",
+    "useradm": "USERADM_REV",
+    "useradm-enterprise": "USERADM_ENTERPRISE_REV",
+    "workflows": "WORKFLOWS_REV",
+    "workflows-enterprise": "WORKFLOWS_ENTERPRISE_REV",
+}
```
lluiscampos, line 48: :+1: 

But we need to make a note in [mender-qa Documentation](https://github.com/mendersoftware/mender-qa/blob/master/Documentation/adding-new-repository.md) to update this list and `REPOS` below if a new repo being added has APIs (of just if it is a backend repository, which will do most of the times)

backend-tests/tests/test_api_endpoints.py
```diff
@@ -0,0 +1,185 @@
...
+        "deviceconfig",
+        "deviceconnect",
+        "inventory",
+        "useradm",
+        "workflows",
+    )
+
+    @pytest.mark.parametrize(
+        "kind,auth,method,scheme,host,path", get_all_api_endpoints(REPOS),
+    )
```
lluiscampos, line 163: Beautiful :heart_eyes:

lluiscampos (approved):
Simply amazing. Great work.

With regards to `SSH_PRIVATE_KEY`, it is okey as this tests will be run from `mender-qa` which is a private project (and has already the key to do many other things). What we are missing is a check at the beginning of the test to `pytest.skip` the test(s) if the env variable is not present, so that it doesn't get in the way when manually running tests.

lluiscampos:
> the usage of the env variables to checkout the correct branches

Forgot about this. I _think_ it is correct. Again, it will break on a local run, but that should be avoided if we implement the skip that I commented above.

backend-tests/tests/test_api_endpoints.py
tranchitella: https://github.com/mendersoftware/mender-qa/pull/519

tranchitella:
New pipeline run: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/334493310

PR #1464, opened 2021-07-08
update RBAC tests for configuration deployments

kjaskiewiczz:
With new, simplified RBAC we do not support any other actions than
VIEW_DEVICE when it comes to group permissions.
If user can view the device then user is allowed to deploy to the
device too.


kjaskiewiczz:

PR #1465, opened 2021-07-09
Make it easier to decide release components by deciding Ent first.

kacf:
Changelog: None


kacf:
@lluiscampos: What do you think about this?

lluiscampos (approved):
Yes! Thanks!

This is the same reason why I originally introduced the "skip" option, to decide later on OS repos. But of course this is much more efficient :)

kacf:
> This is the same reason why I originally introduced the "skip" option, to decide later on OS repos. But of course this is much more efficient :)

Yes, but both are useful independently, I think! 🙂

PR #1471, opened 2021-07-11
MEN-4682 monitor client full integration test

merlin-northern:
MEN-4682 Monitor: enable devicemonitor and monitor-client
MEN-4682 Monitor: alert email on service not running full integration test

    Test scenarios:
    * send CRITICAL alert email on service down
    * send OK alert email on service up

    Added elements:
    * new fixture: monitor_commercial_setup_no_client
    * new service: local-smtp, an smtpd that saves emails to /var/spool/mail
    * new docker compose method: get_file, get file from given container

### DIFF
--- a/docker-compose.monitor-client.commercial.yml
+++ b/docker-compose.monitor-client.commercial.yml
@@ -4,5 +4,12 @@ services:
     #
     # monitor-client
     #
-    monitor-client:
+    mender-client:
         image: registry.mender.io/mendersoftware/mender-monitor-qemu-commercial:master
+        networks:
+            - mender
+        stdin_open: true
+        tty: true
+        privileged: true
+        environment:
+            - TENANT_TOKEN=$TENANT_TOKEN
--- a/config/traefik/traefik.config.yaml
+++ b/config/traefik/traefik.config.yaml
@@ -1,5 +1,6 @@
 http:
   routers:
+
     #
     # deviceconfig
     #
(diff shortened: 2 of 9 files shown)


merlin-northern:
@mender-test-bot start pipeline sugar pretty please

kacf (approved):
Remember that only user facing changes need changelogs, so commits like the one containing your test don't need one (unless you also change functionality).

Other than that, looks good to me!

merlin-northern:
@mender-test-bot start pipeline please with cherry on top

merlin-northern:
@mender-test-bot start pipeline pretty sugar please with cherry on top

docker-compose.monitor-client.commercial.yml
```diff
@@ -4,5 +4,12 @@ services:
     #
     # monitor-client
     #
-    monitor-client:
+    mender-client:
```
lluiscampos, line 7: This will conflict with the `release_tool` querying (and eventually the next release). I am taking note on my TODO to think how to address it.

merlin-northern: > This will conflict with the `release_tool` querying (and eventually the next release). I am taking note on my TODO to think how to address it.

thank you!

merlin-northern: @kacf do you think that can be the reason behind the "old code" problem we discussed today? (I am now looking for every possible hint)

kacf: No, I don't think so, because there already exists an instance of this [here](https://github.com/mendersoftware/integration/blob/master/docker-compose.docker-client.addons.yml#L4-L6).

lluiscampos: I forgot to update here. I was wrong on my original assessment and this is not a problem (if any, if would have been a problem at the time of merging the feature branch).

lluiscampos: It came to my mind now: what we were missing is an update in `components-maps.yml`! See PR:
* https://github.com/mendersoftware/integration/pull/1556

PR #1479, opened 2021-07-13
QA-303: install curl in the mender-client-docker-addons

tranchitella:
curl is needed to report the device configuration to the backend
in the device configuration add-on.

Changelog: none


mzedel (approved):
Just out of curiosity: does the curl dependency exist generally for configure or just in the docker client?

tranchitella:
@mzedel exists in general, see:
https://github.com/mendersoftware/mender-configure-module/blob/master/src/mender-inventory-mender-configure#L64

mzedel:
👌 Thanks for the clarification @tranchitella!

PR #1480, opened 2021-07-14
Update version components for staging branch.

kacf:
Changelog: None

### DIFF
--- a/docker-compose.client.rofs.yml
+++ b/docker-compose.client.rofs.yml
@@ -5,4 +5,4 @@ services:
     # mender-client
     #
     mender-client:
-        image: mendersoftware/mender-client-qemu-rofs:2.6.0
+        image: mendersoftware/mender-client-qemu-rofs:3.0.0
--- a/docker-compose.client.yml
+++ b/docker-compose.client.yml
@@ -5,7 +5,7 @@ services:
     # mender-client
     #
     mender-client:
-        image: mendersoftware/mender-client-qemu:2.6.0
+        image: mendersoftware/mender-client-qemu:3.0.0
         networks:
             - mender
         stdin_open: true
--- a/docker-compose.docker-client.yml
+++ b/docker-compose.docker-client.yml
@@ -3,6 +3,6 @@ services:
 
     mender-client:
         # Needs to be built in mender client's test directory.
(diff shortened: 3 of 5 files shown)


kacf:

lluiscampos (approved):

kacf:
@tranchitella: Do you know if the errors are common on the staging branch?

kacf:
To be specific, I meant [this error](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/1434230522). The "ci/gitlab" error you can ignore.

tranchitella (approved):

lluiscampos:
I've been looking at the errors (they are consisntent). I am merging this PR even with the errors, see below.

* FAILED tests/test_configuration.py::TestConfiguration::test_configuration
This one is pre-existing, the exact same error can be reproduced both locally and [on GitLab](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/1437875301) with client 2.6.0 for integration/staging. My guess is that some other component have been updated in staging without really running all tests. I need to investigate further, but being pre-existing it should not block this PR

* FAILED tests/test_fault_tolerance.py::TestFaultTolerance::test_image_download_retry_timeout[test_set0]
This one fails when running on the client `modprobe xt_string`. The Docker image for mender-client-qemu:3.0.0 does not have it. Pulling the image and running the modprobe command returns "modprobe: FATAL: Module xt_string not found in directory /lib/modules/5.4.123-yocto-standard"

* FAILED tests/test_update_modules.py::TestUpdateModules::test_rootfs_update_module_success
Similar to the above, the module is missing from the image. The error reads: "Artifact Payload type 'rootfs-image-v2' is not supported by this Mender Client. Make sure the Update Module is installed on the device"

For the two last errors, the problem is that we need to update [this bbappend](https://github.com/mendersoftware/meta-mender/blob/903c1e7a660ae89d08171a4014b544aa5efc96dc/meta-mender-qemu/recipes-mender/mender-client/mender-client_2.%25.bbappend) to apply to mender 3.x (launching an independent PR soon).

I think that what happened is that release candidates still resolve to a 2.x kind of git version, so the bbappend did apply, but once we tagged the mender client 3.0.0, the bbapend was ignored. @kacf do you have a good idea to prevent this for happening in the future?

PR #1484, opened 2021-07-15
statistics-generator: Skip missing repositories.

kacf:
This makes it possible to run the generator on an old release, but
using a newer branch of integration. Normally this isn't necessary,
but it helps with the transition from mender-docs to
mender-docs-changelogs. We need to use a more recent branch of
integration in order to omit client components from the changelog, but
still need to be able to include those components in the legacy
changelog, by using the old branch of integration.

Changelog: None

### DIFF
--- a/extra/statistics-generator
+++ b/extra/statistics-generator
@@ -79,6 +79,9 @@ collect_changes() {
             pushd .
         else
             CHANGES="$("$RELEASE_TOOL" --version-of $repo --in-integration-version "$given_range")"
+            if [ -z "$CHANGES" ]; then
+                continue
+            fi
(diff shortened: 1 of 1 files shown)


lluiscampos (approved):

lluiscampos (commented):
There is a corner case, though. If/when we deprecate a repository, it will not be listed from the "recent" integration branch but still needed in the "old" integration branch to get the changelogs from.

I think we can live with this for now. We have plans to move these tools completely out of integration.

kacf:
I know. This is a duct tape solution. But hopefully a temporary one.

PR #1517, opened 2021-08-09
Add file lock around tests using SMTP mock in their docker composition

lluiscampos:


lluiscampos:
The motivation is to avoid errors as [this one](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/1488206018).

merlin-northern (approved):
lgtm thank you. but I will defer to Kristian for the final approval.

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -336,6 +336,8 @@ def setup(self):
 class DockerComposeMonitorCommercialSetup(DockerComposeNamespace):
     def __init__(self, name, num_clients=0):
         self.num_clients = num_clients
+        # Lock to avoid concurrent tests using SMTP mock (port 1025)
+        self.smtp_mock_lock = filelock.FileLock("smtp_mock_lock")
```
kacf, line 340: If all monitoring tests are going to use this, it's going to be slow because of lacking parallelism, and will get slower for every test we add. A couple of suggestions:

* Move the smtp part to a separate setup class and use it only in the tests that really require smtp.
  _-or-_
* Don't use a publicly exposed port, but instead set the hostname of the server by overriding the `WORKFLOWS_SMTP_HOST` variable when launching the environment. This allows it to use the internal server in its Docker network instead of a public address. This is the best solution, IMHO, since they can run in parallel like any other test.

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -346,12 +348,17 @@ def __init__(self, name, num_clients=0):
...
+        self.smtp_mock_lock.acquire(timeout=20 * 60)
         cmd = "up -d"
         if not recreate:
             cmd += " --no-recreate"
         self._docker_compose_cmd(cmd, env=env)
         self._wait_for_containers(self.NUM_SERVICES_ENTERPRISE + 1)
 
+    def teardown(self):
+        super().teardown()
+        self.smtp_mock_lock.release()
```
kacf, line 360: Nitpick: Construction/destruction semantics normally require a LIFO approach, in other words the `super.teardown()` should be called last. Unlikely to be important here, but it's good style nevertheless.

lluiscampos: I would normally do LIFO approach too, but in this particular case it can lead to error: the lock would be released before calling `docker-compose down`, and then the next test could potentially call `docker-compose up` while the port is still in use.

testutils/infra/container_manager/docker_compose_manager.py
lluiscampos: The second alternative sounds much better, of course.

I suggest we merge this as-is now (pipelines are failing as we speak), and I start working on your second approach straight away.

testutils/infra/container_manager/docker_compose_manager.py
kacf: You're right. And actually, your approach *is* LIFO, I just missed that in the construction, the `super` object is constructed last. 🙂

testutils/infra/container_manager/docker_compose_manager.py
kacf: :+1:

kacf (approved):

testutils/infra/container_manager/docker_compose_manager.py
lluiscampos: Actually, I am running the tests locally without exposing the port out of the Docker network and it "just works". I don't think this was needed in the first place... I am doing more local tests to be sure

lluiscampos: Nah, I was wrong. The tests need a bit of rework. Merging as agreed and coming up with a better fix soon :+1:

PR #1518, opened 2021-08-10
smtp.mock.yml: Correctly set network and not expose port

lluiscampos:
The port does not need to be exposed out of the Docker network. The
    problem was a typo in the `networks` key for `local-smtp` which made the
    container not connected to the Docker network. This was hidden by the
    fact that the port was exposed outside, and the tests used this outside
    interface.
    
    This fix allows for multiple tests using smtp mock to run in parallel.

And few minor clean-ups and improvements.

### DIFF
--- a/extra/smtp-testing/smtp.mock.yml
+++ b/extra/smtp-testing/smtp.mock.yml
@@ -7,8 +7,8 @@ services:
             - PYTHONUNBUFFERED=1
         command: bash -c "mkdir -p /var/spool/mail; stdbuf -oL python -m smtpd -n -c DebuggingServer 0.0.0.0:1025 > /var/spool/mail/local"
         ports:
-            - 1025:1025
+            - "1025"
         networks:
-            mender:
+            - mender
 networks:
     mender:
--- a/tests/tests/test_monitor_client.py
+++ b/tests/tests/test_monitor_client.py
@@ -186,7 +186,7 @@ class TestMonitorClientEnterprise:
         expected_from = "alert@mender.io"
         service_name = "crond"
         user_name = "bugs.bunny@acme.org"
-        devid, authtoken, auth, mender_device = self.prepare_env(
+        devid, _, _, mender_device = self.prepare_env(
             monitor_commercial_setup_no_client, user_name
         )
         logger.info("test_monitorclient_alert_email: env ready.")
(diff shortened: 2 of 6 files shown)


lluiscampos:
This PR is a follow-up from https://github.com/mendersoftware/integration/pull/1517#discussion_r685825474

lluiscampos:

lluiscampos:
@mender-test-bot Seems like you ignored me the first time :angry: Please start pipeline :angel:

extra/smtp-testing/smtp.mock.yml
```diff
@@ -7,8 +7,8 @@ services:
             - PYTHONUNBUFFERED=1
         command: bash -c "mkdir -p /var/spool/mail; stdbuf -oL python -m smtpd -n -c DebuggingServer 0.0.0.0:1025 > /var/spool/mail/local"
         ports:
-            - 1025:1025
+            - "1025"
```
merlin-northern, line 10: ah right! it was a copy-paste from a setup where I accessed it outside of the "local" network.

lluiscampos: But this alone didn't fix the issue, as the `networks` below was incorrect. I couldn't find an exact reference in the Docker compose documentation, but it seems like in the top level you use `mender:` and it is enough but here for the services we require `-mender`.

merlin-northern (approved):
lgtm! thanks!

tests/tests/test_monitor_client.py
```diff
@@ -159,7 +159,7 @@ def prepare_env(self, env, user_name):
         devauth_tenant = DeviceAuthV2(auth)
 
         env.new_tenant_client("configuration-test-container", tenant["tenant_token"])
-        mender_device = MenderDevice(env.get_mender_clients()[0])
+        env.device = mender_device = MenderDevice(env.get_mender_clients()[0])
```
merlin-northern, line 162: why exactly do we need this? (I mean env.device assignment)

lluiscampos: For our hook to find the device on errors. See [these lines](https://github.com/mendersoftware/integration/blob/4d4c46dcc29e713a1d0ac4278d674509f4ec15ea/tests/conftest.py#L121-L139).

kacf (approved):
:heart_eyes:

PR #1524, opened 2021-08-12
Use `--help` instead of `--version`

tranchitella:
Changelog: none


tranchitella:

mchalski (approved):

PR #1546, opened 2021-08-31
release_tool: Remove --verify-integration-references option

lluiscampos:


lluiscampos:
See https://github.com/mendersoftware/mender-qa/pull/527/commits/3dc0af52afdc0c66682fec6993bf2fd9d02b3d9b

extra/release_tool.py
```diff
@@ -3085,72 +3085,6 @@ def do_hosted_release(version=None):
...
-
-        # Try some common locations.
-        paths = ["..", "../go/src/github.com/mendersoftware"]
-        path = find_repo_path(repo.git(), paths)
-
-        if path is None:
-            print("%s not found. Tried: %s" % (repo.git(), ", ".join(paths)))
-            sys.exit(2)
-
-        revs = figure_out_checked_out_revision(None, path)
```
kacf, line 3108: You can also remove this function. Not used anywhere else, AFAICS.

kacf (approved):
Pre-approving, but take a look below.

extra/release_tool.py
lluiscampos: Good catch, thank you.

PR #1549, opened 2021-09-01
MEN-4999: test_monitoring.py: POST /alert with array

lluiscampos:


lluiscampos:
The three PRs go together:
* https://github.com/mendersoftware/devicemonitor/pull/45
* https://github.com/mendersoftware/monitor-client/pull/49
* https://github.com/mendersoftware/integration/pull/1549

merlin-northern (approved):
lgtm. one comment (I leave it up to you): do you think we should test also scenarios with empty array POST `[]` and one alert in addition to multiple alerts?

lluiscampos:
> lgtm. one comment (I leave it up to you): do you think we should test also scenarios with empty array POST `[]` and one alert in addition to multiple alerts?

One alert is already tested, see that there are two test cases that use the same code base.

With regards to the empty case... I'll have a look, maybe is already covered in the unit test level.

lluiscampos:
> With regards to the empty case... I'll have a look, maybe is already covered in the unit test level.

Tested in unit tests :heavy_check_mark:

PR #1550, opened 2021-09-01
QA-233: Add progress tracking to pytest suites

lluiscampos:
Using `pytest-sugar` package

### DIFF
--- a/backend-tests/requirements-py3.txt
+++ b/backend-tests/requirements-py3.txt
@@ -17,11 +17,13 @@ pyparsing==2.4.7
 pytest==6.2.5
 pytest-html==3.1.1
 pytest-metadata==1.11.0
+pytest-sugar==0.9.4
 PyYAML==5.4.1
 pyzbar==0.1.8
 redo==2.0.4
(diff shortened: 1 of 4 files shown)


lluiscampos:

lluiscampos:
It doesn't seem to work in CI :shrug:

oleorhagen:
> It doesn't seem to work in CI shrug

:'(

lluiscampos:

lluiscampos:
@oleorhagen Trying with this: https://github.com/Teemu/pytest-sugar/issues/219#issuecomment-786011728 :crossed_fingers:

lluiscampos:
Now it works, do you think is good enough? I am not convinced on how useful it is, because these percentages are not printed after every test, but once on a while for a bunch of tests.

oleorhagen:
Having a look now, but if it is as you say, then I agree :(

oleorhagen:
It seems u are right, this is not what we wanted :'(

lluiscampos:
It was worth a try. Closing.

PR #1576, opened 2021-09-16
test_api_endpoints.py: Add devicemonitor endpoints to the test

lluiscampos:

PR #1577, opened 2021-09-16
3.1.x: test_api_endpoints.py: remove deviceconfig

lluiscampos:
Mender 3.1.x re-uses deviceconfig 1.1.0, which can't run this test.

### DIFF
--- a/backend-tests/tests/test_api_endpoints.py
+++ b/backend-tests/tests/test_api_endpoints.py
@@ -164,7 +164,8 @@ class TestAPIEndpoints(BaseTestAPIEndpoints):
     REPOS = (
         "deployments",
         "deviceauth",
-        "deviceconfig",
+        # Mender 3.1.x re-uses deviceconfig 1.1.0, which can't run this test
+        # "deviceconfig",
         "deviceconnect",
         "inventory",
         "useradm",
@@ -191,7 +192,8 @@ class TestAPIEndpointsEnterprise(BaseTestAPIEndpoints):
         "auditlogs",
         "deployments-enterprise",
         "deviceauth",
-        "deviceconfig",
+        # Mender 3.1.x re-uses deviceconfig 1.1.0, which can't run this test
+        # "deviceconfig",
         "deviceconnect",
         "devicemonitor",
         "inventory-enterprise",


lluiscampos:
And cherry-picking #1576 too.

lluiscampos:
@kacf @tranchitella Does it make sense to you that we just skip the repo this way?

The issue is that `deviceconfig` has not gotten any software fixes so we are not releasing it, however the test requires [this change](https://github.com/mendersoftware/deviceconfig/commit/ea19c1ad2656b3c2bf00bc950ccc93c059aaa8e6) to run correctly.

tranchitella:
@lluiscampos the idea here is NOT releasing a new version of deviceconfig?
My only objection is that the API docs we publish is wrong, as you can see here:
https://docs.mender.io/api/3.0/#internal-api-device-configure
Maybe we should backport the commit to 3.0.x as well.

lluiscampos:
> @lluiscampos the idea here is NOT releasing a new version of deviceconfig?
> My only objection is that the API docs we publish is wrong, as you can see here:
> https://docs.mender.io/api/3.0/#internal-api-device-configure
> Maybe we should backport the commit to 3.0.x as well.

There are no changelogs from 1.1.0 to today's master, so it doesn't make sense to release it.

For the Mender API docs, it will be enough with cherry-picking (as I have done for `deviceconnect` see https://github.com/mendersoftware/deviceconnect/pull/150 ) -> I'll do this right now.

lluiscampos:
> For the Mender API docs, it will be enough with cherry-picking (as I have done for `deviceconnect` see [mendersoftware/deviceconnect#150](https://github.com/mendersoftware/deviceconnect/pull/150) ) -> I'll do this right now.

Done: https://github.com/mendersoftware/deviceconfig/pull/77

tranchitella (approved):

kacf (approved):
Fine with me. This will be immediately reverted afterwards, right?

Another possibility would be to simply ignore this failure, and run the release build without tests, which I guess we do anyway.

lluiscampos:
> Fine with me. This will be immediately reverted afterwards, right?

No, this is going only to `integration/3.1.x` and is not meant to be reverted.

> 
> Another possibility would be to simply ignore this failure, and run the release build without tests, which I guess we do anyway.

That would be possible but I think is inconvenient (it will raise questions when doing Mender 3.1.1, for example).

kacf:
Ah right, ok, it will not be reverted unless we release a new version of deviceconfig. Got it.

PR #1579, opened 2021-09-17
release_tool: Update extra files also from CLI call

lluiscampos:
This is an amend to commit c68e632, which correctly sets the versions of
the extra docker compose files when releasing, but missed them when
using the --set-version-of CLI argument.

Fixes test failures from 3.1.x release branch, where these files are set
to a non-existing version (iow not set to "pr" by CI script).

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -2749,6 +2749,20 @@ def do_set_version_to(args):
         integration_dir(), repo, args.version, git_tag=args.version
     )
 
+    # Update extra files used in integration tests
+    set_docker_compose_version_to(
+        os.path.join(integration_dir(), "extra", "mtls"),
+        repo,
+        args.version,
+        git_tag=args.version,
+    )
+    set_docker_compose_version_to(
+        os.path.join(integration_dir(), "extra", "failover-testing"),
+        repo,
+        args.version,
+        git_tag=args.version,
+    )
+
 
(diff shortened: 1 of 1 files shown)


lluiscampos:
Original PR #1446

lluiscampos:
@mender-test-bot start pipeline just in case

extra/release_tool.py
```diff
@@ -2749,6 +2749,20 @@ def do_set_version_to(args):
...
+        repo,
+        args.version,
+        git_tag=args.version,
+    )
+    set_docker_compose_version_to(
+        os.path.join(integration_dir(), "extra", "failover-testing"),
+        repo,
+        args.version,
+        git_tag=args.version,
+    )
```
kacf, line 2764: I don't have a better suggestion at the moment, but we should think about a better way to do this. It's not very maintainable long term.

kacf (approved):

extra/release_tool.py
lluiscampos: I first thought about making `set_docker_compose_version_to` recursively look for all `yml` files, so that here a single call would do, but the challenge is that there are some files (like the compatibility tests for the client) that we don't want to modify, and then we would need to handle that as a exception of some sort.

Conclusion: I don't have a good suggestion neither :)

PR #1582, opened 2021-09-20
demo: Use enterprise QEMU client for Enterprise testing

lluiscampos:


demo
```diff
@@ -298,7 +299,7 @@ maybe_launch_enterprise_client() {
                     --id $TENANT_ID \
                     | jq -r .tenant_token)
             # Now that we have the tenant token we can enable the client.
-            EXTRA_FILES="$EXTRA_FILES $CLIENT_FILES"
+            EXTRA_FILES="$EXTRA_FILES $ENTERPRISE_CLIENT_FILES"
```
kacf, line 302: I think that only adding it here will give inconsistent handling. I think this runs only after starting the demo, so if you run `./demo pull`, it will pull the wrong image, won't it?

lluiscampos: Assuming you meant `./demo --enterprise-testing --client pull` yes, I think you are right. I'll see how can I rework this :thinking:

lluiscampos: @kacf Can you check the rework?

kacf: :+1:

kacf (approved):

PR #1589, opened 2021-09-24
MEN-5108 Monitoring tests: file descriptors count check

merlin-northern:
ChangeLog:none

### DIFF
--- a/tests/tests/test_monitor_client.py
+++ b/tests/tests/test_monitor_client.py
@@ -1162,6 +1162,30 @@ class TestMonitorClientEnterprise:
                 + devid
             )
 
+        fds_count_timeout_s = 64
+        max_fds_count_diff = 2
+        logger.info("test_monitorclient_alert_email: checking open file descriptors")
+        time.sleep(fds_count_timeout_s)
+        fds_count0 = mender_device.run(
+            "ls /proc/$(cat /var/run/monitoring-client.pid)/fd | wc -l"
+        )
+        assert fds_count0.rstrip().isnumeric()
+        fds_count0 = int(fds_count0.rstrip())
+        logger.info(
+            "test_monitorclient_alert_email: currently %s fds open" % fds_count0
+        )
+        time.sleep(fds_count_timeout_s)
+        fds_count1 = mender_device.run(
+            "ls /proc/$(cat /var/run/monitoring-client.pid)/fd | wc -l"
+        )
+        assert fds_count1.rstrip().isnumeric()
+        fds_count1 = int(fds_count1.rstrip())
+        logger.info(
+            "test_monitorclient_alert_email: currently %s fds open" % fds_count0
+        )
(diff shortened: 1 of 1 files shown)


merlin-northern:
@mender-test-bot start pipeline hola guapo

tests/tests/test_monitor_client.py
```diff
@@ -1162,6 +1162,30 @@ def test_monitorclient_logs_and_services(self, monitor_commercial_setup_no_clien
...
+        time.sleep(fds_count_timeout_s)
+        fds_count1 = mender_device.run(
+            "ls /proc/$(cat /var/run/monitoring-client.pid)/fd | wc -l"
+        )
+        assert fds_count1.rstrip().isnumeric()
+        fds_count1=int(fds_count1.rstrip())
+        logger.info(
+            "test_monitorclient_alert_email: currently %s fds open" % fds_count0
+        )
+        assert abs(fds_count1 - fds_count0) <= max_fds_count_diff
```
lluiscampos, line 1186: Curious: why do we need this tolerance of 2 FD difference?

lluiscampos (approved):
Looks good.

This is a good candidate for a client acceptance tests - I'll take note on my value add TODO list to prepare the framework for it, because once we have it it easy (and cheap) to add more client-only checks like this one.

tests/tests/test_monitor_client.py
merlin-northern: it is possible that at the beginning we caught the daemon before the start of both logging checks, in that case the open fd count is 4. at the end we usually have 6, since both checks add at least one open fd. depending on which "sleep interval" we caught at the last check, the open fd count can fluctuate between these two. but one thing for sure: it never should increase. makes sense?

merlin-northern:
last [force](https://github.com/mendersoftware/integration/compare/f246655ef02a614f15fbb49158c8d12800ddb483..e096af3aa3b62e0145b0c68a42d49115a25cada0) push was just formatting, which I missed:
```diff
            "ls /proc/$(cat /var/run/monitoring-client.pid)/fd | wc -l"
        )
        assert fds_count0.rstrip().isnumeric()
-        fds_count0=int(fds_count0.rstrip())
+        fds_count0 = int(fds_count0.rstrip())
        logger.info(
            "test_monitorclient_alert_email: currently %s fds open" % fds_count0
        )
@@ -1179,7 +1179,7 @@ def test_monitorclient_logs_and_services(self, monitor_commercial_setup_no_clien
            "ls /proc/$(cat /var/run/monitoring-client.pid)/fd | wc -l"
        )
        assert fds_count1.rstrip().isnumeric()
-        fds_count1=int(fds_count1.rstrip())
+        fds_count1 = int(fds_count1.rstrip())
        logger.info(
            "test_monitorclient_alert_email: currently %s fds open" % fds_count0
        )
```

PR #1592, opened 2021-09-26
Monitoring: Perl compatible regular expression explicit test.

merlin-northern:
ChangeLog:none

### DIFF
--- a/tests/tests/test_monitor_client.py
+++ b/tests/tests/test_monitor_client.py
@@ -1009,7 +1009,7 @@ class TestMonitorClientEnterprise:
             monitor_commercial_setup_no_client, user_name
         )
         inventory = Inventory(auth)
-        logger.info("test_monitorclient_alert_email: env ready.")
+        logger.info("test_monitorclient_logs_and_services: env ready.")
 
         logger.info(
             "test_monitorclient_logs_and_services: email alert on systemd service not running scenario."
@@ -1170,7 +1170,9 @@ class TestMonitorClientEnterprise:
 
         fds_count_timeout_s = 64
         max_fds_count_diff = 2
-        logger.info("test_monitorclient_alert_email: checking open file descriptors")
+        logger.info(
+            "test_monitorclient_logs_and_services: checking open file descriptors"
+        )
         time.sleep(fds_count_timeout_s)
         fds_count0 = mender_device.run(
             "ls /proc/$(cat /var/run/monitoring-client.pid)/fd | wc -l"
@@ -1178,7 +1180,7 @@ class TestMonitorClientEnterprise:
         assert fds_count0.rstrip().isnumeric()
         fds_count0 = int(fds_count0.rstrip())
         logger.info(
-            "test_monitorclient_alert_email: currently %s fds open" % fds_count0
+            "test_monitorclient_logs_and_services: currently %s fds open" % fds_count0
         )
         time.sleep(fds_count_timeout_s)
         fds_count1 = mender_device.run(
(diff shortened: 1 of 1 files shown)


merlin-northern:
@mender-test-bot start pipeline really pretty please

merlin-northern:
@mender-test-bot start pipeline really pretty please with a cherry

merlin-northern:
@lluiscampos these go together:

* https://github.com/mendersoftware/integration/pull/1592
* https://github.com/mendersoftware/monitor-client/pull/77

Pipeline without https://github.com/mendersoftware/monitor-client/pull/77:
* https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/377460377 (failed)
Pipeline with https://github.com/mendersoftware/monitor-client/pull/77:
* https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/377460559 (passed)

tests/tests/test_monitor_client.py
```diff
@@ -1187,10 +1189,84 @@ def test_monitorclient_logs_and_services(self, monitor_commercial_setup_no_clien
...
+            monitor_commercial_setup_no_client, user_name
+        )
+        inventory = Inventory(auth)
+        logger.info("test_monitorclient_logs_and_patterns: env ready.")
+
+        logger.info(
+            "test_monitorclient_logs_and_patterns: email alert on log file containing a pattern scenario."
+        )
+        log_file = "/tmp/mylog.log"
+        log_pattern = "session opened for user \\w+"
```
lluiscampos, line 1215: Should this be:
```suggestion
        log_pattern = "session opened for user \w+"
```
?

tests/tests/test_monitor_client.py
```diff
@@ -1187,10 +1189,84 @@ def test_monitorclient_logs_and_services(self, monitor_commercial_setup_no_clien
...
+            "test_monitorclient_logs_and_services: currently %s fds open" % fds_count0
         )
         assert abs(fds_count1 - fds_count0) <= max_fds_count_diff
-        logger.info("test_monitorclient_alert_email: fds count have not increased")
+        logger.info(
+            "test_monitorclient_logs_and_services: fds count have not increased"
+        )
+
+    def test_monitorclient_logs_and_patterns(self, monitor_commercial_setup_no_client):
+        """Tests the monitor client email alerting for multiple patterns types"""
```
lluiscampos, line 1200: It says "multiple patterns types" but I only see one type: PCRE

tests/tests/test_monitor_client.py
```diff
@@ -1187,10 +1189,84 @@ def test_monitorclient_logs_and_services(self, monitor_commercial_setup_no_clien
...
+        mender_device.run(
+            "echo -ne 'a new session opened for user root now\nsome line 5\n' >> "
+            + log_file
+        )
+        time.sleep(2 * wait_for_alert_interval_s)
+
+        alerts, alert_count = self.get_alerts_and_alert_count_for_device(
+            inventory, devid
+        )
+        assert (True, 1) == (alerts, alert_count)
```
lluiscampos, line 1248: Why won't we have two alerts? :thinking:

lluiscampos (approved):
This is a another good example of a test that could have been at acceptance level.

tests/tests/test_monitor_client.py
merlin-northern: I actually paused the test, and verified, that it is correctly present in the check file: 
```bash
2021-09-27 08:07:44 [INFO] [ctl.sh:cleanup:15] exiting.
root@qemux86-64:~# cat /etc/mender-monitor/monitor.d/available/log_mylog-pcre.sh
# This file was autogenerated by Monitoring Utilities based on the configuration
SERVICE_NAME="mylog-pcre"
LOG_PATTERN="session opened for user \w+"
LOG_FILE="@tail -f /tmp/mylog.log"
```

the single \ in the double quotes will get interpreted, while \\ will be verbatim \
in this case you maybe right, since '\w' is not a valid sequence (as \n is) but I believe it is "more correct" to use \\ in this case, since we want "\w" in the string

tests/tests/test_monitor_client.py
merlin-northern: true. what a lie :) I will correct this.

tests/tests/test_monitor_client.py
merlin-northern: the same service, check, on the same device always results in one alert as counted by inventory: this is by design, it is counted as one current critical, see the earlier test logs_and_services, where we have 2, because there are 2 services.

PR #1593, opened 2021-09-27
[MEN-4852] Fix demo script user creation race condition

alfrunes:
The demo script will now probe all healthcheck endpoints assigned to
docker-compose services using labels until all services becomes ready to
accept traffic.

### DIFF
--- a/demo
+++ b/demo
@@ -19,7 +19,7 @@ RUN_UP=0
 UPLOAD_ARTIFACT=1
 PRINT_LOGIN_INFO=0
 PRINT_USER_EXISTS=0
-RETRY_LIMIT=5
+RETRY_LIMIT=30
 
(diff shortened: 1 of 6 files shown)


alfrunes:

lluiscampos (commented):
What about [docker-compose.monitor.yml](https://github.com/mendersoftware/integration/blob/master/docker-compose.monitor.yml)?

alfrunes:
> What about [docker-compose.monitor.yml](https://github.com/mendersoftware/integration/blob/master/docker-compose.monitor.yml)?

I didn't find the endpoint in the API spec so I assumed it wasn't there, however, after inspecting the code it is indeed present.

tranchitella (approved):

PR #1604, opened 2021-09-27
[MEN-4852] Fix demo script user creation race condition

alfrunes:
Cherry-pick #1593 to 3.1.x

PR #1608, opened 2021-09-29
3.1.x: [MEN-4852] Fix demo script user creation race condition

lluiscampos:
Cherry-picking #1593


lluiscampos:
Ups, already done at #1604.

PR #1609, opened 2021-09-29
MEN-5124: adjust tests to the new email sender in workflows-enterprise

tranchitella:
Changelog: none

### DIFF
--- a/tests/tests/test_monitor_client.py
+++ b/tests/tests/test_monitor_client.py
@@ -274,7 +274,7 @@ class TestMonitorClientEnterprise:
         """Tests the monitor client email alerting"""
         mailbox_path = "/var/spool/mail/local"
         wait_for_alert_interval_s = 8
-        expected_from = "noreply@mender.io"
+        expected_from = "no-reply@hosted.mender.io"
         service_name = "crond"
         user_name = "some.user+{}@example.com".format(str(uuid.uuid4()))
         devid, _, auth, mender_device = self.prepare_env(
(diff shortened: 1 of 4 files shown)


tranchitella:
Combined QA pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/379384822

tests/tests/test_monitor_client.py
```diff
@@ -274,7 +274,7 @@ def test_monitorclient_alert_email(self, monitor_commercial_setup_no_client):
         """Tests the monitor client email alerting"""
         mailbox_path = "/var/spool/mail/local"
         wait_for_alert_interval_s = 8
-        expected_from = "noreply@mender.io"
+        expected_from = "no-reply@hosted.mender.io"
```
merlin-northern, line 277: `noreply@mender.io"` was something that was agreed with Eystein, I know these are only tests -- but do we have a good reason not to keep noreply@mender.io here?

tranchitella: Because of SPF/DMARC/DKIM settings, we cannot send emails for `@mender.io` from a non-Google SMTP server. That's why we have to move to `@hosted.mender.io`. We can keep whatever we want in the test, but I tried to keep it as close as possible to the reality.

merlin-northern: exactly my point to keep it close to reality :) if we now cannot send from @medner.io, then it is fine. thanks. I will let Eystein know.

docker-compose.override.os.yml
```diff
@@ -0,0 +1,5 @@
+version: '2.1'
+services:
+    mender-gui:
+        volumes:
+            - ../gui/dist:/var/www/mender-gui/dist
```
merlin-northern, line 5: why is that needed? probably should be a separate commit?

tranchitella: oh sorry, this is a local file which wasn't meant to be added. I'm removing it.

tranchitella:
I won't rerun the pipeline, the force-commit is because of an extra file which was added by mistake (`docker-compose.override.os.yml`).

merlin-northern (approved):
lgtm

PR #1617, opened 2021-09-30
release_tool: Configure child pipelines on build action

lluiscampos:


extra/release_tool.py
```diff
@@ -1738,6 +1741,8 @@ def set_param(name, value):
                 action = "false"
             set_param("BUILD_CLIENT", action)
             set_param("BUILD_SERVERS", action)
+            set_param("BUILD_MENDER_DIST_PACKAGES", action)
+            set_param("BUILD_MENDER_CONVERT", action)
```
lluiscampos, line 1745: @kacf This one I am uncertain about: I don't really see the use case of building this as part of "all platforms" scenario.

kacf: Right, I didn't even think about this scenario before I saw it here. But I think it makes sense. We should start thinking about mender-qa as "the complete thing".

We can at least enable it now, and reassess if it works well in practice after having used it for a while.

kacf (approved):

PR #1646, opened 2021-10-14
tests: Bump the Debian version to Bullseye

oleorhagen:
This will give us a recent enough version of Rust, and make cryptography happy.

Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -83,7 +83,7 @@ test:docs:
     - ./run-tests.sh
 
 test:staging:backend-tests:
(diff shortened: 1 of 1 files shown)


lluiscampos:
This is okey.

One thing we could think about is to use `python` image instead. This is the case for backend acceptance tests, where the cryptography package has never broke anything, but not the case on integration / backend integration tests suites, where we have had problems in the past with this package and Rust. (I am talking about mender-qa piepline test suites, assuming that integration pipeline is inspired by that ones).

PR #1660, opened 2021-10-18
MEN-5149 Azure IoT Manager repository added.

merlin-northern:
ChangeLog:none

### DIFF
--- a/component-maps.yml
+++ b/component-maps.yml
@@ -4,6 +4,13 @@
 # and "docker_container". Binary tools that don't have Docker components, will
 # only have the "git" part.
 git:
+  azure-iot-manager:
+    docker_image:
+    - azure-iot-manager
+    docker_container:
+    - mender-azure-iot-manager
+    release_component: true
+
   deployments:
     docker_image:
     - deployments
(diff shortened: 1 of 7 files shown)


lluiscampos (approved):

demo
```diff
@@ -216,6 +216,7 @@ start_server() {
             -f docker-compose.yml \
             -f docker-compose.storage.minio.yml \
             -f docker-compose.demo.yml \
+            -f docker-compose.azure.yml \
```
kacf, line 219: Regarding the demo: Does this require extra steps to work, or is it just an additional component that is offered, but won't do anything unless you actively engage with it?

component-maps.yml
```diff
@@ -4,6 +4,13 @@
 # and "docker_container". Binary tools that don't have Docker components, will
 # only have the "git" part.
 git:
+  azure-iot-manager:
+    docker_image:
+    - azure-iot-manager
+    docker_container:
+    - mender-azure-iot-manager
+    release_component: false
```
kacf, line 12: This is missing the corresponding `docker_image` and `docker_container` entries.

kacf (changes requested):

component-maps.yml
kacf, line 12: This should be set to true actually.

PR #1666, opened 2021-10-20
MEN-5149 Azure IoT Manager repository added.

merlin-northern:
ChangeLog:none


lluiscampos:
Yeah, if the service is not part of the main `docker-compose.yml` file then we should have not updated the servers count in https://github.com/mendersoftware/mender-docs/pull/1569 :shrug: 

Too late now, I know. But I also support having it in the main `docker-compose.yml` file. This file should contain all OS services except the ones required for add-ons.

PR #1683, opened 2021-10-27
Upgrade nats to 2.6.2 and enable JetStream

tranchitella:
Changelog: none

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -146,6 +146,7 @@ services:
             - mender
         depends_on:
             - mender-mongo
+            - mender-nats
         labels:
             mender.healthcheck.path: "/health"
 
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -197,5 +200,11 @@ services:
                     - mongo-useradm
                     - mongo-workflows
 
+    mender-nats:
+        image: nats:2.6.2-alpine3.14
```
alfrunes, line 204: Latest version of NATS is 2.6.4.
```suggestion
        image: nats:2.6.4-alpine3.14
```

alfrunes (approved):

docker-compose.yml
tranchitella: With mongo, we secify the minor version (`4.4`); should we pin `2.6` for nats, maybe?

PR #1710, opened 2021-11-08
release_tool: Add the option to version the components-map file

oleorhagen:
Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -77,9 +77,14 @@ class Component:
     name = None
     type = None
 
+    _integration_version = None
+
     def __init__(self, name, type):
         self.name = name
         self.type = type
+
+    def set_integration_version(version):
+        Component._integration_version = version
 
     def git(self):
         if self.type != "git":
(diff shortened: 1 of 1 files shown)


extra/release_tool.py
```diff
@@ -112,8 +120,14 @@ def set_custom_component_maps(self, maps):
...
-            with open(os.path.join(integration_dir(), "component-maps.yml")) as fd:
-                Component.COMPONENT_MAPS = yaml.safe_load(fd)
+            component_maps = execute_git(
+                None,
+                integration_dir(),
+                ["show", f"{Component.integration_version()}:component-maps.yml"],
+                capture=True,
+                capture_stderr=True,
+            )
+            Component.COMPONENT_MAPS = yaml.safe_load(component_maps)
```
kacf, line 130: `in_integration_version` is not guaranteed to be set. You need to keep the original code path if `_integration_version is None`.

kacf (changes requested):

extra/release_tool.py
oleorhagen: True, fixed!

extra/release_tool.py
```diff
@@ -112,8 +117,19 @@ def set_custom_component_maps(self, maps):
...
+                    integration_dir(),
+                    ["show", f"{Component._integration_version}:component-maps.yml"],
+                    capture=True,
+                    capture_stderr=True,
+                )
+                Component.COMPONENT_MAPS = yaml.safe_load(component_maps)
+            else:
+                with open(os.path.join(integration_dir(), "component-maps.yml")) as fd:
+                    Component.COMPONENT_MAPS = yaml.safe_load(fd)
+            Component.COMPONENT_MAPS = yaml.safe_load(component_maps)
```
kacf, line 132: Remove? This will probably get "variable not found" in the non-git case, and you already are doing it in each branch.

kacf (changes requested):

extra/release_tool.py
```diff
@@ -112,8 +117,18 @@ def set_custom_component_maps(self, maps):
...
         if Component.COMPONENT_MAPS is None:
-            with open(os.path.join(integration_dir(), "component-maps.yml")) as fd:
-                Component.COMPONENT_MAPS = yaml.safe_load(fd)
+            if Component._integration_version:
+                component_maps = execute_git(
+                    None,
+                    integration_dir(),
+                    ["show", f"{Component._integration_version}:component-maps.yml"],
+                    capture=True,
+                    capture_stderr=True,
```
kacf, line 126: Btw, it occurred to me that this should probably be false, so that if you mistype the version you can actually see it, instead of getting a very cryptic parsing error.

kacf (approved):

oleorhagen:

oleorhagen:

lluiscampos (approved):

PR #1723, opened 2021-11-09
pipeline: test:docs: Run the doc tests on the correct branch

lluiscampos:


.gitlab-ci.yml
```diff
@@ -78,8 +78,28 @@ test:docs:
...
+    - MENDER_DOCS_BRANCH=""
+    - for release_branch in
+          origin/master
+          origin/staging
+          $(git for-each-ref
+            --format="%(refname:short)"
+            --sort=-version:refname:short
+            refs/remotes/mender/[0-9].[0-9].x
+          ); do
+    -   if [[ -n "$(git merge-base --fork-point $release_branch HEAD)" ]]; then
```
kacf, line 90: I *think* that `--fork-point` has no effect here. It only matters if any of the branches are rebased, which they never will be. And in fact, even if they were, this repository would not have enough information to use `--fork-point` effectively, since it depends on the reflog, which is not cloned across repositories.

More generally, I don't think this will work correctly, because merge-base will always return a match, even if it's very old. What you want is actually [this code](https://github.com/mendersoftware/integration/blob/3.1.0/extra/release_tool.py#L549-L560).

lluiscampos: Thanks :+1:

kacf (approved):

PR #1726, opened 2021-11-10
MEN-5178: Enable the demo script to handle the new docker-compose major version

oleorhagen:


mzedel (changes requested):
the version comparison unfortunately still doesn't work with the updated version check: 
```
$: ./verify-docker-versions
ERROR: Unsupported major version: v2
ERROR: Docker Compose must be on version 1.6 or larger, but found version v2.1.1 installed.
DOCKER_COMPOSE_VERSION_STRING=v2.1.1
$: docker info 2>/dev/null | grep "Server Version"
 Server Version: 20.10.10
```
mind the updated version string, starting with a _v_ + the server version...

verify-docker-versions
```diff
@@ -8,14 +8,24 @@ VERSION_MINOR_MIN_DOCKER_ENGINE="11"
...
   if [[ -z $VERSION_ACTUAL ]]; then
     echo "ERROR: Actual version is empty (hint: is it installed, or version format change?)."
     return 1
   fi
 
+  if [ "$VERSION_MAJOR" == "2" ]; then
+    return 0
+  fi
+
+  if [ "VERSION_MAJOR" != "1" ]; then
```
lluiscampos, line 24:
```suggestion
  if [ "$VERSION_MAJOR" != "1" ]; then
```

oleorhagen:
> the version comparison unfortunately still doesn't work with the updated version check:
> 
> ```
> $: ./verify-docker-versions
> ERROR: Unsupported major version: v2
> ERROR: Docker Compose must be on version 1.6 or larger, but found version v2.1.1 installed.
> DOCKER_COMPOSE_VERSION_STRING=v2.1.1
> $: docker info 2>/dev/null | grep "Server Version"
>  Server Version: 20.10.10
> ```
> 
> mind the updated version string, starting with a _v_ + the server version...

Sorry, missed this. Curious, I'm on docker-compose 2.x as well, but my version string is not `v` prefixed :confused:

verify-docker-versions
oleorhagen: :see_no_evil: I tried going with the local flow here, I usually like the `{}` syntax for this reason.
Apologies

mzedel (approved):

oleorhagen:
Unrelated test failures - merging

PR #1728, opened 2021-11-10
staging: Update from master + fix deviceauth-enterprise setup

lluiscampos:
Update from latest master and:

* staging: Fix deviceauth-enterprise setup
    * component-maps.yml: Set release_component to false in OS flavor
    * git-versions.yml: Comment out the OS flavor
    * docker-compose files: use correct image for mender-device-auth

* staging: Set correct git version for devicemonitor

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -78,8 +78,25 @@ test:docs:
     - git config --global user.email "user@example.com"
     - git clone --depth=1 https://github.com/mendersoftware/mender-docs.git mender-docs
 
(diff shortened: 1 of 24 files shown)


lluiscampos:
@alfrunes I believe this should fix the issues you had with release_tool in the last HM release. Otherwise please create a ticket to follow up so that we don't end up doing again a manually tagged release

lluiscampos:
I need to fix relese_tool tests first. I'll re-submit soon

PR #1731, opened 2021-11-10
Revert "release_tool: Add the option to version the components-map file"

oleorhagen:
Reverts mendersoftware/integration#1710

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -77,14 +77,9 @@ class Component:
     name = None
     type = None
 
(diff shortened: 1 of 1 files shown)


lluiscampos:
@oleorhagen Give me few minutes before merging this. I might have found the solution

lluiscampos (commented):
Instead of reverting, I think this should fix it: https://github.com/mendersoftware/integration/pull/1732

PR #1732, opened 2021-11-10
release_tool: Set remote integration version for `--list` option

lluiscampos:
Extend implementation from 2ae692c to also cover `--list`


oleorhagen (approved):
:tada:

PR #1737, opened 2021-11-11
verify-docker-versions: Change the docker-compose version check

oleorhagen:
Previously we relied on the --version flag, gnu style, but this is actually
inconsistent across implementations, like this is the output from the
tiangolo/docker-with-compose container:

```
➜  LH mender-docs git:(d30b120b) ✗ docker-compose --version --short
2.1.1
➜  LH mender-docs git:(d30b120b) ✗ docker run --rm tiangolo/docker-with-compose docker-compose --version --short
docker-compose version 1.29.2, build unknown
```

Therefore, when moving to the subcommand style, it is consistent once again.

Changelog: Unify the docker-compose version check through using the
docker-compose version subcommand, as opposed to the --version flag when
checking which docker-compose version is compatible with the current demo script.


mzedel (approved):
🏆

PR #1736, opened 2021-11-11
pipeline: test:docs: Do a full clone of mender-docs repository

lluiscampos:
So that we can checkout any other branch afterwards

Amends commit d5d3bf2.

PR #1741, opened 2021-11-12
staging: Update from master

lluiscampos:
The main motivation is to bring #1736.


lluiscampos:
Pinging @tranchitella as stand-in Build Master. This PR should get the rid of one of the problems in integration/staging pipeline. The other one I need to follow-up independently.

tranchitella (approved):

PR #1752, opened 2021-11-17
MEN-5024: Check that migration to new server works.

kacf:
Plus fix the existing test which passed incorrectly.


kacf:

kacf:
Both failures are in unrelated backend tests.

lluiscampos (approved):
The second commit looks good :+1: 

The first one is very hard for me to review... but "it makes sense".

Very good catch, by the way. Thank you.

kacf:
> The first one is very hard for me to review... but "it makes sense".

Yes, it's essentially a rewrite, or "re-copy". I guess that if something's wrong there, I can just blame one of the backend devs! 🙂

PR #1753, opened 2021-11-17
MEN-5203: mender-client-docker-addons: Install demo cert in root

lluiscampos:
This allows mender-connect (and every other addon) to connect to the
demo server without additional settings.


mzedel (approved):
thanks for including the reference to the meta-mender recipe - based on this, the change here looks good to me!

PR #1777, opened 2021-11-30
MEN-5216: test_monitor_client: Restart mender-client after network loss

lluiscampos:
In the context of MEN-5216, the actual API calls to the Mender servers
are done with the local proxy via Mender client.

Restart the service so that the client needs to resolve again
docker.mender.io and then gets the connection refused, correctly
simulating the offline scenarios of these tests.


lluiscampos:
These go together:
* https://github.com/mendersoftware/mender/pull/898
* https://github.com/mendersoftware/monitor-client/pull/100
* https://github.com/mendersoftware/integration/pull/1777

merlin-northern (approved):
aha! so that was the trick!

PR #1786, opened 2021-12-09
mender-client-docker-addons: Update base image to latest Ubuntu

lluiscampos:
Ubuntu 20.04 LTS has golang 1.13, which is getting outdated. One of the
problems using it is that the vendor folder is ignored and the
dependencies are being downloaded from upstream. Ubuntu 21.10 comes with
golang 1.17 and respects the vendor folder.


kacf (approved):

lluiscampos:
Interestingly, the two integration tests that we have that use `mendersoftware/mender-client-docker` image failed in CI (the test framework couldn't SSH into the container) but they pass locally with the same image (downloaded from the CI pipeline and loaded locally) :thinking:

PR #1787, opened 2021-12-10
MEN-5272: Add test_monitorclient_alert_store_discard_http_400

lluiscampos:
This test verifies MEN-5272 by recreating an offline scenario where the
    store gets corrupted for some reason. Then the server will reply HTTP
    400, the device will discard the alerts, but further alerts will still
    be sent.

Bonus:

    test_monitor_client.py: Refactor and clean-up
    Namely:
    * Moved some constants to the global scope,
    * refactored getting and parsing the e-mail,
    * refactored asserting on the email messages,
    * moved the "env ready" message to the prepare method
    * and added some missing colons missing in the logging.

### DIFF
--- a/tests/tests/test_monitor_client.py
+++ b/tests/tests/test_monitor_client.py
@@ -1279,3 +1279,150 @@ class TestMonitorClientEnterprise:
                     found = True
                     break
             assert found
+
+    def test_monitorclient_alert_store_discard_http_400(
+        self, monitor_commercial_setup_no_client
+    ):
+        """Tests that malformed alerts in the store (HTTP 400) are discarded"""
+        service_name = "crond"
+        user_name = "some.user+{}@example.com".format(str(uuid.uuid4()))
+        devid, _, _, mender_device = self.prepare_env(
+            monitor_commercial_setup_no_client, user_name
+        )
+
+        prepare_service_monitoring(mender_device, service_name)
+        time.sleep(2 * wait_for_alert_interval_s)
+
+        logger.info(
+            "test_monitorclient_alert_store_discard_http_400: disabling access to docker.mender.io (point to localhost in /etc/hosts)"
+        )
+        mender_device.run("sed -i.backup -e '$a127.2.0.1 docker.mender.io' /etc/hosts")
+        mender_device.run("systemctl restart mender-client")
+        mender_device.run("systemctl stop %s" % service_name)
+        logger.info(
+            "Stopped %s, sleeping %ds." % (service_name, wait_for_alert_interval_s)
+        )
+        time.sleep(wait_for_alert_interval_s)
+        mender_device.run("systemctl start %s" % service_name)
+        logger.info(
+            "Started %s, sleeping %ds" % (service_name, wait_for_alert_interval_s)
+        )
+
+        time.sleep(2 * wait_for_alert_interval_s)
+
+        _, messages = get_and_parse_email(monitor_commercial_setup_no_client)
+        assert len(messages) == 0
+        logger.info(
+            "test_monitorclient_alert_store_discard_http_400: got no alerts, device is offline."
+        )
+
+        logger.info(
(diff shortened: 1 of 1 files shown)


tests/tests/test_monitor_client.py
```diff
@@ -1279,3 +1279,128 @@ def test_monitorclient_send_configuration_data(
...
+        )
+        mender_device.run(
+            "bash -c '"
+            "cd /usr/share/mender-monitor;"
+            ". lib/fixlenstore-lib.sh;"
+            'fixlenstore_put "invalid json break here";'
+            'fixlenstore_put "something else there";'
+            "'"
+        )
+
```
merlin-northern, line 1330: uno momentito, por favor.
the scenario that will convince me:
1. put device off line
2. generate some alerts, by stopping cron
3. add garbage to the store
4. verify with `fixlenstore_count` that we have somehting in the store, advanced level: use mdb_dump or peek from fixlength store to verify that your garbage is there
5. put the device online
6. wait a bit, remember that we try to send the store every minute (you can change that with exporting `ALERT_STORE_RESEND_INTERVAL` and restarting the daemon with it)
7. verify that the alerts were not sent, because one of them was causing 400, so we should get 0 emails
8. verify that alert store is empty, as we are cleaning it on 400

merlin-northern: the fact that we get some alert after all that, as you test now, is a bonus.

tests/tests/test_monitor_client.py
```diff
@@ -250,6 +278,8 @@ def prepare_env(self, env, user_name):
         devid = devices[0]
         authtoken = auth.get_auth_token()
 
+        logger.info("%s: env ready.", inspect.stack()[1].function)
```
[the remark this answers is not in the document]
lluiscampos: I don't think this check is necessary... this method can only be called within this class, and won't make sense to be called out of a test... and if someone writes code that calls this method in a hacky way from main, I rather make the code fail and force that hacker to re-think his/her design... :eye_speech_bubble:

tests/tests/test_monitor_client.py
lluiscampos: > uno momentito, por favor. the scenario that will convince me:
> 
>     1. put device off line
> 
>     2. generate some alerts, by stopping cron
> 
>     3. add garbage to the store
> 
>     4. verify with `fixlenstore_count` that we have somehting in the store, advanced level: use mdb_dump or peek from fixlength store to verify that your garbage is there

I actually had done this while developing the test, but on final revision I thought that calling `fixlenstore_count` as part of the verification didn't belong on a integration level test.

But I'll add it now to make you happy.

> 
>     5. put the device online
> 
>     6. wait a bit, remember that we try to send the store every minute (you can change that with exporting `ALERT_STORE_RESEND_INTERVAL` and restarting the daemon with it)

This test waits the exact same time than `test_monitorclient_alert_store` to **not** receive the alerts, while the former with the same time expects to get the alerts. I think that should be the criteria: that the same scenario in one case the alerts have been received and in the other have been lost

> 
>     7. verify that the alerts were not sent, because one of them was causing 400, so we should get 0 emails

Exactly as it is done. Right?

> 
>     8. verify that alert store is empty, as we are cleaning it on 400

I will add this to make you happy, same as point 4. But again I think that using `fixlenstore_count` does not belong on an integration test.

> the fact that we get some alert after all that, as you test now, is a bonus.

I disagree. The fact that some alerts can be send after corrupting the store is essential to verify that the device doesn't get locked out in an state that no more alerts will be sent.

merlin-northern (approved):
lovely! and thank you for making me happy :*

tests/tests/test_monitor_client.py
merlin-northern: thank you for applying so many of my comments! have I mentioned it is a pleasure working with you?

PR #1798, opened 2021-12-13
tests: Introduce helper new_tenant_client between tests and env managers

lluiscampos:
The story behind this is that we have a hook in conftest to inspect test
environment in search for MenderDevice/MenderDeviceGroup instances. When
found, then we can print systemd logs from the Mender client(s).

This works fine for all OS setups, as common_setup fixtures uses this
convention of attaching the device to the environment. But on Enterprise
setups this is test specific, so it was forgotten most of the times.

Now this common.py level helper sits in the middle of the tests and the
env managers, so that it can create the Mender client container and
attach the MenderDevice instance to the test env (so that, yey, the logs
can be printed).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -193,10 +193,21 @@ def pytest_exception_interact(node, call, report):
         if device is None:
             logger.info("Could not find device in test environment, no printing logs")
         else:
+
+            def print_device_output(instance, output):
+                """Log command output of either a MenderDevice or a MenderDeviceGroup."""
+                if isinstance(instance, MenderDevice):
+                    logger.info(output)
+                else:
+                    for dev, log in output.items():
+                        logger.info("Printing log of device %s", dev)
+                        logger.info(log)
+
             try:
                 logger.info("Printing client deployment log, if possible:")
                 output = device.run("cat /data/mender/deployment*.log || true", wait=60)
-                logger.info(output)
+                print_device_output(device, output)
+
             except:
                 logger.info("Not able to print client deployment log")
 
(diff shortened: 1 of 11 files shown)


lluiscampos:
> @mender-test-bot start pipeline

:repeat_one:

tests/conftest.py
```diff
@@ -193,10 +193,21 @@ def pytest_exception_interact(node, call, report):
...
             logger.info("Could not find device in test environment, no printing logs")
         else:
+
+            def print_device_output(instance, output):
+                """Log command output of either a MenderDevice or a MenderDeviceGroup."""
+                if isinstance(instance, MenderDevice):
+                    logger.info(output)
+                else:
+                    for dev, log in output.items():
+                        logger.info("Device %s", dev)
```
kacf, line 203: Suggestion: This is the kind of thing one will often search for in the log. Maybe it should be a bit more unique? Maybe `Printing log of device %s`?

kacf (approved):
:heart: :heart: :heart:

tests/conftest.py
```diff
@@ -193,10 +193,21 @@ def pytest_exception_interact(node, call, report):
         if device is None:
             logger.info("Could not find device in test environment, no printing logs")
         else:
+
+            def print_device_output(instance, output):
+                """Log command output of either a MenderDevice or a MenderDeviceGroup."""
+                if isinstance(instance, MenderDevice):
+                    logger.info(output)
```
oleorhagen, line 200: Q:
Why  don't we need a header for the client?

oleorhagen (approved):
Real nice!

tests/conftest.py
kacf: There are headers in the places where this function is called. Having a header below is just to tell multiple devices apart.

tests/conftest.py
lluiscampos: Good suggestion, thanks

tests/conftest.py
oleorhagen: Thanks

MaciejTe (approved):

lluiscampos:
Failure in `TestAccessEnterprise::test_upgrades` in unrelated. Last force-push only changed a detail in the log entry. Merging!

PR #1825, opened 2021-12-30
MEN-5344: adjust Azure integration tests to the new implementation

tranchitella:
Update the implementation of the backend integration tests for the Azure
integration. Also, add new tests for the device state management (Device
Twin in IoT Hub).

Changelog: none

Depends on: https://github.com/mendersoftware/iot-manager/pull/13

### DIFF
--- a/backend-tests/tests/test_azure.py
+++ b/backend-tests/tests/test_azure.py
@@ -12,6 +12,7 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+from typing import Dict
 from typing import Optional
 import pytest
 import logging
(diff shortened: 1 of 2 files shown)


backend-tests/tests/test_azure.py
```diff
@@ -12,21 +12,20 @@
...
 import ssl
 import uuid
-from base64 import b64decode
-from typing import Dict
-from unittest.mock import Mock, MagicMock, PropertyMock
-
 
 import trustme
+
+from base64 import b64decode
```
MaciejTe, line 24: From: https://www.python.org/dev/peps/pep-0008/#imports
```
Imports should be grouped in the following order:

    1. Standard library imports.
    2. Related third party imports.
    3. Local application/library specific imports.
```

kjaskiewiczz (approved):
looks good

tranchitella:
Ignoring `test_upgrades`

PR #1827, opened 2021-12-31
backend tests: perpare test to work with existing tenants

kjaskiewiczz:
related task: https://tracker.mender.io/browse/QA-324

### DIFF
--- a/backend-tests/tests/test_artifact.py
+++ b/backend-tests/tests/test_artifact.py
@@ -548,8 +548,11 @@ class TestUploadArtifactEnterprise:
             auth_token = self.get_auth_token(user.name, user.pwd)
 
             # create and upload the mender artifact
+            tenant.test_artifact_name = user.name.translate(
+                {ord(c): None for c in ".@+"}
+            )
             with get_mender_artifact(
-                artifact_name=user.name,
+                artifact_name=tenant.test_artifact_name,
                 device_types=["arm1"],
                 depends=("key1:value1", "key2:value2"),
                 provides=("key3:value3", "key4:value4", "key5:value5"),
@@ -576,12 +579,13 @@ class TestUploadArtifactEnterprise:
             user = tenant.users[0]
             auth_token = self.get_auth_token(user.name, user.pwd)
             api_client.with_auth(auth_token)
-            r = api_client.call("GET", deployments.URL_DEPLOYMENTS_ARTIFACTS)
+            r = api_client.call(
+                "GET",
+                deployments.URL_DEPLOYMENTS_ARTIFACTS
+                + "?name="
+                + tenant.test_artifact_name,
+            )
             assert r.status_code == 200
-            artifacts = [
-                artifact
-                for artifact in r.json()
-                if not artifact["name"].startswith("mender-demo-artifact")
-            ]
+            artifacts = r.json()
             assert len(artifacts) == 1
-            assert artifacts[0]["name"] == user.name
+            assert artifacts[0]["name"] == tenant.test_artifact_name
--- a/backend-tests/tests/test_deployments.py
+++ b/backend-tests/tests/test_deployments.py
@@ -39,6 +39,7 @@ from testutils.common import (
     make_accepted_devices,
     make_device_with_inventory,
     submit_inventory,
+    useExistingTenant,
 )
 from testutils.infra.container_manager.kubernetes_manager import isK8S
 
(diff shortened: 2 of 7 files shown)


kjaskiewiczz:

kjaskiewiczz:

kjaskiewiczz:

kjaskiewiczz:

backend-tests/tests/test_artifact.py
```diff
@@ -549,7 +549,7 @@ def test_artifacts_exclusive_to_user(self, mongo, clean_mongo):
 
             # create and upload the mender artifact
             with get_mender_artifact(
-                artifact_name=user.name,
+                artifact_name=user.name.translate({ord(c): None for c in ".@+"}),
```
tranchitella, line 552: You could store this value in a variable and reuse it.

backend-tests/tests/test_deployments.py
```diff
@@ -89,10 +89,13 @@ def create_tenant_test_setup(
...
-    # Verify that the 'nr_deployments' expected deployments have been created
-    resp = api_mgmt_deploy.with_auth(user.utoken).call("GET", "/deployments")
-    assert resp.status_code == 200
-    assert len(resp.json()) == nr_deployments
+
+    if not isK8S():
+        # Verify that the 'nr_deployments' expected deployments have been created
+        resp = api_mgmt_deploy.with_auth(user.utoken).call("GET", "/deployments")
+        assert resp.status_code == 200
+        assert len(resp.json()) == nr_deployments
```
tranchitella, line 97: Just to understand: would it be possible to get the "count" of deployments before running the test, then getting the "count" after it, and test if the difference is `== nr_deployments`. This applies to all asserts that are now conditional if `not isK8S()`.

backend-tests/tests/test_devauth.py
```diff
@@ -833,6 +851,9 @@ def test_device_count(self, devs_authsets, user):
 
 
 class TestDeviceMgmtEnterprise(TestDeviceMgmtBase):
+    @pytest.mark.skipif(
+        isK8S(), reason="not feasible to test with existing tenant",
+    )
```
tranchitella, line 856: I'm not 100% sure it makes sense to skip thesee tests if `isK8S()` is True.
We are running these tests in staging today, without reusing existing tenants.
Maybe we should mark it differently, based on an environment variable we'll set when running in "existing tenant" mode?

backend-tests/tests/test_deployments.py
kjaskiewiczz: In case of deployments, it should be possible, because the endpoint for listing deployments returns `X-Total-Count`.
The endpoint for listing devices in the deviceauth doesn't have this functionality, but there is a separate endpoint for counting devices, so we can use it instead. I'll check rest of the cases.
There are places where we're not only checking number of the results, but also going through all the results. I'd still skip those cases when using existing tenant.

backend-tests/tests/test_devauth.py
kjaskiewiczz: Using different variable make sense; I'll update the PR.

backend-tests/tests/test_artifact.py
kjaskiewiczz: done

backend-tests/tests/test_deployments.py
kjaskiewiczz: done where it was enough to count documents before and after the test

backend-tests/tests/test_devauth.py
kjaskiewiczz: done

tranchitella (approved):

tranchitella:

backend-tests/tests/test_artifact.py
```diff
@@ -576,12 +579,12 @@ def test_artifacts_exclusive_to_user(self, mongo, clean_mongo):
...
-                for artifact in r.json()
-                if not artifact["name"].startswith("mender-demo-artifact")
-            ]
+            r = api_client.call(
+                "GET",
+                deployments.URL_DEPLOYMENTS_ARTIFACTS
+                + "?name="
+                + tenant.test_artifact_name,
+            )
+            artifacts = r.json()
```
alfrunes, line 588:
```suggestion
            assert r.status_code == 200
            artifacts = r.json()
```

alfrunes (approved):

backend-tests/tests/test_artifact.py
kjaskiewiczz: fixed

kjaskiewiczz:

MaciejTe (approved):

PR #1836, opened 2022-01-05
Many fixes and additions to support QA-310.

kacf:
https://tracker.mender.io/browse/QA-310

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -1196,6 +1196,10 @@ def find_prev_version(tag_list, version):
     """Finds the highest version in tag_list which is less than version.
     tag_list is expected to be sorted with highest version first."""
 
+    if version == "master" and len(tag_list) > 0:
+        # For master, return the newest released version.
+        return tag_list[0]
+
     try:
         (
             version_major,
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -3164,10 +3164,11 @@ def main():
         "--in-integration-version",
         dest="in_integration_version",
         metavar="VERSION",
-        help="Used together with the `-g` argument to query for a version of a "
+        help="Used together with the `--version-of` argument to query for a version of a "
         + "service which is in the given version of integration, instead of the "
         + "currently checked out version of integration. If a range is given here "
-        + "it will return the range of the corresponding service. It is also used together with the --generate-release-notes argument.",
+        + "it will return the range of the corresponding service. "
+        + "It is also used together with the `--generate-release-notes` argument.",
     )
     parser.add_argument(
         "-s",
@@ -3188,7 +3189,7 @@ def main():
         "--feature-branches",
         action="store_true",
         default=False,
-        help="When used with -f, include upstream feature branches",
+        help="When used with `--integration-versions-including`, include upstream feature branches",
     )
     parser.add_argument(
         "-v",
@@ -3210,7 +3211,7 @@ def main():
         dest="pr",
         metavar="REPO/PR-NUMBER",
         action="append",
-        help="Can only be used with -b. Specifies a repository and pull request number "
+        help="Can only be used with `--build`. Specifies a repository and pull request number "
         + "that should be triggered with the rest of the repository versions. It is "
         + "also possible to specify a branch name instead of a pull request number. "
         + "May be specified more than once.",
@@ -3231,27 +3232,27 @@ def main():
         "--all",
         action="store_true",
         default=False,
-        help="When used with -l, list all repositories, including optional ones. "
-        + "When used with -f, include local branches in addition to upstream branches.",
+        help="When used with `--list`, list all repositories, including optional ones. "
+        + "When used with `--integration-versions-including`, include local branches in addition to upstream branches.",
     )
     parser.add_argument(
         "--only-backend",
         action="store_true",
         default=False,
-        help="When used with -l, list only backend repositories; ignored otherwise",
+        help="When used with `--list`, list only backend repositories; ignored otherwise",
     )
     parser.add_argument(
         "--only-client",
         action="store_true",
         default=False,
-        help="When used with -l, list only non-backend repositories; ignored otherwise",
+        help="When used with `--list`, list only non-backend repositories; ignored otherwise",
     )
     parser.add_argument(
         "--list-format",
         metavar="simple|table|json",
         default="simple",
         nargs="?",
-        help="When used with -l, 'simple' prints only the repos, 'table' adds the versions "
+        help="When used with `--list`, 'simple' prints only the repos, 'table' adds the versions "
         + "of each component in a table format, and 'json' composes a json object",
     )
     parser.add_argument(
@@ -3291,7 +3292,8 @@ def main():
     parser.add_argument(
         "--generate-release-notes",
         action="store_true",
-        help="Generate changelogs and statistics and put them in `release_notes_*.txt` files. Use -i argument to choose which integration range to generate notes for.",
+        help="Generate changelogs and statistics and put them in `release_notes_*.txt` files. "
+        + "Use `--in-integration-version` argument to choose which integration range to generate notes for.",
     )
     args = parser.parse_args()
 
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -2450,8 +2450,9 @@ def run_and_combine_statistics_and_changelog(
         workdir if single_repo else integration_dir(),
         [
             "for-each-ref",
-            r"--format=%(taggerdate:format:%m.%d.%Y)",
+            r"--format=%(creatordate:format:%m.%d.%Y)",
             "refs/tags/" + to_version,
+            "refs/heads/" + to_version,
         ],
         capture=True,
     )
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -3205,7 +3205,7 @@ def main():
         metavar="simple|table|json",
         default="simple",
         nargs="?",
-        help="When used with -l, 'simple' prints only the repos, 'table' adds the versions"
+        help="When used with -l, 'simple' prints only the repos, 'table' adds the versions "
         + "of each component in a table format, and 'json' composes a json object",
     )
     parser.add_argument(
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -41,14 +41,6 @@ os.environ["GIT_PAGER"] = "cat"
 # The repositories are indexed by their Git repository names.
 RELEASE_TOOL_STATE = None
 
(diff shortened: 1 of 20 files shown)


kacf:
Updated! I had to rewrite the test a little bit to work on Gitlab.

The `TEST_RELEASE_TOOL_LIST_OPEN_SOURCE_ONLY` variable is a bit evil. However I couldn't find a good way without significantly more effort, because the `release_tool.py --list` command is invoked several times as sub-commands, so I can't just pass in an option. Let me know if you have any better ideas than what I proposed.

kacf:

extra/statistics-generator
```diff
@@ -112,14 +112,34 @@ collect_changes() {
 
 post_process() {
     sed -nre '
+:start
+# If empty line.
```
oleorhagen, line 116: SO MUCH YES!

extra/release_tool.py
```diff
@@ -2460,17 +2460,30 @@ def log_and_run_cmd(cmd, workdir):
...
 
 
-def do_generate_release_notes(state):
-    """Generate Release Notes for the released tags"""
+def do_generate_release_notes(
+    base_dir, version_of_integration, prev_of_integration=None
+):
+    split = version_of_integration.split("..")
+    if len(split) > 1:
+        if prev_of_integration is not None:
```
oleorhagen, line 2484: Suggestion:
```
if prev_of_integration is not None
```
=
```
if prev_of_integration
```

extra/release_tool.py
```diff
@@ -2460,17 +2460,30 @@ def log_and_run_cmd(cmd, workdir):
         fd.write(changelog_out.decode())
 
 
-def do_generate_release_notes(state):
-    """Generate Release Notes for the released tags"""
```
oleorhagen, line 2612: Q:
Why remove the comment all together?

extra/release_tool.py
```diff
@@ -2460,17 +2460,30 @@ def log_and_run_cmd(cmd, workdir):
...
+def do_generate_release_notes(
+    base_dir, version_of_integration, prev_of_integration=None
+):
+    split = version_of_integration.split("..")
+    if len(split) > 1:
+        if prev_of_integration is not None:
+            raise Exception(
+                "version_of_integration is a range, and prev_of_integration is non-empty. This should not happen!"
+            )
+        prev_of_integration = split[0]
```
oleorhagen, line 2488: Tricks:
`prev_of_integration, version_of_integration = split[0], split[1]`
But splitting hairs ofc :)

extra/release_tool.py
```diff
@@ -2460,17 +2460,30 @@ def log_and_run_cmd(cmd, workdir):
...
+        version_of_integration = split[1]
+
+    Component.set_integration_version(version_of_integration)
+
+    if prev_of_integration is None:
```
oleorhagen, line 2493: Same as above.

extra/release_tool.py
```diff
@@ -3087,10 +3107,10 @@ def main():
         "--in-integration-version",
         dest="in_integration_version",
         metavar="VERSION",
-        help="Used together with the above argument to query for a version of a "
+        help="Used together with the `-g` argument to query for a version of a "
```
oleorhagen, line 3126: Suggestion:
Use the long option in the help text?
:)

extra/release_tool.py
```diff
@@ -3087,10 +3107,10 @@ def main():
...
         + "service which is in the given version of integration, instead of the "
         + "currently checked out version of integration. If a range is given here "
-        + "it will return the range of the corresponding service.",
+        + "it will return the range of the corresponding service. It is also used together with the --generate-release-notes argument.",
```
oleorhagen, line 3129: Nit:
quote --generate-releasenotes endquote ?

extra/statistics-generator
```diff
@@ -63,7 +63,7 @@ collect_changes() {
     else
         # Sorting is important because we need "*-enterprise" repositories to
         # come after their Open Source counterparts (see below).
-        REPO_LIST="$("$RELEASE_TOOL" --list git --only-backend | sort)"
+        REPO_LIST="$("$RELEASE_TOOL" --list git --in-integration-version $1 --only-backend | sort)"
         REPO_LIST="$REPO_LIST integration"
```
oleorhagen, line 67: Question:
In newer branches `integration` will be present twice now, is this a problem?

oleorhagen (approved):
Oh my lord! 

What a pleasure!

extra/release_tool.py
kacf: Brainfart. I have no idea why I removed it, I will put it back.

extra/release_tool.py
kacf: Strictly speaking it doesn't matter in this case, but I would like it to match the default argument, and the lower variant will match `""` and `0` as well.

extra/statistics-generator
kacf: `release_tool` doesn't list itself (integration), so this is correct, actually. But you're right that this is a gotcha, I'm not sure if I've taken this into account every time I've written a piece of code which loops over this.

extra/release_tool.py
kacf: It's pre-existing, but worth a cleanup anyway, enjoy https://github.com/mendersoftware/integration/pull/1836/commits/95ad4e735d036e20bbb88d8afe00bfc2841ef263!

kacf:

extra/release_tool.py
```diff
@@ -263,6 +272,9 @@ def is_independent_component(self):
 BACKEND_SERVICES = (
     BACKEND_SERVICES_OPEN | BACKEND_SERVICES_ENT | BACKEND_SERVICES_OPEN_ENT
 )
+CLIENT_SERVICES_ENT = {
```
lluiscampos, line 275: Is this strictly necessary? :cry: 

Update note in [point 3 here](https://github.com/mendersoftware/mender-qa/blob/master/Documentation/adding-new-repository.md#integration) so that it is followed for all repos (not only backend repos)

extra/test/release_notes_mender-artifact.txt
```diff
@@ -0,0 +1,10 @@
+## mender-artifact 3.6.1
+
+_Released 09.28.2021_
+
+### Statistics
+
+A total of 0 lines added, 0 removed (delta 0)
+
+### Changelogs
+
```
lluiscampos, line 10: These (here but also mender-cli and mender-connect) do not look correct to me. There must be statistics and changelogs.

extra/statistics-generator
kacf: I now came across this again, and let me make a small modification: It doesn't list itself **when using `--only-backend`**. We can debate whether this is correct I suppose, but I'm not going to change it now. If we do decide to change it later, this argument is pretty easy to search for and change everywhere.

extra/release_tool.py
kacf: Thanks, I'll do that!

For whether it's necessary, see [this comment](https://github.com/mendersoftware/integration/pull/1836#issuecomment-1005715282). I welcome other approaches, this part is tricky because it is invoked so many times as a sub command.

extra/test/release_notes_mender-artifact.txt
kacf: That depends. What you mean is that this file should be removed actually, since there are no changes. This makes sense for releases, where it's a manual process, and we just won't paste it if it says "0 lines".

But [QA-310](https://tracker.mender.io/browse/QA-310) says to "completely replace" what's there, which I take to mean that it shouldn't try to mix it in with the content which is already in mender-docs-changelog, so if nothing is generated, this component won't even exist in the menu. We could try to mix it in with the pre-existing text, but it's more error prone. What do you think?

kacf: It's good that you pointed it out though, because it looks like they *always* are zero, even when they shouldn't be. I'll fix that, but my point above still stands.

lluiscampos: Yes, what I meant is that this specific release, `mender-artifact 3.6.1` cannot be blank.

kacf: No, because mender-artifact in integration 3.0.1 is 3.6.1, and in integration 3.1.0, it's also 3.6.1. So the queried range is `3.6.1..3.6.1`, in other words empty. What I could do is to change the test to test integration `3.0.0..3.1.0` instead of `3.0.1..3.1.0`. This diff would then not be empty anymore, but some of the others still would be, so this might be a test with better coverage of the corner cases. I think I'll do that.

kacf: Done! See https://github.com/mendersoftware/integration/pull/1836/commits/0f34e8573b3937ae3ee1ed462806368b0f30e122.

Note that mender-cli is still empty, as it should be.

lluiscampos: I understand it now! Thanks.

Yes, that range seems to be better for test purposes.

lluiscampos (approved):

extra/test_release_tool.py
```diff
@@ -142,25 +142,67 @@ def test_version_of(capsys):
...
+    # multiple containers and we don't know which one to choose.
     run_main_assert_result(capsys, ["--version-of", "mender"], "master")
+    try:
+        run_main_assert_result(
+            capsys, ["--version-of", "mender", "--version-type", "docker"], "master"
+        )
+    except:
+        pass
+    else:
+        pytest.fail("Expected exception!")
```
lluiscampos, line 155: An alternative using [pytest functionality](https://docs.pytest.org/en/6.2.x/reference.html#pytest.raises) would be:
```
with pytest.raises(RuntimeError):
    run_main_assert_result(
        capsys, ["--version-of", "mender", "--version-type", "docker"], "master"
    )
```

lluiscampos (commented):
Thanks for the human friendly commit messages. The tests in the last commit are also very good to illustrate the changes.

I'll give it a try now pretending to do a new release :+1:

lluiscampos (approved):
> I'll give it a try now pretending to do a new release +1

It passed my test :heavy_check_mark:

kacf:

extra/test_release_tool.py
kacf: Thanks, another nugget to remember! Changed.

kacf: And by changed, I mean, I did it locally now, and will apply it together with the test re-enabling.

kacf:
The prebuilt client failure is expected because the image has changed name. This should pass on master after this is merged and the client is rebuilt.

Merging.

PR #1838, opened 2022-01-07
QA-321: Add all components to the release tool

oleorhagen:
This includes support for:

* mender-binary-delta
* mender-convert
* mender-configure-module

Which are independent components, as well as `non-core`.

List them with `relase_tool --list --all`

Changelog: None

### DIFF
--- a/component-maps.yml
+++ b/component-maps.yml
@@ -6,6 +6,28 @@
 
 git:
 
+  mender-convert:
+    docker_image: []
+    docker_container: []
+    release_component: true
+    independent_component: true
+    non_core_component: true
+
+  mender-binary-delta:
+    docker_image: []
+    docker_container: []
+    release_component: true
+    independent_component: true
+    non_core_component: true
+
+  mender-configure-module:
+    docker_image:
+    - mender-client-docker-addons
+    docker_container: []
+    release_component: true
+    independent_component: true
+    non_core_component: true
+
   iot-manager:
     docker_image:
     - iot-manager
@@ -304,6 +326,7 @@ docker_image:
     - integration
     - mender
     - mender-connect
+    - mender-configure-module
     docker_container:
     - mender-client
     release_component: true
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -153,6 +153,7 @@ class Component:
         only_non_release=False,
         only_independent_component=False,
         only_non_independent_component=False,
+        only_core_components=True,
     ):
         Component._initialize_component_maps()
         if only_release is None:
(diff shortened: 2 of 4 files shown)


oleorhagen:
Only tested meakly, by playing around with it locally.

Seems to work.

oleorhagen:
```
$ release_tool --list --all
auditlogs
create-artifact-worker
deployments
deployments-enterprise
deviceadm
deviceauth
deviceauth-enterprise
deviceconfig
deviceconnect
devicemonitor
gui
integration
inventory
inventory-enterprise
iot-manager
mender
mender-api-gateway-docker
mender-artifact
mender-binary-delta <- New
mender-cli
mender-conductor
mender-conductor-enterprise
mender-configure-module <- New
mender-connect
mender-convert <- New
monitor-client
mtls-ambassador
reporting
tenantadm
useradm
useradm-enterprise
workflows
workflows-enterprise
```

component-maps.yml
```diff
@@ -6,6 +6,27 @@
...
+
+  mender-binary-delta:
+    docker_image: []
+    docker_container: []
+    release_component: true
+    independent_component: true
+    non_core_component: true
+
+  mender-configure-module:
+    docker_image: []
```
lluiscampos, line 24: It is part of `mender-client-docker-addons` image

extra/release_tool.py
```diff
@@ -172,6 +173,7 @@ def get_components_of_type(
             is_release_component = Component.COMPONENT_MAPS[type][comp][
                 "release_component"
             ]
+            non_core_component = Component.COMPONENT_MAPS.get(type,{}).get(comp, {}).get("non_core_component", False)
```
lluiscampos, line 176: Nitpick: for consistency with the others I suggest renaming to `is_non_core_component`

extra/release_tool.py
```diff
@@ -153,6 +153,7 @@ def get_components_of_type(
         only_non_release=False,
         only_independent_component=False,
         only_non_independent_component=False,
+        only_core_components=True,
```
lluiscampos, line 156: By looking at all places we had to set `only_core_components=False` to call this, should we reconsider the default value?

lluiscampos (approved):
Looks good! Missing docker image for mender-configure-module and suggesting two internal things. Pre-aproving in any case.

Also:

> Only tested meakly, by playing around with it locally.

On the high level looks good: all parts for release and/or trigger builds use non-core components, while the behavior of cli options is not modified

> ```
> $ release_tool --list --all
> auditlogs
> ...
> deviceadm
> ...
> mender-binary-delta

While this is correct (`all` lists everything) I don't see it is much useful: I can either get core or get all (including deprecated repos!). Not required for this task, but maybe we want a cli option here to list "all active components" or "all Mender components" idk.

extra/release_tool.py
oleorhagen: We could, I do like the explicitness though

lluiscampos: okey, no strong feelings here.

extra/release_tool.py
oleorhagen: I'm probably misunderstanding, but the other's are named:
* `independent_component`
* `release_component`
Where does the `is` come in?

lluiscampos: I meant in the code, the Python variable. Two lines above you have `is_release_component = ...`.

PR #1856, opened 2022-01-17
3.2.x: Updates

oleorhagen:


component-maps.yml
```diff
@@ -249,13 +249,6 @@ git:
     release_component: true
     independent_component: true
 
-  reporting:
```
oleorhagen, line 252: It's this, or reverting 02d734e69d893e8a2757b7132d7a74abe9cee816 fully.

docker-compose.reporting.yml
```diff
@@ -1,21 +1,5 @@
...
-        extends:
-            file: common.yml
-            service: mender-base
-        networks:
-            - mender
-        depends_on:
-            - mender-elasticsearch
-        environment:
-            REPORTING_ELASTICSEARCH_ADDRESSES: "http://mender-elasticsearch:9200"
-
```
lluiscampos, line 18: If we remove the main service here, we should remove the whole file

git-versions.yml
```diff
@@ -31,9 +31,6 @@ services:
     mender-deviceconfig:
         image: mendersoftware/deviceconfig:1.2.x
 
-    mender-reporting:
-        image: mendersoftware/reporting:master
-
```
lluiscampos, line 36: Removing this should be enough, the release tool should play well with the rest as long as we have `release_component: false`. 

So honestly I vote for restoring the rest and treat it just as a component not ready to be released.

docker-compose.reporting.yml
oleorhagen: https://github.com/mendersoftware/integration/pull/1856#discussion_r785735584

Which brings me to: it's fine to revert the whole shebang?

git-versions.yml
oleorhagen: https://github.com/mendersoftware/integration/pull/1856/commits/e535ee8353a475c4d591c9341aee30ead4e7d32c

kacf (approved):

PR #1870, opened 2022-01-19
test: fix Azure tests after MEN-5386

tranchitella:
Changelog: none

### DIFF
--- a/backend-tests/tests/test_azure.py
+++ b/backend-tests/tests/test_azure.py
@@ -117,9 +117,19 @@ class _TestAzureBase:
         # Check for equality by parts:
         # Check that actual properties are a subset of expected integrations
         for part in actual.split(";"):
+            # SharedAccessKey will be masked, with only the first 4 characters visible
+            # and the rest of the string replaced with a place holder. For this reason,
+            # we'll test the first 20 bytes only
+            if part.startswith("SharedAccessKey="):
+                part = part[:20]
             assert part in expected_integration["credentials"]["connection_string"]
         # Check that expected properties are a subset of actual integrations
         for part in expected_integration["credentials"]["connection_string"].split(";"):
+            # SharedAccessKey will be masked, with only the first 4 characters visible
+            # and the rest of the string replaced with a place holder. For this reason,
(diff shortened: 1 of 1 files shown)


tranchitella:
Required by https://github.com/mendersoftware/iot-manager/pull/26

backend-tests/tests/test_azure.py
```diff
@@ -117,9 +117,19 @@ def check_integrations(self, user: User, expected_integration: Dict):
...
+            if part.startswith('SharedAccessKey='):
+                part = part[:20]
             assert part in expected_integration["credentials"]["connection_string"]
         # Check that expected properties are a subset of actual integrations
         for part in expected_integration["credentials"]["connection_string"].split(";"):
+            # SharedAccessKey will be masked, with only the first 4 characters visible
+            # and the rest of the string replaced with a place holder. For this reason,
+            # we'll test the first 20 bytes only
+            if part.startswith('SharedAccessKey='):
+                part = part[:24]
```
alfrunes, line 132:
```suggestion
                part = part[:20]
```

backend-tests/tests/test_azure.py
```diff
@@ -117,9 +117,19 @@ def check_integrations(self, user: User, expected_integration: Dict):
         # Check for equality by parts:
         # Check that actual properties are a subset of expected integrations
         for part in actual.split(";"):
+            # SharedAccessKey will be masked, with only the first 4 characters visible
+            # and the rest of the string replaced with a place holder. For this reason,
+            # we'll test the first 20 bytes only
+            if part.startswith('SharedAccessKey='):
+                part = part[:20]
             assert part in expected_integration["credentials"]["connection_string"]
```
alfrunes, line 125: Doesn't both sides of the assertion require truncating the access key?

tranchitella: no, I don't think so. The portion we take from part should be contained in the string.

alfrunes: Right! I thought that the expected value was a list of the connection string key/value pairs.

alfrunes (approved):

PR #1879, opened 2022-01-20
Fix client components being listed as backend components.

kacf:


kacf:
@lluiscampos, @oleorhagen: Doesn't solve any problems, but worth the cleanup! 🙂

oleorhagen:
Why not fix the typo, and use it?

lluiscampos (commented):
Ahá, so this is why the `--list` is listing everything!

I think we should fix the typo instead. For example we saw today `mender-docs` build failing when querying `mender-configure-module`. And there will probably be other parts that expect the short list (or at least they will run faster with the shorter list).

lluiscampos (changes requested):
Ahá, so this is why the `--list` is listing everything!

I think we should fix the typo instead. For example we saw today `mender-docs` build failing when querying `mender-configure-module`. And there will probably be other parts that expect the short list (or at least they will run faster with the shorter list).

kacf:
> Ok, maybe this is the reason for the changelog mis-rendering also.

Unfortunately it was not related to this.

kacf:
@lluiscampos: After some discussion with @oleorhagen, neither of us can really understand why there needs to be a short list. All those components really are part of the release now.

Since the `non_core_component` change also overlaps quite a bit with my next fix, which is necessary for correct changelog generation, as well as for correct list output, I decided to go back to the previous fix, a full revert. If you really need the `core_component` mechanic to work, then you guys will have to think of something next week. Just be aware that `--list` needs to list all current release components; if you want a shorter list it will need an extra argument, ala `--only-core` or something.

kacf:

lluiscampos (approved):
> @lluiscampos: After some discussion with @oleorhagen, neither of us can really understand why there needs to be a short list. All those components really are part of the release now.
> ...

I have elaborated a bit more in Slack - but the conclusion is that I agree with the full revert of this filtering.

oleorhagen:
@mender-test-bot 
cherry-pick to:
* 3.2.x

oleorhagen:
cherry: https://github.com/mendersoftware/integration/pull/1886

PR #1884, opened 2022-01-22
test: fix test_set_version_of unit test for staging

tranchitella:
Changelog: none


tranchitella:
We don't need a full integration pipeline here.

oleorhagen (approved):

lluiscampos:
Good :+1:

PR #1894, opened 2022-01-28
test(reporting): backend-end tests for the reporting service (MEN-4974)

tranchitella:


tranchitella:

tranchitella:

tranchitella:

lluiscampos:
@tranchitella If you rebase now from master this PR should be good to go!

backend-tests/tests/test_reporting.py
```diff
@@ -0,0 +1,1225 @@
...
+    invd = ApiClient(inventory.URL_DEV)
+
+    for inv in dev_inventories:
+        user = tenant.users[0]
+        utoken = useradmm.call(
+            "POST", useradm.URL_LOGIN, auth=(user.name, user.pwd)
+        ).text
+        assert utoken != ""
+
+        tenant.api_token = utoken
```
alfrunes, line 76: I think this can be moved outside the loop.

backend-tests/tests/test_reporting.py
```diff
@@ -0,0 +1,1225 @@
...
+            {"artifact": "v2", "py3": "3.6", "py2": "2.7", "idx": 5},
+            {"artifact": ["v2", "v3"], "py3": "3.6", "idx": 6},
+            {"artifact": ["v2", "v3"], "py3": "3.7", "idx": 7},
+            {"artifact": ["v2", "v3"], "py3": "3.7", "idx": 8},
+            {"artifact": ["v1", "v2", "v3"], "py3": "3.8", "idx": 9},
+        ],
+    )
+    # sleep a few seconds waiting for the data propagation to the reporting service
+    # and the Elasticsearch indexing to complete
+    time.sleep(reporting.REPORTING_DATA_PROPAGATION_SLEEP_TIME_SECS)
```
alfrunes, line 151: Historically speaking, waiting for a fixed duration is a common source of spurious errors. Does it make sense to instead poll for the given number of devices to reach the ES index with a more relaxed time constraint (e.g. a minute or two)?

alfrunes (approved):
Those reporting tests looked oddly familiar :thinking:
No hard blockers, but if you could turn those delays into slightly more robust poll-based waiting that'd be great :+1:

kjaskiewiczz (approved):

backend-tests/tests/test_reporting.py
tranchitella: I don't know, to be honest I liked the idea of waiting a fixed amount of time; if the reporting service is unable to catch up with the data within N seconds, then it is correct that the test fails, as this affects the UX. Does it then make sense to poll every 500ms up to 4 seconds?

alfrunes: 4 seconds might be loads of time on the local machine, but considering that this test is running on build hosts running multiple containerized jobs in parallel there might be some unexpected hiccups. If this is not the case, then ignore this comment.

tranchitella: The theory says that we should not wait more than 2 seconds (the indexer runs every 1000ms, see [here](https://github.com/mendersoftware/reporting/blob/master/config.yaml#L65)); I doubled it to 4 seconds to be safe. Even under load, if it takes more we have an issue. I'm tempted to leave it as it is. We'll monitor the test results and see how it goes.

backend-tests/tests/test_reporting.py
```diff
@@ -0,0 +1,1225 @@
...
+    devauthm = ApiClient(deviceauth.URL_MGMT)
+    invd = ApiClient(inventory.URL_DEV)
+
+    for inv in dev_inventories:
+        utoken = useradmm.call(
+            "POST", useradm.URL_LOGIN, auth=(user.name, user.pwd)
+        ).text
+        assert utoken != ""
+
+        user.api_token = utoken
```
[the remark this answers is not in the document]
tranchitella: https://github.com/mendersoftware/integration/pull/1894/commits/6f1ca8e8f0e87a827a771bef7ea326f1085899f7

backend-tests/tests/test_reporting.py
tranchitella: https://github.com/mendersoftware/integration/pull/1894/commits/6f1ca8e8f0e87a827a771bef7ea326f1085899f7

tranchitella:

tranchitella:

tranchitella:

tranchitella:
Ignoring failure in `tests/test_update_control.py::TestUpdateControlEnterprise::test_update_control_limit`

PR #1895, opened 2022-01-31
release_tool fixes for mender-reporting-indexer

lluiscampos:


lluiscampos:
@kacf I am not sure why the limitation was put in the first place.

I manually tested the two entry points where I can see that `container` key was being used, `--list container` and `--map git repo container`, and both seem to work just fine.

lluiscampos:
The first commit comes from https://github.com/mendersoftware/integration/pull/1894

component-maps.yml
```diff
@@ -251,7 +251,8 @@ git:
     - reporting
     docker_container:
     - mender-reporting
-    release_component: false
+    - mender-reporting-indexer
+    release_component: true
```
tranchitella, line 255: If you set it as a release component, the image tag will be `mender-master`, isn't it?
This makes the commit you cherry-picked wrong, because I replaced the tags with `master`.

tranchitella:
> @kacf I am not sure why the limitation was put in the first place.
> 
> I manually tested the two entry points where I can see that `container` key was being used, `--list container` and `--map git repo container`, and both seem to work just fine.

Same here, I tried and it worked fine.

component-maps.yml
lluiscampos: That I didn't completely understand. The image tag should actually be `mender-master` regardless of the component being enabled or disabled for the release :shrug:

Why was disabled in the first place? I think we should have it enabled in master and if it happens that on Mender 3.3 time we are  not ready for any reason we disabled it then o the release branch.

lluiscampos: If you agree with being enabled, I'll just amend the commit to have `mender-master` tags

tranchitella: @lluiscampos yes I agree we need to release it. Regarding `mender-master` vs. `master`, please see the latest pipeline for the `master` branch here: https://gitlab.com/Northern.tech/Mender/reporting/-/pipelines/457570643

The `publish:image:mender` job is here: https://gitlab.com/Northern.tech/Mender/reporting/-/jobs/2022678264, and as you can see, it exits with:
```
Repository reporting version master is not part of any Mender release. Exiting
$ exit 0
```

Thus, no `mender-master` image gets pushed. Am I wrong?

lluiscampos: Ah, I see. Makes sense, it needs to be part of one or more "Mender" releases to have these tags. I got wrong the logic.

tranchitella (approved):

kacf (approved):
I guess we support N-to-N in all directions now!

PR #1897, opened 2022-01-31
test: Add a test for testing rate-limiting the client

oleorhagen:
Changelog: None

### DIFF
--- a/tests/tests/test_update_control.py
+++ b/tests/tests/test_update_control.py
@@ -16,14 +16,23 @@
 import json
 import os
 import tempfile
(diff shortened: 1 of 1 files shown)


tranchitella (approved):

tranchitella:
> Fails on 3.2.0 locally.
> Do you reckon it's really needed to run a legacy client in the test here?

In my opinion, we don't need to run the legacy client.

oleorhagen:
3.2.x cherry: https://github.com/mendersoftware/integration/pull/1900

PR #1905, opened 2022-02-02
fix: run `build:docker` unless running staging tests

tranchitella:
Changelog: none

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -262,16 +262,27 @@ build:docker:
   rules:
     - if: $RUN_TESTS_STAGING == "true"
       when: never
+    - when: on_success
   variables:
     DOCKER_REPOSITORY: mendersoftware/mender-client-docker-addons
     DOCKER_DIR: extra/mender-client-docker-addons
 
 publish:image:
+  rules:
+    - if: $RUN_TESTS_STAGING == "true"
+      when: never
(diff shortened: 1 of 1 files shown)


lluiscampos (approved):
It looks good.

.gitlab-ci.yml
```diff
@@ -262,16 +262,25 @@ build:docker:
...
+    - when: on_success
   variables:
     DOCKER_REPOSITORY: mendersoftware/mender-client-docker-addons
     DOCKER_DIR: extra/mender-client-docker-addons
 
 publish:image:
+  rules:
+    - if: $RUN_TESTS_STAGING == "true"
+      when: never
+    - when: on_success
```
lluiscampos, line 274: The fallback rule here (and below) should be [this line](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-docker-build.yml#L95),

lluiscampos (changes requested):
The publish images cannot be run always but only on special branches. See comment below.

.gitlab-ci.yml
tranchitella: Force-pushed

lluiscampos (approved):

tranchitella:
Merging despite the failure, which is unrelated to this PR.

PR #1906, opened 2022-02-02
release_tool.py: Extend the handling of reporting service for 3.2.x

lluiscampos:
Latest tag 3.2.1 has the same problem as 3.2.0 because it came from the
same release branch.

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -534,7 +534,7 @@ def version_specific_docker_compose_data_patching(data, rev):
             "version": "1.0.%s" % patch,
         }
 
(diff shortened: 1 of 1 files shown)


oleorhagen (approved):

lluiscampos:
This fix is [not enough](https://gitlab.com/Northern.tech/Mender/integration/-/jobs/2045185750). I'll leave it to @kacf ...

kacf:
I'm taking a look now..

kacf:
Argh, I forgot to fix the underlying problem in 3.2.x. I have fixed it now in #1912, and this workaround has been repeated in #1911.

PR #1911, opened 2022-02-04
release_tool: Repeat reporting fix for 3.2.1.

kacf:
The underlying problem, which is that reporting is missing in
`git-versions.yml`, has also been fixed on the 3.2.x branch, so it
should not happen again.

Changelog: None

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -408,18 +408,6 @@ def filter_docker_compose_files_list(list, version):
 
     assert version in ["git", "docker"]
 
-    _DOCKER_ONLY_YML = [
-        "docker-compose.yml",
-        "docker-compose.enterprise.yml",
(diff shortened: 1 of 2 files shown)


extra/release_tool.py
```diff
@@ -541,6 +528,18 @@ def version_specific_docker_compose_data_patching(data, rev):
             "version": "master",
         }
 
+    if rev == "3.2.1":
+        for comp in [
+            ("mender-artifact", ".0"),
+            ("mender-cli", ".0"),
+            ("mender-binary-delta", ".0"),
+            ("mender-convert", ".2"),
```
oleorhagen, line 536: :+1:

oleorhagen (approved):

lluiscampos:
Thank you, @kacf

PR #1914, opened 2022-02-08
Pipeline: split specification into staging and default pipelines

lluiscampos:
It should help us with easier maintenance and lower risk of errors when
using mendertesting templates.


lluiscampos:
* Regular pipeline [here](https://gitlab.com/Northern.tech/Mender/integration/-/pipelines/466025152).
* Staging tests pipeline [here](https://gitlab.com/Northern.tech/Mender/integration/-/pipelines/466025512).

lluiscampos:
Seems to work just fine :) What do you guys think? The GitLab docs for this feature can be found [here](https://docs.gitlab.com/ee/ci/yaml/includes.html#use-rules-with-include).

tranchitella (approved):

PR #1916, opened 2022-02-10
Work around and skip problematic release_tool tests on 3.1.x and older.

kacf:
Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -54,8 +54,9 @@ test:extra-tools:
     # Add github remote for tests using --integration-versions-including
     - git remote add github https://github.com/mendersoftware/integration.git
     - git fetch github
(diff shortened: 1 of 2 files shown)


kacf:
@lluiscampos: What do you think about this fix? I'm thinking it's a risk vs benefit balance: We don't need the N-to-N mapping in older branches, so let's avoid cherry-picking and just skip those tests that depend on it. That means that the `--integration-versions-including` argument won't work on 3.1.x and older, but I don't think this is needed for releasing, so it should be fine.

Another option is to keep them very closely in sync, but again I'm thinking to avoid risk in each branch's release process.

lluiscampos (approved):

kacf:
@lluiscampos: Thanks!

PR #1940, opened 2022-02-23
release_tool.py: Purge build tags based on git ls-remote

lluiscampos:
To rely on local tags is not very robust as many of us have several
remotes where the build tags are not being purged after releases (for
example, GitLab remote...)

This commit inverts purge_build_tags logic to look for remote tags and
then delete both local and remote.

It gets significantly slower, so we add a print message to ask for
patience to the caller :)

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -1957,12 +1957,19 @@ def purge_build_tags(state, tag_avail):
     of each repository and ends in '-build[0-9]+'. Then deletes this from
     upstream as well."""
 
+    print("Checking which remote tags need to be purged...")
     git_list = []
     for repo in Component.get_components_of_type("git"):
         remote = find_upstream_remote(state, repo.git())
-        tag_list = execute_git(state, repo.git(), ["tag"], capture=True).split("\n")
+        remote_tag_list = [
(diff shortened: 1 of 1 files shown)


oleorhagen (approved):
Nice!

Nothing like some annoyance to trigger a client friendly patch :)

Maybe we should run prune also? :thinking:

kacf (approved):

lluiscampos:
> Maybe we should run prune also? thinking

You mean something like `git fetch --prune`? That would only remove local references.

Although I don't understand exactly what are you suggesting, I will be merging as it is because it does not modify the expected outcome: remove only remote and local build tags associated to the current release (other build tags are kept).

lluiscampos:
* 3.2.x
* 3.0.x

oleorhagen:
> > Maybe we should run prune also? thinking
> 
> You mean something like `git fetch --prune`? That would only remove local references.
> 
> Although I don't understand exactly what are you suggesting, I will be merging as it is because it does not modify the expected outcome: remove only remote and local build tags associated to the current release (other build tags are kept).

Yeh, not meant to hold off on this functionality :)

PR #1964, opened 2022-03-15
Add docker-compose file to easily demo mender-gateway

lluiscampos:


docker-compose.mender-gateway.commercial.demo.yml
```diff
@@ -0,0 +1,30 @@
...
+    #
+    # mender-client
+    #
+    mender-client:
+        image: mendersoftware/mender-client-qemu:mender-master
+        networks:
+            mender_local:
+        stdin_open: true
+        tty: true
+        privileged: true
```
oleorhagen, line 23: Q:
Why privileged?

oleorhagen (approved):

docker-compose.mender-gateway.commercial.demo.yml
lluiscampos: I honestly don't know. But it is [already in the base image](https://github.com/mendersoftware/integration/blob/master/docker-compose.client.yml#L13).

PR #1965, opened 2022-03-16
MEN-5458 Monitor: tests: we expect only one CRITICAL for log subsystem.

merlin-northern:
ChangeLog:none

### DIFF
--- a/tests/tests/test_monitor_client.py
+++ b/tests/tests/test_monitor_client.py
@@ -510,7 +510,7 @@ class TestMonitorClientEnterprise:
         prepare_log_monitoring(
             mender_device,
             service_name,
-            "@journalctl -u " + service_name,
+            "@journalctl -f -u " + service_name,
             "State transition: .*",
         )
         mender_device.run("systemctl restart mender-monitor")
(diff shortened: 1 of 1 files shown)


merlin-northern:
goes with 
https://github.com/mendersoftware/monitor-client/pull/117
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/493346162

merlin-northern:
failure in the tests is unrelated.

tests/tests/test_monitor_client.py
```diff
@@ -1121,8 +1121,10 @@ def test_monitorclient_logs_and_services(self, monitor_commercial_setup_no_clien
         _, messages = get_and_parse_email_n(
             monitor_commercial_setup_no_client, user_name, 4
         )
-        assert messages_count >= 4
-        for m in [messages[-1], messages[-2]]:
+        messages_count = len(messages)
+        assert messages_count == 5
```
lluiscampos, line 1125: I think we are abusing this `message_count` variable. It is being re-defined in many places and prone to errors. I think we should remove it completely and do only:
```suggestion
        assert len(messages) == 5
```

We can also clean-up this messy block above where `message_count` is being redefined without even checking, and we have two asserts for the same thing:
image

lluiscampos (changes requested):
Please squash both commits, both are changing the two tests and it is not clear that the first one makes sense alone.

tests/tests/test_monitor_client.py
merlin-northern: cleaned-fixed-squashed. thanks.

lluiscampos (approved):

PR #1973, opened 2022-03-23
QA-378: pipeline: tests:docs: Adapt job for new tests

lluiscampos:


lluiscampos:
See:
* https://github.com/mendersoftware/mender-docs/pull/1723

.gitlab-ci-default-pipeline.yml
```diff
@@ -48,12 +48,20 @@ test:extra-tools:
     - python3 -m pytest extra/test_statistics_generator.py
 
 test:docs:
-  image: tiangolo/docker-with-compose
+  image: ubuntu:focal
```
oleorhagen, line 51: Q:
I usually like using the `20.04`, or in other words their yearly tags, as after a few years, I rarely remember which beaver belongs to which year

oleorhagen (approved):

.gitlab-ci-default-pipeline.yml
kacf: +1

kacf (approved):

.gitlab-ci-default-pipeline.yml
lluiscampos: I would do too. But in this case I used `:focal` (same as in `mender-docs` PR) because that is the name we use for the APT repo too. It made sense to refer them this way.

PR #1976, opened 2022-03-30
chore(self-service-sign-up): Add tenantadm environment variable responsible for self service sign up

MaciejTe:
chore(self-service-sign-up): Add tenantadm environment variable responsible for self service sign up

Changelog: None

Signed-off-by: Maciej Tomczuk <maciej.tomczuk@northern.tech>

### DIFF
--- a/docker-compose.enterprise.yml
+++ b/docker-compose.enterprise.yml
@@ -34,6 +34,7 @@ services:
         image: registry.mender.io/mendersoftware/tenantadm:mender-master
         environment:
             TENANTADM_ORCHESTRATOR_ADDR: http://mender-workflows-server:8080/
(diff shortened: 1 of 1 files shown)


docker-compose.enterprise.yml
```diff
@@ -31,9 +31,10 @@ services:
 
     # add services
     mender-tenantadm:
-        image: registry.mender.io/mendersoftware/tenantadm:mender-master
+        image: tenantadm:dev # registry.mender.io/mendersoftware/tenantadm:mender-master
```
tranchitella, line 34: Uh?

MaciejTe: Oops! Fixed

PR #1977, opened 2022-04-01
feat: added version information update job

mzedel:
- this will run on every tag and take existing version information from the docs into account
- the call to `deno fmt` is only for formatting purposes (to ease the review on the docs site) - unfortunately I couldn't find a way to get a pretty + somewhat space efficient JSON formatting with python alone 🤷‍♂️

This would work best with mendersoftware/mender-docs-site#357 in, but should also work without it...

### DIFF
--- a/extra/release_info_generator.py
+++ b/extra/release_info_generator.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+# Copyright 2022 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+from argparse import Namespace
+from calendar import isleap, monthrange
+from collections import namedtuple
+from contextlib import redirect_stdout
+from datetime import datetime, time, timedelta
+from io import StringIO
+from os import getenv
+from re import compile
+import json
+
+from release_tool import do_list_repos
+
+Tag = namedtuple("Tag", ["version", "datetime"])
+LtsEnd = namedtuple("LtsEnd", ["isExpired", "date"])
+
+releaseMatcher = compile(r"([0-9]+)\.([0-9]+)\.([0-9]+)")
+tagMatcher = compile(r"tag:'(.*)' datetime:'(\d+)'")
+
+lastChecked = getenv("LAST_CHECKED", "2.2.0")
+endOfToday = datetime.combine(datetime.now(), time.max)
+
+ltsChecks = {
+    # until 3.1 all even releases were LTS
+    "until-3.2": lambda minorVersion, minorValue: minorVersion < "3.2"
+    and not int(minorValue) % 2,
+    # from 3.3 all odd releases are assumed to be LTS
+    "from-3.3": lambda minorVersion, minorValue: minorVersion >= "3.2"
+    and int(minorValue) % 2,
+}
+
+
+def determine_lts_end(tag, currentMinor, releases):
+    match = releaseMatcher.match(f"{currentMinor}.0")
+    if not any(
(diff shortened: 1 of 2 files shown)


extra/release_info_generator.py
```diff
@@ -0,0 +1,131 @@
...
+
+lastChecked = getenv("LAST_CHECKED", "2.2.0")
+endOfToday = datetime.combine(datetime.now(), time.max)
+
+
+def determine_lts_end(tag, currentMinor, releases):
+    match = releaseMatcher.match(tag.version)
+    # for now we assume all even releases are LTS
+    if int(match.group(2)) % 2:
+        return
```
kacf, line 27: This is already false because we are making 3.3 into an LTS. I don't really have a good solution for this other than manually adjusting it in the map, and then using it here as input... Can we pre-populate it with `lts` versions that don't exist yet, or will that throw the logic off?

extra/release_info_generator.py
```diff
@@ -0,0 +1,131 @@
...
+taggedReleases = get_releases()
+releases = taggedReleases["tags"]
+releaseInformation = {"lts": [], "releases": {}, "saas": taggedReleases["saasTags"]}
+
+existingReleaseInformation = {}
+try:
+    with open("versions.json", "r") as current:
+        existingReleaseInformation = json.load(current)
+except:
+    existingReleaseInformation = existingReleaseInformation
```
kacf, line 102: Nitpick: Isn't this really a no-op, essentially `pass`?

extra/release_info_generator.py
```diff
@@ -0,0 +1,131 @@
...
+    currentMinor = f"{match.group(1)}.{match.group(2)}"
+    initialMinorRelease = next(
+        (rel for rel in releases if rel.version == f"{currentMinor}.0"), release
+    )
+    ltsEnd = determine_lts_end(
+        initialMinorRelease, currentMinor, releaseInformation["releases"]
+    )
+    if ltsEnd and not ltsEnd.isExpired and currentMinor not in ltsReleases:
+        ltsReleases.append(currentMinor)
+    if currentMinor not in releaseInformation["releases"]:
```
kacf, line 119: I take it this is the magic line which prevents old entries from being updated?

extra/release_info_generator.py
mzedel: For now I put these checks in one place to be able to adjust the checks depending on the minor version... that should always work for as long as we keep the LTS schedule and won't require us to modify the versions file or LTS list manually, but it's still not super pretty...

extra/release_info_generator.py
mzedel: ah yeah, it's been a while I `python`ed - I forgot about `pass` 🙈 - adjusted!

extra/release_info_generator.py
mzedel: yes, for all the information not sourced from the `release_tool` (there is also [this part](https://github.com/mendersoftware/integration/pull/1977/files#diff-e2e57cbd86e3596e05d64332a47f53983a12ea4fdd664cd2b9378af89e09f4c2R70-R74) that should ensure some extra information is also kept)

extra/release_info_generator.py
kacf: Ok, better than nothing I guess. The comment should be adjusted though.

extra/release_info_generator.py
```diff
@@ -0,0 +1,136 @@
...
+
+ltsChecks = {
+    "until-3.2": lambda minorVersion, minorValue: minorVersion < "3.2" and not int(minorValue) % 2,
+    "from-3.3": lambda minorVersion, minorValue: minorVersion >= "3.2" and int(minorValue) % 2
+}
+
+
+def determine_lts_end(tag, currentMinor, releases):
+    match = releaseMatcher.match(f"{currentMinor}.0")
+    # for now we assume all even releases are LTS
```
kacf, line 30: Needs adjustment.

kacf (approved):

extra/release_info_generator.py
mzedel: 🤦 argh! Adjusted!

mzedel:
> @mender-test-bot, Let me know if you want to start the integration pipeline by mentioning me and the command "start pipeline".
> 
> my commands and options

🚂

mzedel:
> > @mender-test-bot, Let me know if you want to start the integration pipeline by mentioning me and the command "start pipeline".
> > my commands and options
> 
> 🚂

🚂🚂

PR #1978, opened 2022-04-03
test: backend integration tests for the new RBAC implementation

kjaskiewiczz:
Issue: https://tracker.mender.io/browse/MEN-5162

### DIFF
--- a/backend-tests/tests/test_rbac.py
+++ b/backend-tests/tests/test_rbac.py
@@ -31,6 +31,7 @@ from testutils.common import (
     mongo,
     clean_mongo,
     update_tenant,
+    setup_tenant_devices,
 )
 from testutils.util.artifact import Artifact
 from testutils.api.client import ApiClient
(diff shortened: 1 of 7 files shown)


backend-tests/tests/test_rbac.py
```diff
@@ -474,3 +427,85 @@ def test_get_configuration(self, clean_mongo, test_case):
...
+        )
+
+        rsp = useradm_INT.call(
+            "GET",
+            useradm.URL_EMAILS,
+            path_params={"tenant_id": tenant.id, "device_id": device_id},
+        )
+        assert rsp.status_code == test_case["status_code"], rsp.text
+        emails = rsp.json()["emails"]
+        print("EMAILS: {}".format(emails))
```
tranchitella, line 506: Maybe we don't need the print, here.

tranchitella (approved):

backend-tests/tests/test_rbac.py
kjaskiewiczz: yes, we don't; thanks for spotting!

kjaskiewiczz:
pipeline - https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/511749657
one unrelated failure and one expected (will pass with useradm changes - that's why the PR is addressing feature branch)

PR #1980, opened 2022-04-07
Changelog-generator to group entries by conventional commit types

snusifer:
Ticket: QA-370
Signed-off-by: Mikael Torp-Holte <mikael.torp-holte@northern.tech>

### DIFF
--- a/extra/changelog-generator/changelog-generator
+++ b/extra/changelog-generator/changelog-generator
@@ -20,7 +20,8 @@ LINKED_SHAS = {}
 SHA_TO_TRACKER = {}
 
 JIRA_REGEX = r"\[?(?:Jira:? *)?(?:https?://tracker.mender.io/browse/)?((?:INF|ARCHIVE|MEN|QA|MC)(?:-| )[0-9]+)\]?"
-TRACKER_REGEX = r"\(?(?:Ref:? *)?%s\)?:? *" % (JIRA_REGEX)
+TRACKER_REGEX = r"\(?(?:(?:Ref|Ticket):? *)?%s\)?:? *" % (JIRA_REGEX)
+CONVENTIONAL_COMMIT_TYPE_REGEX = r"^(fix|feat)(?:\(\S+\))?!?: "
 ORGANIZATION = "mendersoftware"
 
 
@@ -161,6 +162,36 @@ def get_range_for_repo(repo, range):
     )
 
 
+def extract_message_and_bucket(message, fixes, feats, others):
+    bucket = others
+    match = re.match(CONVENTIONAL_COMMIT_TYPE_REGEX, message, re.IGNORECASE)
+    if not match:
+        return message, others
+    t = match.group(1).lower()
+    stripped_message = re.sub(
+        CONVENTIONAL_COMMIT_TYPE_REGEX, "", message, count=1, flags=re.IGNORECASE
+    )
+    if t == "fix":
+        bucket = fixes
+    elif t == "feat":
+        bucket = feats
+    return stripped_message, bucket
+
+
+def print_category_entries(entry_list, category, print_category=True):
+    if not entry_list:
+        return
+    entry_list = aggregate_dependabot_changelogs(entry_list)
+    if print_category:
+        print("##### " + category + "\n")
+    for entry in entry_list:
+        entry = "* " + entry
+        # Indent all non-empty lines.
+        entry = re.sub(r"\n([^\n])", r"\n  \1", entry)
+        print(entry)
+    print()
+
+
 print("### Changelogs\n")
 
 for repo in repos:
@@ -232,16 +263,28 @@ for repo in repos:
                 tracker_removed = re.sub(TRACKER_REGEX, "", line, flags=re.IGNORECASE)
                 line = tracker_removed.strip(" ")
 
+            if re.match("^ticket: *none *$", line, re.IGNORECASE):
+                continue
+
             if not title_fetched:
                 title = line
                 title_fetched = True
 
-            match_changelog = re.match("^ *Changelog: *(.*)", line, re.IGNORECASE)
+            match_changelog = re.match(
+                "^ *Changelog(?:\((fix|feat)\))?: *(.*)", line, re.IGNORECASE
+            )
             if match_changelog:
                 if log_entry:
                     add_entry(sha, log_entry.strip())
                     log_entry = ""
                 log_entry_local = False
+
+                type_keyword = match_changelog.group(1)
+                if type_keyword:
+                    add_entry(
+                        sha, type_keyword.lower() + ": " + match_changelog.group(2)
+                    )
+                    continue
 
                 # The logic of this regex is:
                 # - Shall start with Title|Commit|All|None
@@ -249,7 +292,7 @@ for repo in repos:
                 # - Else just expect end line
                 match_keyword = re.match(
                     "^(Title|Commit|All|None) *(?:[.:]+ *(.*)$|$)",
-                    match_changelog.group(1),
+                    match_changelog.group(2),
                     re.IGNORECASE,
                 )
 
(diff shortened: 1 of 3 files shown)


extra/changelog-generator/test-changelog-generator
```diff
@@ -292,9 +294,9 @@ TREE=$(git cat-file commit HEAD | grep '^tree ' | sed -e 's/^tree //')
 git reset --hard $(git commit-tree -p $(git rev-parse HEAD) $TREE <<EOF
 Changelog with Windows line endings N66
 
-Changelog: title
-
-stuff stuff
+Changelog: title
+
+stuff stuff
```
snusifer, line 299: What? Was there some invisible characters here or something

snusifer:
@mender-test-bot 
start pipeline

extra/changelog-generator/test-changelog-generator
snusifer: nvm, carriage return at the end of this lines originally. Weird that vscode removed these when simply opening the file...

extra/changelog-generator/changelog-generator
```diff
@@ -161,6 +162,36 @@ def get_range_for_repo(repo, range):
     )
 
 
+def strip_type(message, fixes, feats, others):
```
oleorhagen, line 165: Q:
Just looking at it, it's a little odd to me that a `strip_type` function returns anything else than what it has stripped.

It's no biggie, but I would not consider it exactly a clean API.

extra/changelog-generator/changelog-generator
```diff
@@ -161,6 +162,36 @@ def get_range_for_repo(repo, range):
...
+def print_category_entries(entry_list, category, print_category=True):
+    if not entry_list:
+        return
+    entry_list = aggregate_dependabot_changelogs(entry_list)
+    if print_category:
+        print("##### " + category + "\n")
+    for entry in entry_list:
+        entry = "* " + entry
+        # Indent all non-empty lines.
+        entry = re.sub(r"\n([^\n])", r"\n  \1", entry)
```
oleorhagen, line 190: Q: Interesting way of doing it. Why this way?

extra/changelog-generator/test-changelog-generator
```diff
@@ -486,6 +589,9 @@ cat > expected.txt <<EOF
         dependency-type: direct:production
         update-type: version-update:semver-patch
       \`\`\`
+
+##### Other
```
oleorhagen, line 593: I'd like to discuss this one with you @kacf :smile:

extra/changelog-generator/changelog-generator
snusifer: It returns the stripped message:
`return stripped_message, bucket`

extra/changelog-generator/changelog-generator
```diff
@@ -416,13 +462,16 @@ for repo in repos:
...
-        entry = re.sub(r"\n([^\n])", r"\n  \1", entry)
-        print(entry)
-    if len(entry_list) > 0:
-        print()
+        (entry, bucket) = strip_type(entry, fixes, feats, others)
+        bucket.append(entry)
+
+    print_category_entries(fixes, "Bug Fixes")
+    print_category_entries(feats, "Features")
+    print_category_entries(others, "Other", fixes or feats)
```
kacf, line 474: `fixes or feats` - This one was new to me. It's the union of the two lists?

extra/changelog-generator/changelog-generator
```diff
@@ -232,24 +263,36 @@ for repo in repos:
                 tracker_removed = re.sub(TRACKER_REGEX, "", line, flags=re.IGNORECASE)
                 line = tracker_removed.strip(" ")
 
+            if re.match("^Ticket: [Nn]one", line):
```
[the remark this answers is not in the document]
snusifer: So the changelog generator iterates over the lines in every commit message. This check ensures that a line that is `Ticket: None` is ignored. A `Changelog` line will be scrutinized in some other iteration and handled accordingly.

extra/changelog-generator/changelog-generator
kacf: I think I would agree. It's mostly the name, the function itself is fine. How about `extract_message_and_bucket`?

extra/changelog-generator/test-changelog-generator
```diff
@@ -382,23 +384,27 @@ Should be included N74'
 
 ################################################################################
 
-git commit --allow-empty -m 'Changelog:All: Bump github.com/klauspost/pgzip from 1.2.3 to 1.2.4
+git commit --allow-empty -m 'fix: Bump github.com/klauspost/pgzip from 1.2.3 to 1.2.4
```
kacf, line 387: I don't think we can remove these strings, because we have limited control over what dependabot posts. Have you discussed how to handle this?

extra/changelog-generator/test-changelog-generator
kacf: F2F?

extra/changelog-generator/changelog-generator
snusifer: Expression is false if and only if both lists are empty.

extra/changelog-generator/test-changelog-generator
snusifer: We do now. Commenting `mark-pr as (type)` on a dependabot PR will amend the commit to this exact format. After enforcing the conventional commit rule, every dependabot PR will most certainly be on this format.

extra/changelog-generator/changelog-generator
snusifer: The idea is that if the lists are empty then we do not print the `##### Other` string. This is to ensure that the script is still compatible with the tests in `../tests`.

extra/changelog-generator/changelog-generator
snusifer: Oh yeah, I can see that it could be interpreted that way :)

extra/changelog-generator/test-changelog-generator
oleorhagen: Yes, @kacf see https://tracker.mender.io/browse/QA-370

extra/changelog-generator/changelog-generator
```diff
@@ -285,7 +328,11 @@ for repo in repos:
                             commit_msg += match_keyword.group(2)
                 else:
                     log_entry_local = True
-                    log_entry = match_changelog.group(1)
+                    type_match = re.match(
+                        CONVENTIONAL_COMMIT_TYPE_REGEX, commit_msg, re.IGNORECASE
+                    )
+                    cc_prefix = type_match.group(0) if type_match else ""
+                    log_entry = cc_prefix + match_changelog.group(2)
```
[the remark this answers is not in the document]
snusifer: 285 doesn't really have anything to do with `cc_prefix`. 285 makes it possible that a line like `Changelog(fix): fix desription` adds a completely separate entry into the list of entries. This particular snippet makes it so that a `Changelog: all` with a type prefix in the title will be added according to that type.

Doesn`t really matter that cc_prefix is lower case, upper case, or anything in between since we ignore case when extracting the message in `strip_type`.

extra/changelog-generator/changelog-generator
kacf: Ah, I misread the arguments. Smart!

extra/changelog-generator/changelog-generator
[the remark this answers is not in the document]
kacf: Doh, you're right, it's line based. Easy to get confused in this long loop.

I would suggest adding a `$` at the end to match the full line though, just to make sure it doesn't match by accident (unlikely, but it's always good to be strict when parsing).

extra/changelog-generator/test-changelog-generator
kacf: 👍

kacf (approved):
I'm assuming you'll fix the minor things mentioned. Other than that, this looks good to me!

PR #1982, opened 2022-04-11
tests: Add several missing waits for containers. 

kacf:


kacf:
Looks promising, the first build above was green.

kacf:
I got two errors on the last run, but I'm fairly certain they are unrelated, especially since no code has changed since the previous green run. Merging.

PR #1983, opened 2022-04-12
QA-387 Extract clean image for tests.

merlin-northern:
Those go together:

* https://github.com/mendersoftware/mender-qa/pull/580
  build and prepare so called clean images and pass them to build-docker
  to embed in the images we later /extract_fs in integration tests
* https://github.com/mendersoftware/meta-mender/pull/1663
  support for a new flag that allows to specify an extra image
  to embed in the container
* https://github.com/mendersoftware/integration/pull/1983
  use the clean images in the tests and test for artifact_info
  in a explicit manner

ChangeLog: none

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -153,15 +153,23 @@ if [[ -z "$BUILDDIR" ]] && [[ -n "$DOWNLOAD_REQUIREMENTS" ]]; then
 fi
 
 # Extract file system images from Docker images
+# the argument for the /extract_fs is the prefix of the image name
+# as per QA-387 now we are explicitly extracting a clean image
+# that is: an image which is crafted to be the one we update to.
+IMG_PREFIX="clean-"
 mkdir -p output
-docker run --rm --privileged --entrypoint /extract_fs -v $PWD/output:/output \
-       mendersoftware/mender-client-qemu:$(../extra/release_tool.py --version-of mender-client-qemu --version-type docker)
-docker run --rm --privileged --entrypoint /extract_fs -v $PWD/output:/output \
-        mendersoftware/mender-client-qemu-rofs:$(../extra/release_tool.py --version-of mender-client-qemu-rofs --version-type docker)
-docker run --rm --privileged --entrypoint /extract_fs -v $PWD/output:/output \
-        registry.mender.io/mendersoftware/mender-gateway-qemu-commercial:$(../extra/release_tool.py --version-of mender-gateway --version-type docker)
+docker run --rm --privileged --entrypoint /extract_fs -v "${PWD}"/output:/output \
+       mendersoftware/mender-client-qemu:$(../extra/release_tool.py --version-of mender-client-qemu --version-type docker) "${IMG_PREFIX}"
+docker run --rm --privileged --entrypoint /extract_fs -v "${PWD}"/output:/output \
+        mendersoftware/mender-client-qemu-rofs:$(../extra/release_tool.py --version-of mender-client-qemu-rofs --version-type docker) "${IMG_PREFIX}"
+docker run --rm --privileged --entrypoint /extract_fs -v "${PWD}"/output:/output \
+        registry.mender.io/mendersoftware/mender-gateway-qemu-commercial:$(../extra/release_tool.py --version-of mender-gateway --version-type docker) "${IMG_PREFIX}"
 mv output/* .
 rmdir output
+
+for i in ${IMG_PREFIX}*; do
+    mv -v "$i" "${i/${IMG_PREFIX}/}"
+done
 
 modify_services_for_testing
 
--- a/tests/tests/test_basic_integration.py
+++ b/tests/tests/test_basic_integration.py
@@ -69,6 +69,17 @@ class BaseTestBasicIntegration(MenderTesting):
         # Verify that partition is read-only as expected
         mender_device.run("mount | fgrep 'on / ' | fgrep '(ro,'")
 
+        # in the so-called clean image the artifact_info is "artifact_name=mender-image-clean\n"
+        # see: https://github.com/mendersoftware/mender-qa/pull/580/commits/a4c54e9ebf5879f0837789cbc7c74cb1102b2caa#diff-7031fcd5c934f225781febe446a34594a9a36b335c540379c0dd376a8f06ba28R263
+        # so first we expect it to be set to master, as this is what we
+        # have in the initial image, after update we expect it to change
+        expected_artifact_info = "artifact_name=mender-image-master\n"
+        artifact_info = mender_device.run(
+            "cat /etc/mender/artifact_info | sed -e '/artifact_name/q'"
+        )
+        logger.info("before update got %s" % artifact_info)
+        assert artifact_info == expected_artifact_info
+
         host_ip = env.get_virtual_network_host_ip()
         update_image(
             mender_device,
@@ -80,6 +91,17 @@ class BaseTestBasicIntegration(MenderTesting):
         )
         mender_device.run("mount | fgrep 'on / ' | fgrep '(ro,'")
 
+        # after the update we expect the artifact info to be different
+        # ("artifact_name=mender-image-clean\n")
+        # since this is what is in the embedded image decompressed
+        # in the tests/run.sh
+        expected_artifact_info = "artifact_name=mender-image-clean\n"
+        artifact_info = mender_device.run(
+            "cat /etc/mender/artifact_info | sed -e '/artifact_name/q'"
+        )
+        logger.info("after update got %s" % artifact_info)
+        assert expected_artifact_info == artifact_info
+
         update_image(
             mender_device,
             host_ip,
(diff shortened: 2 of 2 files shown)


tests/run.sh
```diff
@@ -153,16 +153,44 @@ if [[ -z "$BUILDDIR" ]] && [[ -n "$DOWNLOAD_REQUIREMENTS" ]]; then
...
+for image_tag in mendersoftware/mender-client-qemu:$(../extra/release_tool.py --version-of mender-client-qemu --version-type docker) mendersoftware/mender-client-qemu-rofs:$(../extra/release_tool.py --version-of mender-client-qemu-rofs --version-type docker) registry.mender.io/mendersoftware/mender-gateway-qemu-commercial:$(../extra/release_tool.py --version-of mender-gateway --version-type docker); do
+ id=`docker run --rm -d "${image_tag}"`
+ docker exec "${id}" apk update;
+ docker exec "${id}" apk add openssh-client;
+ docker exec "${id}" bash -c 'max_iterations=255; while [ $max_iterations -gt 0 ]; do sleep 1; nc -z -v 127.0.0.1 8822 && break; let max_iterations--; done'
+ artifact_info=`docker exec "${id}" ssh -o StrictHostKeyChecking=no -p8822 127.0.0.1 cat /etc/mender/artifact_info`
+# we expect the artifact info to be different than the one we expected from the clean image
+ [[ "${expected_artifact_info}" == "${artifact_info}" ]] && { echo "artifact_info is the same in the clean image and the image we run. got: ${artifact_info}"; exit 1; };
+ docker stop "$id"
+done
```
kacf, line 192: Indentation is off in this whole section.

tests/run.sh
```diff
@@ -153,16 +153,44 @@ if [[ -z "$BUILDDIR" ]] && [[ -n "$DOWNLOAD_REQUIREMENTS" ]]; then
...
+        registry.mender.io/mendersoftware/mender-gateway-qemu-commercial:$(../extra/release_tool.py --version-of mender-gateway --version-type docker) "${IMG_PREFIX}"
 mv output/* .
 rmdir output
 
+# check if the expected artifact_name is present
+expected_artifact_info="artifact_name=mender-image-clean"
+for i in ${IMG_PREFIX}*; do
+ mv -v "$i" "${i/${IMG_PREFIX}/}"
+ mender-artifact write rootfs-image -t beaglebone -n release-1 --software-version rootfs-v1 -f "${i/${IMG_PREFIX}/}" -o "/tmp/${i/${IMG_PREFIX}/}.mender"
+ actual_artifact_info=`mender-artifact cat "/tmp/${i/${IMG_PREFIX}/}.mender:/etc/mender/artifact_info"`
```
kacf, line 175: Why are you building an artifact first? Is it because of [this bug](https://tracker.mender.io/browse/MEN-5462)?

tests/run.sh
```diff
@@ -153,16 +153,44 @@ if [[ -z "$BUILDDIR" ]] && [[ -n "$DOWNLOAD_REQUIREMENTS" ]]; then
...
+ [[ "${expected_artifact_info}" == "${actual_artifact_info}" ]] || { echo "mismatched artifact_info. expected: $expected_artifact_info got: ${actual_artifact_info}"; exit 1; };
```
kacf, line 176: I think it would be better if we did this check inside a dedicated test. The problem is that this will now always run, regardless of whether you're debugging a different unrelated test, and this will significantly degrade test startup time.

tests/run.sh
```diff
@@ -153,16 +153,44 @@ if [[ -z "$BUILDDIR" ]] && [[ -n "$DOWNLOAD_REQUIREMENTS" ]]; then
...
+ rm -f "/tmp/${i/${IMG_PREFIX}/}.mender"
+done
+
+# check if artifact_info from inside the image we run is different than the one we update to
+# the condition/naming is misleading since we expect the $expected_artifact_info to not be equal to 
+# $artifact_info in the below images
+for image_tag in mendersoftware/mender-client-qemu:$(../extra/release_tool.py --version-of mender-client-qemu --version-type docker) mendersoftware/mender-client-qemu-rofs:$(../extra/release_tool.py --version-of mender-client-qemu-rofs --version-type docker) registry.mender.io/mendersoftware/mender-gateway-qemu-commercial:$(../extra/release_tool.py --version-of mender-gateway --version-type docker); do
+ id=`docker run --rm -d "${image_tag}"`
```
kacf, line 184: I thought `--rm` and `-d` were mutually exclusive?

tests/run.sh
```diff
@@ -153,16 +153,44 @@ if [[ -z "$BUILDDIR" ]] && [[ -n "$DOWNLOAD_REQUIREMENTS" ]]; then
...
+ docker exec "${id}" apk update;
+ docker exec "${id}" apk add openssh-client;
+ docker exec "${id}" bash -c 'max_iterations=255; while [ $max_iterations -gt 0 ]; do sleep 1; nc -z -v 127.0.0.1 8822 && break; let max_iterations--; done'
+ artifact_info=`docker exec "${id}" ssh -o StrictHostKeyChecking=no -p8822 127.0.0.1 cat /etc/mender/artifact_info`
+# we expect the artifact info to be different than the one we expected from the clean image
+ [[ "${expected_artifact_info}" == "${artifact_info}" ]] && { echo "artifact_info is the same in the clean image and the image we run. got: ${artifact_info}"; exit 1; };
```
kacf, line 190: Same here, I think this should be done in a test, not here.

kacf (changes requested):

tests/tests/test_basic_integration.py
```diff
@@ -52,6 +52,54 @@ def get_devauth_base_path(self):
...
+
+        # after the update we expect the artifact info to be different
+        # ("artifact_name=mender-image-clean\n")
+        # since this is what is in the embedded image decompressed
+        # in the tests/run.sh
+        artifact_info = mender_device.run(
+            "cat /etc/mender/artifact_info | sed -e '/artifact_name/q'"
+        )
+        logger.info("after update got %s" % artifact_info)
+        assert expected_artifact_info != artifact_info
```
kacf, line 101: Why not `"artifact_name=mender-image-clean" == artifact_info`?

kacf, line 101: Also, given that this test is almost identical to `test_double_update_rofs`, except for this check, can we just do the comparison there? In the interest of saving test time?

tests/tests/test_basic_integration.py
```diff
@@ -124,8 +172,48 @@ def do_test_failed_updated_and_valid_update(
...
+        # after the update we expect the artifact info to be different
+        # ("artifact_name=mender-image-clean\n")
+        # since this is what is in the embedded image decompressed
+        # in the tests/run.sh
+        expected_artifact_info = "artifact_name=mender-image-master\n"
+        artifact_info = mender_device.run(
+            "cat /etc/mender/artifact_info | sed -e '/artifact_name/q'"
+        )
+        logger.info("after update got %s" % artifact_info)
+        assert artifact_info != expected_artifact_info
```
kacf, line 213: Same here.

tests/tests/test_basic_integration.py
merlin-northern: mainly because I thought that checking that it is not equal looks better. but since on the other hand we are expecting it to be `artifact_name=mender-image-clean`, I changed it as you suggest.

tests/run.sh
merlin-northern: removed and moved to `test_artifact_name_rofs` and `test_artifact_name` tests.

tests/run.sh
merlin-northern: it works, in a sense: it runs and returns id, the other question is whether it removes when stopped -- I have no idea, I assume it does. this whole section is now removed, as per you suggestions. see `test_artifact_name_rofs` and `test_artifact_name` tests.

tests/run.sh
merlin-northern: wilco. removed and moved to `test_artifact_name_rofs` and `test_artifact_name` tests.

tests/run.sh
merlin-northern: the main reason: I wanted this for the completeness, and that was the way I tested it. as per your suggestions I removed this section.

tests/run.sh
merlin-northern: section removed and tests moved to `test_artifact_name_rofs` and `test_artifact_name`.

tests/tests/test_basic_integration.py
merlin-northern: those tests are identical -- in a sense that they check exactly the same for two different images rofs and regular one. does that make sense?
do you think we should also check for mender monitor and mender-gateway images? (we are not deploying them, so I thought that we do not have to)

merlin-northern: after we discussed privately, and after I understood, I is fixed.

tests/tests/test_basic_integration.py
merlin-northern: same here. I am a bit slow on the uptake. but after discussing it is fixed.

kacf (approved):

PR #1990, opened 2022-04-21
fix: change the default sleep time for the redo loop

oleorhagen:
By default the redo loop uses a jitter variable to add randomness to the
sleeptime. This value defaults to 1.

Only seen locally, but did experience underflow in the case of the jitter value
being close to the sleeptime of 1.

Changelog: None


oleorhagen:

PR #1991, opened 2022-04-22
Fix staging tests

tranchitella:


tranchitella:

oleorhagen (approved):

oleorhagen:
Thanks!

backend-tests/tests/test_azure.py
```diff
@@ -140,14 +140,14 @@ class TestAzureIntegrations(_TestAzureBase):
             {
                 "provider": "iot-hub",
                 "credentials": {
-                    "connection_string": "HostName=localhost;SharedAccessKey=thisIsBase64;SharedAccessKeyName=OldKey",
+                    "connection_string": "HostName=test.azure-devices.net;SharedAccessKey=thisIsBase64;SharedAccessKeyName=OldKey",
```
alfrunes, line 143: Is it a problem to use `localhost` for these tests? If so, can we use a domain that we own (e.g. `mender-test-hub.azure-devices.net` or `mender.io`)? Just in case whoever owns `test.azure-devices.net` figures out they don't need it anymore...

tranchitella: Using `localhost` makes the tests fail when targeting the staging environment.
I can replace `test.azure-devices.net` with `mender-test-hub.azure-devices.net`, if you think it is better.

tranchitella: let's do it, I force-pushed using `mender-test-hub.azure-devices.net`.

tranchitella:

alfrunes (approved):

PR #1998, opened 2022-04-28
test: Ensure that path and exist in OK emails

merlin-northern:
goes with: https://github.com/mendersoftware/monitor-client/pull/122

Ticket:MEN-5600

### DIFF
--- a/tests/tests/test_monitor_client.py
+++ b/tests/tests/test_monitor_client.py
@@ -461,7 +461,7 @@ class TestMonitorClientEnterprise:
             update_check_file_only=True,
         )
         time.sleep(2 * wait_for_alert_interval_s)
-        _, messages = get_and_parse_email_n(
+        mail, messages = get_and_parse_email_n(
             monitor_commercial_setup_no_client, user_name, messages_count + 1
         )
         messages_count = len(messages)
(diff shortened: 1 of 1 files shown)


tests/tests/test_monitor_client.py
```diff
@@ -470,6 +470,10 @@ def test_monitorclient_alert_email(self, monitor_commercial_setup_no_client):
             user_name,
             "OK: Monitor Alert for Log file contains " + log_pattern + " on " + devid,
         )
+        # in each CRITICAL and OK email we expect the pattern to be present at least 3 times
+        assert mail.count(log_pattern) >= 6
+        # in each CRITICAL and OK email we expect the log path to be present at least 1 time
+        assert mail.count(log_pattern) >= 2
```
merlin-northern, line 476: for a second I thought about parsing html, or comparing the decoded email body parts with a reference emails hardcoded in the test, but it would bring additional complexity as well as would bind us to the email format (dates in the footer, and possible layout changes, would make it a bit impractical). perhaps it is enough to ensure the number of times the path and pattern occur in the email.

oleorhagen, line 476: present at least `2` times it looks like, no :thinking:

merlin-northern: > present at least `2` times it looks like, no 🤔

nice that you read the comments! it is present at least 2 times in the emails, but we have two emails. so it is 1 each.

oleorhagen (approved):

merlin-northern:
failure in [TestRemoteTerminalEnterprise.test_session_recording](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/2410819097#L2925) is unrelated.

PR #2000, opened 2022-04-29
chore: enhance tests to run on scratch images

tranchitella:
The resulting image doesn't provide `sed` or other CLI tools when
building the back-end containers using the scratch Docker image.
Therefore, we must use the `cp` command to download and upload files to
and from the container when we need to change their content.

Changelog: none

### DIFF
--- a/testutils/infra/cli.py
+++ b/testutils/infra/cli.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Northern.tech AS
+# Copyright 2022 Northern.tech AS
 #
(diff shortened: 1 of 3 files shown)


kjaskiewiczz (approved):

mzedel (approved):
Nice 👍 - I honestly completely forgot about the possibility of using `docker cp`

tranchitella:
Failure in `tests/test_portforward.py::TestPortForwardOpenSource::test_portforward` is unrelated.

PR #2007, opened 2022-05-02
Disable reporting backend tests

tranchitella:


tranchitella:

backend-tests/tests/test_reporting.py
```diff
@@ -238,6 +238,7 @@ def tenant_os(clean_mongo):
     isK8S(),
     reason="reporting service not deployed to staging or production environment",
 )
+@pytest.mark.skip(reason="reporting is not part of the release")
```
oleorhagen, line 241: :thinking:  shouldn't this still be skipped?

oleorhagen (approved):

backend-tests/tests/test_reporting.py
tranchitella: I don't get your question... marking it this way, we'll skip it, right?
I'm keeping the other conditional skips because we'll revert this commit (sooner or later) and get back to the previous state once reporting becomes a default component.

alfrunes (approved):

backend-tests/tests/test_reporting.py
oleorhagen: lol, nvm. I thought it was removed...

PR #2019, opened 2022-05-18
chore(gitdm): Update the mailmap for the 3.3.0 release

oleorhagen:

### DIFF
--- a/extra/gitdm/company-map
+++ b/extra/gitdm/company-map
@@ -44,3 +44,8 @@ sentinel.tech Sentinel.tech
 timesys.com Timesys Corporation
 tiscali.it Tiscali S.p.a.
 jester@theyoctojester.info Northern.tech
+engineeredarts.co.uk Engineered Arts
+code@jackben.net Engineered Arts
+disruptive-technologies.com Disruptive Technologies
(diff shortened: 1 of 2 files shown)


oleorhagen:
TODO - cherry to `3.3.x`

extra/gitdm/company-map
```diff
@@ -44,3 +44,8 @@ sentinel.tech Sentinel.tech
 timesys.com Timesys Corporation
 tiscali.it Tiscali S.p.a.
 jester@theyoctojester.info Northern.tech
+j.bennett@engineeredarts.co.uk Engineered Arts
+code@jackben.net Engineered Arts
+sven.schwermer@disruptive-technologies.com Disruptive Technologies
+tobias.zimmerer@zf.com ZF
+r.schwarz@sotec.eu SOTEC
```
kacf, line 51: For addresses where the domain clearly belongs to the company, it is better to add them without a username, so that future contributors are also covered. This is in contrast to gmail addresses, which need to be specified per user. You can see `tiscali.it` above is done this way.

kacf (approved):
Pre-approving, but see below.

PR #2021, opened 2022-05-19
fix: return 404 instead of 500 when a device updates non-existent dep…

tranchitella:
…loyments

Changelog: None
Ticket: MEN-5602

(cherry picked from commit 6c01c96857582e504001cd8e92a97c651d2a7980)


tranchitella:
@mender-test-bot start pipeline --pr deployments/729

PR #2038, opened 2022-05-27
feat: make use of the S3 external URI configuration for deployments

tranchitella:
The deployments service supports a new setting which allows specifying
the external URL to use when generating pre-signed URL. Therefore, the
communication between the deployments service and the storage layer can
now use the internal service name (minio) instead of routing requests to
the API gateway with the public-facing URL.

Changelog: Title
Ticket: MEN-5280


tranchitella:

alfrunes (approved):

tranchitella:

PR #2051, opened 2022-06-01
test: Keep Alive client test fix.

merlin-northern:
Changelog: None
Ticket: None


merlin-northern:
passed here: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/553117033 but still will run the full integration.

PR #2054, opened 2022-06-04
test: Personal Access Tokens (PAT) backend tests expansion.

merlin-northern:
useradm RBAC tests are extended with a tests case with PAT

Changelog: None
Ticket: MEN-5524


backend-tests/tests/test_rbac_v2.py
```diff
@@ -439,31 +797,73 @@ def test_set_and_deploy_configuration(self, clean_mongo, test_case):
...
+                ],
+                "device_groups": {"test": 1, "production": 1},
+                "view_group": "production",
+                "get_configuration_status_code": 403,
+            },
+        ],
+    )
+    def test_get_configuration(self, clean_mongo, test_case):
+        """
+        Tests adding group restrinction to roles and checking that users
```
MaciejTe, line 846: typo: `restrinction`

backend-tests/tests/test_rbac_v2.py
```diff
@@ -36,21 +36,32 @@
...
     login authenticates the user and saves the user token
     with user object
+    in case of use_personal_access_token===True it saves the personal access token
     """
     useradm_MGMT = ApiClient(useradm.URL_MGMT)
     rsp = useradm_MGMT.call("POST", useradm.URL_LOGIN, auth=(user.name, user.pwd))
     assert rsp.status_code == 200, "Failed to setup test environment"
     user.token = rsp.text
 
+    if use_personal_access_token:
```
MaciejTe, line 50: If we want to use personal access token, I'd move this part to the beginning of function, like this:
```
if use_personal_access_token:
   . . . 
else:
    useradm_MGMT = ApiClient(useradm.URL_MGMT)
    rsp = useradm_MGMT.call("POST", useradm.URL_LOGIN, auth=(user.name, user.pwd))
    assert rsp.status_code == 200, "Failed to setup test environment"

user.token = rsp.text
```

Do we want to execute this part of code even if PAT is being used?
```   useradm_MGMT = ApiClient(useradm.URL_MGMT)
    rsp = useradm_MGMT.call("POST", useradm.URL_LOGIN, auth=(user.name, user.pwd))
    assert rsp.status_code == 200, "Failed to setup test environment"
    user.token = rsp.text
```

backend-tests/tests/test_rbac_v2.py
```diff
@@ -36,21 +36,32 @@
...
+        pat_req = {"name": "my_token_0", "expires_in": 1024}
+        rsp = useradm_MGMT.with_auth(user.token).call(
+            "POST", useradm.URL_TOKENS, pat_req
+        )
+        assert (
+            rsp.status_code == 200
+        ), "Failed to setup test environment with personal token"
+        user.token = rsp.text
```
MaciejTe, line 58: It's not good practice to edit objects fields outside the class - `login` function should be `user` object method IMHO

backend-tests/tests/test_rbac_v2.py
merlin-northern: yes, we do. we need to be logged in to issue a token.

backend-tests/tests/test_rbac_v2.py
merlin-northern: yeah I know, but it was there. do you think I should fix that everywhere? or perhaps in another PR?

backend-tests/tests/test_rbac_v2.py
```diff
@@ -36,21 +36,32 @@
 from testutils.api.client import ApiClient
 
 
-def login(user):
+def login(user, use_personal_access_token=False):
```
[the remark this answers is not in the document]
merlin-northern: fixed

backend-tests/tests/test_rbac_v2.py
MaciejTe: Excuse me, my bad! I'll raise another concern in different comment

backend-tests/tests/test_rbac_v2.py
MaciejTe: Another PR; another epic ;-) Resolving since we cannot do about it here.

backend-tests/tests/test_rbac_v2.py
MaciejTe: Please at least update the docstring with precise information. Thanks!

kjaskiewiczz (approved):

backend-tests/tests/test_rbac_v2.py
merlin-northern: I was so happy I saw all green here, that I just pressed merge. updated the comment here: https://github.com/mendersoftware/integration/pull/2064

PR #2063, opened 2022-06-08
useradm RBAC tests are extended with a tests case with PAT

kjaskiewiczz:
Ticket: [MEN-5524](https://tracker.mender.io/browse/MEN-5524)


backend-tests/tests/test_rbac_v2.py
```diff
@@ -118,6 +118,7 @@ def logger(self):
                                 "name": "DeployToDevices",
                                 "scope": {"type": "DeviceGroups", "value": ["test"],},
                             },
+                            {"name": "Basic"},
```
merlin-northern, line 121: I do not have a problem with adding basic here, my reasoning for adding explicit POST and GET for /settings/tokens here https://github.com/mendersoftware/integration/pull/2054/commits/9b3d4f1ecc369e4e931dbeca865e07d371577144 was:
* add minimal set of permissions required for the test to pass
* verify that the test passes with only PAT access
* do not introduce dependency on Basic permissions, as these may change, and the test behaviour maybe influenced by it. (that is: we introduce a dependency on Basic permissions in this test, which I do not entirely like)

merlin-northern, line 121: I do not have a problem with adding basic here, my reasoning for adding explicit POST and GET for /settings/tokens here https://github.com/mendersoftware/integration/pull/2054/commits/9b3d4f1ecc369e4e931dbeca865e07d371577144 was:
* add minimal set of permissions required for the test to pass
* verify that the test passes with only PAT access
* do not introduce dependency on Basic permissions, as these may change, and the test behaviour maybe influenced by it. (that is: we introduce a dependency on Basic permissions in this test, which I do not entirely like)

PR #2076, opened 2022-06-17
Update keygen

sxz0:
Newer openssl versions (+3.0.2) no longer generate the key certificate in the old format using the previous command ("openssl rsa -in $FILE_NAME_PRIVATE_KEY -out $FILE_NAME_PRIVATE_KEY"). The option "-traditional" should be added to the openssl rsa key conversion command in line 100.

### DIFF
--- a/keygen
+++ b/keygen
@@ -97,7 +97,7 @@ do
 
     # convert to RSA private key format, otherwise services complain:"
     # level=fatal msg="failed to read rsa private key: jwt: can't open key - not an rsa private key" file=proc.go func=runtime.main line=183
-    openssl rsa -in $FILE_NAME_PRIVATE_KEY -out $FILE_NAME_PRIVATE_KEY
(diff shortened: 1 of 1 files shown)


sxz0:
This issue has been happening to other users too (I am the one resolving the question):
https://hub.mender.io/t/mender-server-3-2-container-deviceauth-and-useradm-continuously-restarting/5000

keygen
```diff
@@ -97,7 +97,7 @@ do
 
     # convert to RSA private key format, otherwise services complain:"
     # level=fatal msg="failed to read rsa private key: jwt: can't open key - not an rsa private key" file=proc.go func=runtime.main line=183
-    openssl rsa -in $FILE_NAME_PRIVATE_KEY -out $FILE_NAME_PRIVATE_KEY
+    openssl rsa -in $FILE_NAME_PRIVATE_KEY -out $FILE_NAME_PRIVATE_KEY -traditional
```
kacf, line 100: Thanks. I think we need to make this conditional though. OpenSSL 1.x series doesn't support this flag. Maybe grep `openssl rsa -help` looking for it?

PR #2084, opened 2022-06-29
test: Additional sleeps around echo on/off in test_session_recording

merlin-northern:
Changelog: None
Ticket: None

### DIFF
--- a/tests/tests/test_mender_connect.py
+++ b/tests/tests/test_mender_connect.py
@@ -210,14 +210,18 @@ class _TestRemoteTerminalBase:
             shell.sendInput("echo 'now you see me'\n".encode())
             session_bytes += get_cmd(ws)
             # Disable echo
+            time.sleep(1)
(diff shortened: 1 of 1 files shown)


merlin-northern:

merlin-northern:
@kacf maybe lets give it a go? (assuming the pipeline will pass :>)

alfrunes (approved):
I think only one sleep is needed. I'm pretty sure only the first two sleep commands have any practical effect in this PR since the real problem is receiving input before the terminal process has started. Anyway, approving since I'm pretty sure this will probably fix the spurious errors.

merlin-northern:
> I think only one sleep is needed. I'm pretty sure only the first two sleep commands have any practical effect in this PR since the real problem is receiving input before the terminal process has started. Anyway, approving since I'm pretty sure this will probably fix the spurious errors.

ok thanks for approval, I will monitor that, and let's after a week or so if we can reduce the number of sleeps.

PR #2085, opened 2022-07-01
test: drop the -O option to scp

merlin-northern:

### DIFF
--- a/testutils/infra/device.py
+++ b/testutils/infra/device.py
@@ -330,7 +330,7 @@ def _put(device, file, local_path=".", remote_path="."):
     (scp, host, port) = _scp_prep_args(device)
 
     subprocess.check_call(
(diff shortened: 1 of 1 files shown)


merlin-northern:

oleorhagen (changes requested):
The `-O` option is needed for the Ubuntu 22 runner, as it defaults to sftp, instead of legacy scp, and we don't have sftp support.

oleorhagen:
Locally, I would suggest a `git smudge` workaround for you here, so you can jsut revert the change locally, and then apply it again when you push :)

merlin-northern:
yeh makes sense.

merlin-northern:
> Locally, I would suggest a `git smudge` workaround for you here, so you can jsut revert the change locally, and then apply it again when you push :)

I was confused as ubuntu 22.04 the only distribution that apparently has this. (that I could find)

PR #2090, opened 2022-07-05
fix(test): Make default tenant token tests not break the configuration

alfrunes:
Instead of appending to the yaml file and breaking it on the second
entry of the function modifying the runtime configuration of deviceauth,
the function is now changed to a contextmanager and tries to restore the
previous state of the configuration.

Changing the runtime configuration of a container is extremely prone to error and for a more permanent fix we should instead change the container setup, e.g. by temporarily bringing in another replica of the modified container.


alfrunes:

kjaskiewiczz (approved):

PR #2098, opened 2022-07-16
feat: 2FA is now available in all plans

tranchitella:
Changelog: None
Ticket: MEN-5680


tranchitella:

tranchitella:
`tests/test_useradm_enterprise.py::Test2FAEnterprise::test_enable_disable[os]` correctly failed in the run without the updated `useradm-enterprise` (https://github.com/mendersoftware/useradm-enterprise/pull/498).

backend-tests/tests/test_useradm_enterprise.py
```diff
@@ -49,8 +49,8 @@ def clean_migrated_mongo(clean_mongo):
     yield clean_mongo
 
 
-@pytest.fixture(scope="function")
-def tenants_users(clean_migrated_mongo):
+@pytest.fixture(scope="function", params=["os", "professional", "enterprise"])
+def tenants_users(request, clean_migrated_mongo):
```
merlin-northern, line 53: idle comment/I am wondering shall we add a scenario with param `"some-other-plan"`? shall it fail or pass?

backend-tests/tests/test_useradm_enterprise.py
```diff
@@ -49,8 +49,8 @@ def clean_migrated_mongo(clean_mongo):
     yield clean_mongo
 
 
-@pytest.fixture(scope="function")
-def tenants_users(clean_migrated_mongo):
+@pytest.fixture(scope="function", params=["os", "professional", "enterprise"])
```
merlin-northern, line 52: nit: isn't something in the lines of `@pytest.mark.parametrize("plan", ["os", "professional", "enterprise"])` more elegant?

merlin-northern (approved):
one question and a suggestion, pre-approving since it looks good to me.

backend-tests/tests/test_useradm_enterprise.py
tranchitella: I don't think you can use pytest.mark.parametrize on a fixture, only on a test.

backend-tests/tests/test_useradm_enterprise.py
tranchitella: I don't think so, it will fail as you cannot create a tenant with a non existing plan.

kjaskiewiczz (approved):

PR #2099, opened 2022-07-19
test: print the list of containers when tests fail

tranchitella:
Changelog: None
Ticket: None


tranchitella:
Manual pipeline (tests in parallel: 1):
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/591132921

PR #2102, opened 2022-07-22
test: remove ports when running tests with the storage proxy

tranchitella:
Tests running the legacy client can occasionally fail because of `Bind
for 0.0.0.0:9000 failed: port is already allocated`. This change
prevents it removing the listening ports from the Docker composition
files.

Changelog: None
Ticket: None


tranchitella:

kacf (approved):
Nice catch!

PR #2103, opened 2022-07-22
Changelog generator improvements

ghost:
See individual commits.

This is not urgent at all, just small improvements I made while working on the kirkstone release.

### DIFF
--- a/extra/changelog-generator/changelog-generator
+++ b/extra/changelog-generator/changelog-generator
@@ -108,6 +108,14 @@ parser.add_argument(
     help=argparse.SUPPRESS,
 )
 parser.add_argument(
+    "--append-commit-sha",
+    dest="append_commit_sha",
+    action="store_true",
+    help="Append to every entry the SHA that it came from. "
+    + "This can help when tidying up history with Cancel-changelog "
+    + "after a long release cycle.",
+)
+parser.add_argument(
     "range", metavar="<commit-range> [--]", help="Range of commits to generate log for"
 )
 parser.add_argument(
@@ -486,6 +494,10 @@ for repo in repos:
                 else:
                     entry += " "
                 entry += tracker
+
+            if args.append_commit_sha:
+                entry += " (Git SHA: %s)" % sha_entry
+
             entry_list.append(entry)
 
     if args.sort_changelog:


extra/changelog-generator/changelog-generator
```diff
@@ -23,7 +23,7 @@ SHA_TO_TRACKER = {}
 
 JIRA_REGEX = r"\[?(?:Jira:? *)?(?:https?://tracker.mender.io/browse/)?((?:INF|ARCHIVE|MEN|QA|MC)(?:-| )[0-9]+)\]?"
 TRACKER_REGEX = r"\(?(?:(?:Ref|Ticket):? *)?%s\)?:? *" % (JIRA_REGEX)
-CONVENTIONAL_COMMIT_TYPE_REGEX = r"^(fix|feat|chore)(?:\(\S+\))?!?: "
+CONVENTIONAL_COMMIT_TYPE_REGEX = r"^(fix|feat|chore)(?:\(\S+\))?!?: ?"
```
snusifer, line 26: A small downside to this:
The conventional commit standard states that a colon and space is mandatory after the initial commit keyword. There might be problems (small, but problems nontheless) if we were to adopt any third party conventional commit tooling in the future.

kacf: I see. We might want to make the opposite change then: Make sure [`check_commits.sh`](https://github.com/mendersoftware/mendertesting/blob/master/check_commits.sh) doesn't accept this. Because right now it accepts it, and then the entry is misrendered in the changelog.

snusifer: 100% agree.

kacf: Agreed!

However, I think we should keep this change then, since there are already commits like that in our history, and this needs to render them correctly. It's not responsible for correctness, it's just responding to what's there.

snusifer: If they already exist then yes. But perhaps amend the commit message of this particular commit for environment friendly purposes.

kacf: :+1:

lluiscampos (approved):
The improvements look good

oleorhagen (approved):
Nice !

Also, :laughing:  https://github.com/mendersoftware/integration/pull/2103/commits/7f568195a36c1830bf6f7dd52952ebf30a7da2a7

extra/changelog-generator/changelog-generator
kacf: https://github.com/mendersoftware/mendertesting/pull/206

PR #2105, opened 2022-07-25
chore: enhance wait_for_connect making it more robust

tranchitella:
TestAccessEnterprise.test_upgrades occasionally fails because
mender-connect running on the device is not connected when performing
troubleshoot add-on tests. This can happen because the tenant upgrade
triggers a workflow that resets the JWT tokens, and it can take a few
seconds for the Mender client to authorize again.

This change makes sure wait_for_connect waits for two consecutive checks
with 5 seconds interval before declaring the mender-connect client
component is available.

Changelog: None
Ticket: None

### DIFF
--- a/tests/tests/common_connect.py
+++ b/tests/tests/common_connect.py
@@ -76,6 +76,7 @@ def wait_for_connect(auth, devid):
         base_url=deviceconnect.URL_MGMT,
     )
 
+    connected = 0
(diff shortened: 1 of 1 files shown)


merlin-northern (approved):
lgtm. good catch.

tranchitella:

PR #2106, opened 2022-07-25
test: increase recv timeout to 5 seconds

tranchitella:
When running multiple tests in parallel, there is a high risk the shell
running in the Qemu device and controlled by mender-connect is not fast
enough at sending the output to deviceconnect. We increase the timeout
from one second to five to make sure we receive the full output before
proceeding with the tests.

Changelog: None
Ticket: None


tranchitella:

merlin-northern (approved):

PR #2108, opened 2022-07-26
ci: install openssh-client instead of openssh

tranchitella:
Avoid the following error:

ERROR: unable to select packages:
  openssh-client-common-9.0_p1-r1:
    breaks: openssh-client-default-9.0_p1-r2[openssh-client-common=9.0_p1-r2]

Changelog: None
Ticket: None

### DIFF
--- a/tests/requirements/apk-requirements.txt
+++ b/tests/requirements/apk-requirements.txt
@@ -1,7 +1,7 @@
 bash
 curl
 git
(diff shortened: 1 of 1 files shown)


tranchitella:

merlin-northern (approved):
how can it break like that all of a sudden?

PR #2110, opened 2022-07-27
ci: install openssh-client instead of openssh

0lmi:
Avoid the following error:

ERROR: unable to select packages:
  openssh-client-common-9.0_p1-r1:
    breaks: openssh-client-default-9.0_p1-r2[openssh-client-common=9.0_p1-r2]

Changelog: None
Ticket: None


0lmi:

tranchitella (approved):

PR #2123, opened 2022-08-19
Update test_rootfs_image_rejected + Improve robustness of test_monitorclient_alert_email

lluiscampos:
test(test_rootfs_image_rejected): Update test
    
    Following implementation of bootstrap Artifact support.
    
    Ticket: MEN-2583

    test(test_monitorclient_alert_email): Improve robustness
    
    Disable the checks of each scenario when the verification is done.
    
    This test covers different (independent!) scenarios, creating and
    enabling checks for each of them. In some cases the alerts from one
    scenario can leak to the next, and then the logic of "expect the alert
    in the last received message" is too naive.

    The last scenario (expiration) needs now to create the check from
    scratch because it is disabled previously.

    Moved also the mender-monitor paths definitions to a single place.

### DIFF
--- a/tests/tests/test_monitor_client.py
+++ b/tests/tests/test_monitor_client.py
@@ -76,6 +76,11 @@ expected_from = (
 )
 
 
+# mender-monitor paths
+MONITOR_AVAILABLE_DIR = "/etc/mender-monitor/monitor.d/available"
+MONITOR_ENABLED_DIR = "/etc/mender-monitor/monitor.d/enabled"
+
+
 @retriable(sleeptime=60, attempts=5)
 def get_and_parse_email_n(env, address, n):
     mail, messages = get_and_parse_email(env, address)
@@ -140,6 +145,10 @@ def assert_valid_alert(message, bcc, subject):
     assert message["Subject"].startswith(subject)
 
 
+def disable_all_checks(mender_device):
+    mender_device.run(f"rm -f {MONITOR_ENABLED_DIR}/*")
+
+
 def prepare_service_monitoring(mender_device, service_name, use_ctl=False):
     if use_ctl:
         mender_device.run(
@@ -148,10 +157,8 @@ def prepare_service_monitoring(mender_device, service_name, use_ctl=False):
         mender_device.run('mender-monitorctl enable service "%s"' % (service_name))
     else:
         try:
-            monitor_available_dir = "/etc/mender-monitor/monitor.d/available"
-            monitor_enabled_dir = "/etc/mender-monitor/monitor.d/enabled"
-            mender_device.run("mkdir -p '%s'" % monitor_available_dir)
-            mender_device.run("mkdir -p '%s'" % monitor_enabled_dir)
+            mender_device.run(f"mkdir -p '{MONITOR_AVAILABLE_DIR}'")
+            mender_device.run(f"mkdir -p '{MONITOR_ENABLED_DIR}'")
             mender_device.run("systemctl restart %s" % service_name)
             tmpdir = tempfile.mkdtemp()
             service_check_file = os.path.join(tmpdir, "service_" + service_name + ".sh")
@@ -161,14 +168,14 @@ def prepare_service_monitoring(mender_device, service_name, use_ctl=False):
             mender_device.put(
                 os.path.basename(service_check_file),
                 local_path=os.path.dirname(service_check_file),
-                remote_path=monitor_available_dir,
+                remote_path=MONITOR_AVAILABLE_DIR,
             )
             mender_device.run(
                 "ln -s '%s/service_%s.sh' '%s/service_%s.sh'"
                 % (
-                    monitor_available_dir,
+                    MONITOR_AVAILABLE_DIR,
                     service_name,
-                    monitor_enabled_dir,
+                    MONITOR_ENABLED_DIR,
                     service_name,
                 )
             )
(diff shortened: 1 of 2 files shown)


lluiscampos:
@kacf The first commit is a behavior change with the introduction of the bootstrap Artifact. That behavior I also tested in unit tests.

@merlin-northern The second commit I think makes the test more stable in the future. What I think happened here is that the daemon timing changed so that another alert from the previous scenario gets send afterwords. See my commit message. In any case I believe that my change does not affect the design of the test.

kacf (approved):
Great investigation/improvement! 👍 👍

tests/tests/test_monitor_client.py
```diff
@@ -140,6 +145,10 @@ def assert_valid_alert(message, bcc, subject):
     assert message["Subject"].startswith(subject)
 
 
+def stop_all_monitoring(mender_device):
+    mender_device.run(f"rm -f {MONITOR_ENABLED_DIR}/*")
```
merlin-northern, line 149: nit: technically it is `disable_all_checks`
another nit: I leave it up to you to decide, I believe we should use monitorctl to disable the checks (that way we do not have to update the test in case we change some logic behind the disable/enable operations). I know it is higher effort, since we do not have --disable-all-checks in monitorctl. I also know that parts of this file use the rad `ln` logic, but those parts predate the monitorctl command :)

merlin-northern (approved):
pre-approving, two nit-picking comments. since Kristian already approved, I reviewed the second commit. General comment: one of the ideas of the test was to test the scenario where there are lots of check enabled, thats why it was like that. but if there is a timing issue and as we discussed on Friday the emails are arriving out of order, we can live with your change.

tests/tests/test_monitor_client.py
lluiscampos: > nit: technically it is `disable_all_checks`

I was trying to mirror the method that creates the checks, which is called `prepare_service_monitoring`. But okey, I'll rename it.

> another nit: I leave it up to you to decide, I believe we should use monitorctl to disable the checks (that way we do not have to update the test in case we change some logic behind the disable/enable operations). I know it is higher effort, since we do not have --disable-all-checks in monitorctl. I also know that parts of this file use the raw `ln` logic, but those parts predate the monitorctl command :)

Nice suggestion but I am sorry to pass. Basically I would need to call `mender-monitorctl list`, parse the "enabled" part (both the subsystems and the check name!), then call `mender-monitorctl disable ...` - which I think is too much work for a boy scout.

lluiscampos (commented):
> General comment: one of the ideas of the test was to test the scenario where there are lots of checks enabled, thats why it was like that. but if there is a timing issue and as we discussed on Friday the emails are arriving out of order, we can live with your change.

@merlin-northern Yes, and that is a good scenario to test, but:
* We should design a test specifically for that and it should be robust. For example we should enable all checks, excessive the device to provoke alerts, and then parse all alerts with a criteria like "for this check I need at least two alerts, for this other check this and that" and in no way depend on the order of these alerts
* As the test was now, it was testing that several checks can be enabled but still that the alerts are send in an specific order and never expecting two alert-able events to happen at the same time. So in my opinion it was not really valuable anyway.

merlin-northern (approved):

merlin-northern:
> > General comment: one of the ideas of the test was to test the scenario where there are lots of checks enabled, thats why it was like that. but if there is a timing issue and as we discussed on Friday the emails are arriving out of order, we can live with your change.
> 
> @merlin-northern Yes, and that is a good scenario to test, but:
> 
> * We should design a test specifically for that and it should be robust. For example we should enable all checks, excessive the device to provoke alerts, and then parse all alerts with a criteria like "for this check I need at least two alerts, for this other check this and that" and in no way depend on the order of these alerts

later we added something like it with logs and services test.

> * As the test was now, it was testing that several checks can be enabled but still that the alerts are send in an specific order and never expecting two alert-able events to happen at the same time. So in my opinion it was not really valuable anyway.

the order of emails was killing us many times. and the email checking should have been different from the beginning (we should grep the inbox more consciously and e.g. for a specific test check order only to verify that CRITICAL was followed by the OK (if expected) and otherwise the order of emails should not matter), apparently there was never enough time for that.

yeah in general I agree. let's have this one in.

tests/tests/test_monitor_client.py
merlin-northern: > > nit: technically it is `disable_all_checks`
> 
> I was trying to mirror the method that creates the checks, which is called `prepare_service_monitoring`. But okey, I'll rename it.
> 
> > another nit: I leave it up to you to decide, I believe we should use monitorctl to disable the checks (that way we do not have to update the test in case we change some logic behind the disable/enable operations). I know it is higher effort, since we do not have --disable-all-checks in monitorctl. I also know that parts of this file use the raw `ln` logic, but those parts predate the monitorctl command :)
> 
> Nice suggestion but I am sorry to pass. Basically I would need to call `mender-monitorctl list`, parse the "enabled" part (both the subsystems and the check name!), then call `mender-monitorctl disable ...` - which I think is too much work for a boy scout.

I know. it is ok.

boy scout as in [here](https://github.com/mendersoftware/integration/pull/2123#discussion_r951152056) ? :)

lluiscampos: Almost :stuck_out_tongue_closed_eyes: Rather [this other one](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html).

lluiscampos:
The [last force-push](https://github.com/mendersoftware/integration/compare/edf3c65b3c0e84c4d150dddcd2f398878bbfc04a..e7db1f22ff4bf8e4eb0eb3774868e3616b50f883) had only cosmetic changes. Merging.

PR #2128, opened 2022-08-26
[test] adapt the RBAC tests to the adjusted permission sets

0lmi:
Changelog: None
Ticket: MEN-5777
Signed-off-by: Alex Miliukov <oleksandr.miliukov@northern.tech>


0lmi:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/624008100

backend-tests/tests/test_rbac_v2.py
```diff
@@ -496,42 +475,66 @@ def test_deploy_to_group(self, clean_mongo, test_case):
...
+                    "name": "test1-UUID@example.com",
+                    "pwd": "password",
+                    "roles": ["test"],
+                },
+                "roles": [
+                    {
+                        "name": "test",
+                        "permission_sets_with_scope": [
+                            {
+                                "name": "DeployToDevices",
```
kjaskiewiczz, line 526: what about checking that configuration is not accessible for users with built-in Read Access role?

MaciejTe: Done!

backend-tests/tests/test_rbac_v2.py
```diff
@@ -439,19 +552,9 @@ def test_deploy_to_group(self, clean_mongo, test_case):
                 "user": {
                     "name": "test1-UUID@example.com",
                     "pwd": "password",
-                    "roles": ["test"],
+                    "roles": ["RBAC_ROLE_PERMIT_ALL"],
```
kjaskiewiczz, line 555: would be better to test it with test role containing "ConfigureDevices" permission

backend-tests/tests/test_rbac_v2.py
```diff
@@ -471,19 +588,32 @@ def test_deploy_to_group(self, clean_mongo, test_case):
...
+                            }
                         ],
-                    },
+                    }
                 ],
                 "device_groups": {"test": 1, "production": 1},
-                "deploy_group": "test",
-                "set_configuration_status_code": 204,
-                "deploy_configuration_status_code": 200,
+                "deploy_group": "production",
```
kjaskiewiczz, line 597: this way you test the scope, not the permission;
to test the permission, the deploy_group should be "test"

backend-tests/tests/test_rbac_v2.py
```diff
@@ -471,19 +588,32 @@ def test_deploy_to_group(self, clean_mongo, test_case):
...
+                    "roles": ["RBAC_ROLE_OBSERVER"],
+                },
+                "roles": [],
+                "device_groups": {"test": 1, "production": 1},
+                "deploy_group": "production",
+                "set_configuration_status_code": 403,
+                "deploy_configuration_status_code": 403,
+            },
+            {
+                "name": "Test RBAC configuration deployment forbidden for ConfigureDevices permission set",
```
kjaskiewiczz, line 616: would be good to also check if the configure permission allows to set the configuration

backend-tests/tests/test_rbac_v2.py
```diff
@@ -657,8 +815,21 @@ def test_set_and_deploy_configuration(self, clean_mongo, test_case):
...
+                    "pwd": "password",
+                    "roles": ["RBAC_ROLE_OBSERVER"],
+                },
+                "roles": [],
+                "device_groups": {"test": 1, "production": 1},
+                "view_group": "production",
+                "get_configuration_status_code": 403,
+            },
+            {
+                "name": "Test RBAC GET deployment configuration forbidden with ConfigureDevices permission set",
```
kjaskiewiczz, line 831: forbidden?

kjaskiewiczz (changes requested):

backend-tests/tests/test_rbac_v2.py
MaciejTe: Another test cases contains `ConfigureDevices` permission, in line 628

backend-tests/tests/test_rbac_v2.py
MaciejTe: Thanks, done!

backend-tests/tests/test_rbac_v2.py
MaciejTe: Thanks, done!

backend-tests/tests/test_rbac_v2.py
MaciejTe: Thanks, done!

kjaskiewiczz:
tests are failing, because user with "DeployToDevices" permission is trying to upload artifact;
it was possible before fix, but with https://github.com/mendersoftware/useradm-enterprise/pull/536 it is no longer possible;
there is also one issue with RBAC itself I'll address asap

kjaskiewiczz (approved):

tranchitella (approved):

PR #2130, opened 2022-09-01
fix: Bump mender-binary-delta to 1.4.x.

kacf:
Although it may seem wrong to do this in a patch release, the reality
is that we never documented the 1.3 series in the 3.3 mender-docs
site. Instead we documented 1.4. Since there is no strong dependency
relationship between mender-binary-delta and the rest of the
components anyway, it's better to change it to the one we document,
and then stick to that from now on.

Changelog: None
Ticket: None

### DIFF
--- a/other-components.yml
+++ b/other-components.yml
@@ -11,7 +11,7 @@ services:
         image: mendersoftware/mender-cli:1.8.x
 
     mender-binary-delta:
-        image: mendersoftware/mender-binary-delta:1.3.x
+        image: mendersoftware/mender-binary-delta:1.4.x
 
     mender-convert:
         image: mendersoftware/mender-convert:3.0.x


oleorhagen (approved):

oleorhagen:
I guess a bump in the `versions.json` file is warranted too?

oleorhagen:
> Although it may seem wrong to do this in a patch release, the reality is that we never documented the 1.3 series in the 3.3 mender-docs site. Instead we documented 1.4. Since there is no strong dependency relationship between mender-binary-delta and the rest of the components anyway, it's better to change it to the one we document, and then stick to that from now on.
> 
> Changelog: None Ticket: None
> 
> Signed-off-by: Kristian Amlie [kristian.amlie@northern.tech](mailto:kristian.amlie@northern.tech)

Story is a little different. `3.3` used to use `1.3`, but `1.4` was released later on and then the docs upgraded to follow `1.4`, but not integration.

lluiscampos (approved):

kacf:
> I guess a bump in the `versions.json` file is warranted too?

Yes, it does, good catch. I'm a little fuzzy on how we are keeping these in sync in general. I can't see anything in the release checklist about it... @oleorhagen, @lluiscampos, @mzedel: Do you know?

mzedel:
> > I guess a bump in the `versions.json` file is warranted too?
> 
> Yes, it does, good catch. I'm a little fuzzy on how we are keeping these in sync in general. I can't see anything in the release checklist about it... @oleorhagen, @lluiscampos, @mzedel: Do you know?

this should get bumped once the update is tagged here... in general the versions file relies on 2 things to happen:
  1. the release_tool to output the version information for the components (used in [here](https://github.com/mendersoftware/integration/blob/master/extra/release_info_generator.py#L77-L83) via `do_list_repos`)
2. a tag to be pushed to integration then [this](https://github.com/mendersoftware/integration/blob/master/.gitlab-ci.yml#L43-L44) should run

I guess we could add a manual trigger to the gitlab job, for situations like this one...

oleorhagen:
@kacf I guess that also means, single-component releases, like the one done for binary-delta by @tranchitella here, will need to trigger this job manually afterwards.

kacf:
Got it! Thanks for explaining!

> I guess we could add a manual trigger to the gitlab job, for situations like this one...

I think we can just leave it. It's not that important since they are only loosely coupled. And I think it would anyway be reverted at the next automatic update, or?

> I guess that also means, single-component releases, like the one done for binary-delta by @tranchitella here, will need to trigger this job manually afterwards.

Actually, what it means is that we didn't do it correctly. There is no version of integration which contains mender-binary-delta 1.4.0, because it was released outside of it. We should have used the release_tool and bumped integration so that they are bumped in sync, and one contains the other. This would also have triggered the job.

Let's just leave it for now, and hopefully we won't have a lot of these cases. Not to mention this could all change with our new client/backend split release plan.

PR #2149, opened 2022-09-15
License generation fixes

lluiscampos:


extra/release_tool.py
```diff
@@ -965,7 +965,7 @@ def query_execute_git_list(execute_git_list):
     return True
 
 
-def query_execute_list(execute_list):
+def query_execute_list(execute_list, env={}):
```
kacf, line 968: Shouldn't this be `None` instead? The documentation is a little ambiguous, but it states this:

> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of the default behavior of inheriting the current process’ environment.

So if it's `{}`, I read it as though you empty the environment completely. This could have some unforeseen consequences.

lluiscampos: I think you are right. I'll amend.

kacf (approved):

lluiscampos:
* 3.4.x

PR #2159, opened 2022-09-20
chore: Revert 36d1bba98e8275dae3b6f7ce84d331046ed5c1aa from master

alfrunes:


lluiscampos (changes requested):
We should do this revert in a different way, so that we don't have a single commit with the whole diff: we need to merge back the old `master` commit following ours strategy... @kacf is the expert on this

PR #2161, opened 2022-09-21
Staging

kacf:


alfrunes (commented):
There's no change associated with this PR. Shouldn't we "revert the revert" as well?

kacf:
Actually that's a method I didn't think about.

This PR does the same thing though, the special "ours" merge gets the history from master, but without the changes, so the revert has no effect. The advantage with this method is that "git blame" keeps correct history.

tranchitella (approved):

lluiscampos (approved):
As commented privately, could you put here in the PR the exact commands you used for the merge?

alfrunes (approved):

PR #2170, opened 2022-09-23
chore(release_info_generator.py): Explicitly fail when no pre-existing versions.json

lluiscampos:


mzedel (changes requested):
[Here](https://github.com/mendersoftware/integration/blob/master/.gitlab-ci.yml#L31) the script should take care of the initialization...
Since there might be other (non-generated) adjustments in the json file I would rather not ignore the absence of the existing json file but let the job fail instead. (+ any of the older releases will not be rebuilt, due to the version set [here](https://github.com/mendersoftware/integration/blob/master/extra/release_info_generator.py#L34))
Maybe the pass in the try block should rather go then 🙈

lluiscampos:
> [Here](https://github.com/mendersoftware/integration/blob/master/.gitlab-ci.yml#L31) the script should take care of the initialization... 
> ...

Yep. I saw that. I just wanted to run it locally.

> Maybe the pass in the try block should rather go then see_no_evil

Exactly! It fails "by accident", so I'll make it fail more explicitly following your suggestion :+1:

mzedel (approved):
👍 excellent

PR #2171, opened 2022-09-24
ci: publish:versions: use privileged SSH key instead of token

lluiscampos:
The token is not privileged as is meant to be used on status APIs. Use the SSH key that we have across the Mender group to avoid introducing a different privileged token and a risk of leak by mistake when rotating them.

By the looks of it, we have never run this step fully automated so this went unnoticed until the time of tagging the ongoing release.

Other minor beautifications.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -20,21 +20,26 @@ variables:
 publish:versions:
   stage: publish
   image: python:slim
-  variables:
-    GITHUB_TOKEN: $GITHUB_BOT_TOKEN
   before_script:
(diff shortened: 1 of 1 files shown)


mzedel (approved):

tranchitella (approved):

lluiscampos:
* 3.4.x

lluiscampos:
Unfortunately, this fix isn't enough. We indeed need a new token for this job, I don't see any other way while using the `hub` command. I'll update the `versions.js` file manually this time and I'll follow-up after the release.

PR #2182, opened 2022-10-07
test(backend): Support running tests against Azure blob storage

alfrunes:
Requires mendersoftware/mender-qa#616

### DIFF
--- a/backend-tests/run
+++ b/backend-tests/run
@@ -1,28 +1,35 @@
 #!/bin/bash
 
-HERE=$(dirname $(readlink -f "$0"))
-INTEGRATION_PATH=$(dirname "$HERE")
+INTEGRATION_PATH=$(git rev-parse --show-toplevel)
 export INTEGRATION_PATH
 
 DOWNLOAD_REQUIREMENTS="true"
 
-COMPOSE_CMD_BASE="docker-compose -p backend-tests \
-        -f $INTEGRATION_PATH/backend-tests/docker/docker-compose.backend-tests.yml
-        "
-COMPOSE_CMD_OPEN="docker-compose -p backend-tests \
+COMPOSE_FILES_BASE="-f $INTEGRATION_PATH/backend-tests/docker/docker-compose.backend-tests.yml"
+COMPOSE_FILES_OPEN=" \
         -f $INTEGRATION_PATH/docker-compose.yml \
         -f $INTEGRATION_PATH/docker-compose.demo.yml \
-        -f $INTEGRATION_PATH/docker-compose.storage.minio.yml
-        -f $INTEGRATION_PATH/extra/integration-testing/docker-compose.yml
-        -f $INTEGRATION_PATH/backend-tests/docker/docker-compose.backend-tests.yml
+        -f $INTEGRATION_PATH/docker-compose.storage.minio.yml \
+        -f $INTEGRATION_PATH/extra/integration-testing/docker-compose.yml \
+        -f $INTEGRATION_PATH/backend-tests/docker/docker-compose.backend-tests.yml \
         "
-COMPOSE_CMD_ENTERPRISE="${COMPOSE_CMD_OPEN} \
+COMPOSE_FILES_ENTERPRISE="${COMPOSE_FILES_OPEN} \
         -f $INTEGRATION_PATH/docker-compose.enterprise.yml \
         -f $INTEGRATION_PATH/extra/recaptcha-testing/tenantadm-test-recaptcha-conf.yml \
         -f $INTEGRATION_PATH/extra/smtp-testing/workflows-worker-smtp-test.yml \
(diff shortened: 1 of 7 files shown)


alfrunes:
@mender-test-bot start pipeline --pr mender-qa/616 --fast

alfrunes:
@mender-test-bot start pipeline --pr mender-qa/616 --fast

alfrunes:
The backend-integration test (enterprise) failure is a known error and can be ignored. This PR only contributes to the `*:azblob` tests.

backend-tests/run
```diff
@@ -1,28 +1,35 @@
...
 export INTEGRATION_PATH
 
 DOWNLOAD_REQUIREMENTS="true"
 
-COMPOSE_CMD_BASE="docker-compose -p backend-tests \
-        -f $INTEGRATION_PATH/backend-tests/docker/docker-compose.backend-tests.yml
-        "
-COMPOSE_CMD_OPEN="docker-compose -p backend-tests \
+COMPOSE_FILES_BASE="-f $INTEGRATION_PATH/backend-tests/docker/docker-compose.backend-tests.yml"
+COMPOSE_FILES_OPEN=" \
```
lluiscampos, line 9: We could make `COMPOSE_FILES_OPEN` extend `COMPOSE_CMD_BASE` instead of having `backend-tests/docker/docker-compose.backend-tests.yml` in the two lists.

backend-tests/run
```diff
@@ -36,6 +43,8 @@ PYTEST_REPORT_ENTERPRISE="--self-contained-html \
         --html=report_backend_integration_enterprise.html"
 PYTEST_REPORT=""
 
+PYTEST_EXTRA_ARGS=""
+
```
lluiscampos, line 47: Unused. This functionality is already captured with `USER_PYTEST_ARGS`

backend-tests/tests/pytest.ini
```diff
@@ -3,3 +3,5 @@
 # We keep it in xunit1 until we align the tooling between test frameworks
 [pytest]
 junit_family = xunit1
+markers =
+    storage_test: marks tests that uses artifact storage
```
lluiscampos, line 7: The only inconvenient that I see with this design is that we need to add the marker as we add tests in the future.

Another way to engineer this can be to have a marker for tests that need to be skept for Azure storage (similar as we skip tests when running in K8S). And with this way you don't risk missing tests (you only risk that your new test fails because cannot be run in such setup, and they you just mark it so that it gets skept).

backend-tests/tests/test_deployments.py
```diff
@@ -1510,7 +1518,8 @@ def test_deployment_status_update(self, clean_mongo):
         )
 
 
-class TestDeploymentsToGroupStatusUpdateEnterprise(TestDeploymentsStatusUpdateBase):
+@pytest.mark.storage_test
+class TestDeploymentsToGroupStatusUpdateEnterprise(_TestDeploymentsStatusUpdateBase):
```
lluiscampos, line 1522: Nice clean-ups :+1:

lluiscampos (approved):

backend-tests/run
alfrunes: The `docker-compose.backend-tests.yml` needs to be the last compose file in the list since it's an override. Do you want me to append `_BASE` to `_OPEN`?

backend-tests/run
alfrunes: It is used only when the `AZURE_STORAGE_CONTAINER_PREFIX` env. variable is set. I can reuse the existing `USER_PYTEST_ARGS` if that's preferred.

backend-tests/run
lluiscampos: It was a minor suggestion. On the high level it makes sense to inherit like `BASE` > `OS` > `ENTERPRISE`. But if this complicates the script let's leave it as it (as it was).

backend-tests/run
lluiscampos: Oh, I didn't realize. Yeah, I think it is cleaner if we reuse what we have :+1:

backend-tests/tests/pytest.ini
lluiscampos: Let's go with opt-in. You (backend team) are the owners of this test suite, so let's do it as you think is best!

backend-tests/run
alfrunes: I did some [small refactoring](https://github.com/mendersoftware/integration/compare/9979b7cbb6cef8732f564c42ae3a4c2ef6375cd7..d2941d4a1dfd3b71d7b3e40b390496fd9f0064fa) of the pytest arguments for the run script removing the `PYTEST_EXTRA_ARGS` variable.

alfrunes:
Started a new pipeline [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/662664834).

kjaskiewiczz (approved):

PR #2193, opened 2022-10-17
test(integration): Keepalive test ensure connection is open

alfrunes:
Use the dbus API to open a connection to the backend before proceeding with the assertion.

### DIFF
--- a/tests/tests/test_keepalive.py
+++ b/tests/tests/test_keepalive.py
@@ -131,6 +131,13 @@ class TestMenderClientKeepAlive:
 
         idle_connections_timeout_seconds = 15
         disable_keep_alive = False
+        # Make sure the device opens a connection
+        mender_device.run(
+            """dbus-send --print-reply --system \\
+              --dest=io.mender.AuthenticationManager \\
+              /io/mender/AuthenticationManager \\
+              io.mender.Authentication1.FetchJwtToken"""
+        )
         output = mender_device.run(
             "cat /proc/`pidof mender`/net/tcp | grep -E '[^:]+: [^ ]+ [^ ]+:01BB' | wc -l"
         )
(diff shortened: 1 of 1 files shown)


tests/tests/test_keepalive.py
```diff
@@ -131,6 +131,13 @@ def test_keepalive_idle_connections(self, monitor_commercial_setup_no_client):
 
         idle_connections_timeout_seconds = 15
         disable_keep_alive = False
+        # Make sure the device opens a connection
+        mender_device.run(
+            """dbus-send --print-reply --system \\
+              --dest=io.mender.AuthenticationManager \\
+              /io/mender/AuthenticationManager \\
+              io.mender.Authentication1.FetchJwtToken"""
+        )
```
merlin-northern, line 140: this is something I have been repeating for some time: we have to check that the device is connected before starting any tests. this will just fail the test, what we should do is: wait for the device to connect, since there is no point in running that test if device is not connected.

merlin-northern (changes requested):

tests/tests/test_keepalive.py
alfrunes: I'm not sure I understand what you mean. We want to make sure the device run a request (which also resets the keep alive timer) before proceeding with the assertion. By running `dbus-send --print-reply` it will block for a response, which the client sends only after it has made the request. How do you suggest we want to ensure the device has an open connection?

alfrunes: Ok, I think I understand what you mean. Would you prefer if I run this command in a loop, polling the dbus api every second (for a configurable timeout duration) in case the client hasn't started yet?

alfrunes: Actually, by further inspection, this line is actually polling the command for up to 25 minutes (default value) until the command passes.

merlin-northern:
```bash
# time { systemctl stop mender-client; dbus-send --print-reply --system \
>               --dest=io.mender.AuthenticationManager \
>               /io/mender/AuthenticationManager \
>               io.mender.Authentication1.FetchJwtToken; }
Error org.freedesktop.DBus.Error.ServiceUnknown: The name io.mender.AuthenticationManager was not provided by any .service files

real	0m0.075s
user	0m0.015s
sys	0m0.031s
```

it takes some time for the Client to start:

```bash
# time { systemctl start mender-client; dbus-send --print-reply --system               --dest=io.mender.AuthenticationManager               /io/mender/AuthenticationManager               io.mender.Authentication1.FetchJwtToken; }
```

so basically, yes: a loop. there are more tests that need that: all the remote terminal tests for instance. this should be a fixture and the test that require a connected client should not start until the correct setup for them is prepared.

merlin-northern: wow, I did not expect that, I remember it failing immediately. ok then, lets have longer than 25s -- client takes 15minutes to start in some cases.

PR #2196, opened 2022-10-19
fix: Mark generate-delta-worker as an Enterprise repository. 

kacf:

### DIFF
--- a/extra/release_tool.py
+++ b/extra/release_tool.py
@@ -257,6 +257,7 @@ BACKEND_SERVICES_ENT = {
     "mtls-ambassador",
     "devicemonitor",
     "mender-conductor-enterprise",
+    "generate-delta-worker",
 }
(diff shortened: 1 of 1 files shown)


merlin-northern:
> For some reason (which I don't understand) this only affects 3.3.x, which is why I made the fix here. But I will cherry-pick it to 3.4.x and master too.

@MaciejTe was it a decision not to add it here?

kacf:
> > For some reason (which I don't understand) this only affects 3.3.x, which is why I made the fix here. But I will cherry-pick it to 3.4.x and master too.
> 
> @MaciejTe was it a decision not to add it here?

Ah of course, it was not added here because it's not part of this release.

Ok, so not adding it here actually makes sense. However, we still have to, because we have to able to list components from the master branch, and it needs to know about the component then. I've checked the code and I'm fairly certain that this does not have any other side effects, so it should be safe.

PR #2225, opened 2022-11-08
perf: Run the tests under xdist with the file scheduler

oleorhagen:
This makes tests scheduled onto an xdist worker whilst respecting (to a large
degree) the files under which they are defined.

This is beneficial, as say tests under a class scope are otherwise sorted onto
different runners, and the time saved by having the same scope on all of them is
lost, as they will all have to do the common setup and teardown multiple times.

This can also apply within a given runner, if the scoped tests are not run sequentially.


oleorhagen:

PR #2229, opened 2022-11-17
chore(changelog-generator): Add support for ME project references

lluiscampos:


extra/changelog-generator/changelog-generator
```diff
@@ -21,7 +21,7 @@ LINKED_SHAS = {}
 # messages.
 SHA_TO_TRACKER = {}
 
-JIRA_REGEX = r"\[?(?:Jira:? *)?(?:https?://tracker.mender.io/browse/)?((?:INF|ARCHIVE|MEN|QA|MC)(?:-| )[0-9]+)\]?"
+JIRA_REGEX = r"\[?(?:Jira:? *)?(?:https?://tracker.mender.io/browse/)?((?:INF|ARCHIVE|MEN|QA|MC|ME)(?:-| )[0-9]+)\]?"
```
oleorhagen, line 24: Since were on the topic already, what's an `ARCHIVE` ticket?

Also, maybe we should split these out into a separate list? (Nit/suggestion)

oleorhagen (approved):

extra/changelog-generator/changelog-generator
lluiscampos: The `ARCHIVE` seems to be very old, but as we have tests for it I decided not to touch it.

> Also, maybe we should split these out into a separate list? (Nit/suggestion)

Something like this? https://github.com/mendersoftware/integration/compare/40110dbc919e28e22576dca5f1e1b6503760c0b1..7a907c0cab69b1df01f9bbc55ffa29d17e5f83f7

oleorhagen: Looks good!

kacf: I believe ARCHIVE tickets originally came from our old bug tracker, Redmine. I don't remember exactly how they were created, but it was related to the migration, somehow. Anyway, the changelog-generator originally comes from CFEngine's generator, where ARCHIVE tickets occasionally appeared. Probably we have seen the last of those by now though.

oleorhagen: @lluiscampos actually needs more ticket references, so that the array grows long enough for `black` to break it into multiple lines!

PR #2238, opened 2022-11-25
test(release-tool): Skip test in not master branches

lluiscampos:
Test `test_generate_release_notes_from_master` has limited value in release branches and it is painful to maintain (requires useless updates in component-maps just to get the new but unused repos).

### DIFF
--- a/extra/test_release_tool.py
+++ b/extra/test_release_tool.py
@@ -62,6 +62,17 @@ def is_staging():
         return not any(c in non_staging_components for c in content["services"])
 
 
+@pytest.fixture(scope="session")
+def is_master():
+    """Fixture to figure out if we are running the tests in master branch
+    """
+
+    with open(os.path.join(INTEGRATION_DIR, "docker-compose.yml")) as fd:
+        content = fd.read()
+        assert "mendersoftware/gui:" in content
+        return re.search("mendersoftware/gui:.*master", content)
+
+
 def run_main_assert_result(capsys, args, expect=None):
     testargs = [RELEASE_TOOL] + args
     with patch.object(sys, "argv", testargs):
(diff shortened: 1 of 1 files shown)


lluiscampos:
Needless to say, that this test is currently failing both in 3.4.x and 3.3.x. I vouch for just not running it there.

extra/test_release_tool.py
```diff
@@ -62,6 +62,17 @@ def is_staging():
         return not any(c in non_staging_components for c in content["services"])
 
 
+@pytest.fixture(scope="session")
```
oleorhagen, line 65: Hmm, does it really need to be `session` scoped?

extra/test_release_tool.py
```diff
@@ -76,13 +87,9 @@ def run_main_assert_result(capsys, args, expect=None):
...
 
-def test_version_of(capsys):
-    with open(os.path.join(INTEGRATION_DIR, "docker-compose.yml")) as fd:
-        content = fd.read()
-        assert "mendersoftware/gui:" in content
-        if re.search("mendersoftware/gui:.*master", content) is None:
-            # Skip this test for non-master branches.
-            pytest.skip("This test requires master tags in the docker-compose files.")
+def test_version_of(capsys, is_master):
+    if not is_master:
```
oleorhagen, line 91: Suggestion: How about https://docs.pytest.org/en/6.2.x/skipping.html#id1

oleorhagen (approved):

extra/test_release_tool.py
lluiscampos: (related to your comment above), I went for this approach to be able to use a `session` coped fixture. I think is safest (and slightly more optimal) to read the file only once and at before any test executes.

The safety point is that this test suite is about reading and writing these components-maps.yml files. If we make a mistake and a test does not clean properly, it could happen that `mendersoftware/gui` gets some other version and then the check will misbehave.

lluiscampos: In case it wasn't obvious: `skipif` cannot execute a `fixture`.

oleorhagen: Makes sense!

oleorhagen: > In case it wasn't obvious: skipif cannot execute a fixture.

TIL

lluiscampos:
* 3.4.x
* 3.3.x

kacf (commented):
Deja vu, I thought we had already done this.

PR #2245, opened 2022-11-28
test: compat: start sequentially.

merlin-northern:
Ticket: QA-443


merlin-northern:

merlin-northern:
I wanted to make something else out of it, but this was the fastest route.

tests/tests/test_compat.py
```diff
@@ -307,16 +311,46 @@ def compatibility_test_impl(self, env):
...
+            "2.3",
+            "2.4",
+            "2.5",
+            "2.6",
+            "3.0",
+            "3.1",
+            "mender-3.2",
+            "mender-3.3",
+            "mender-3.4",
+        ]:
```
lluiscampos, line 329: This list we will need to update after every release. Therefore:
* Set it in a single place, with a meaningful name like `COMPAT_MENDER_VERSIONS`
* Add a comment with instructions on how to update, like "After a new release of the client components, add the corresponding `extra/integration-testing/docker-compose.compat-X-X.yml` and update this list" ...

merlin-northern: fixed. thank you.

merlin-northern:
@mender-test-bot start pipeline please sugar sugar

tests/tests/test_compat.py
mzedel: you could also get the versions file and generate the list of versions from there...

merlin-northern: wait, sorry, what versions file?

lluiscampos: I think he meant the composition files, which is a nice maintainability improvement.

extra/integration-testing/docker-compose.compat-2.0.yml
```diff
@@ -0,0 +1,17 @@
...
+    tty: true
+    privileged: true
+    environment:
+      TENANT_TOKEN: ""
+    volumes:
+      # Provision demo certificate
+      - ./cert/cert.crt:/mnt/config/server.crt
+      # Add updated setup that puts the new demo certificate in the global ca trust store
+      - ./extra/integration-testing/setup-mender-configuration.py:/setup-mender-configuration.py
+
```
lluiscampos, line 17: I suggest to put all the yaml files for this test in a subdir, something like `extra/integration-testing/test-compat/...`

The parent directory, `extra/integration-testing` contains files used by other parts (`extra/integration-testing/docker-compose.yml` is used by backend tests!) so I think it will look tidier to get these used by a single test in a subdir.

tests/tests/test_compat.py
```diff
@@ -34,56 +34,80 @@
 container_factory = factory.get_factory()
 
 TIMEOUT = timedelta(minutes=5)
+"""
+ COMPAT_MENDER_VERSIONS array stores the client version we test against.
```
lluiscampos, line 38: Technically is more than the "client" version as it is really a combination of client and addons. I encourage to avoid the "client" keyword as much as possible. For example
```suggestion
 COMPAT_MENDER_VERSIONS array stores the versions of the virtual device images we test against.
```

tests/tests/test_compat.py
```diff
@@ -34,56 +34,80 @@
...
+        )
+        env.populate_clients()
+
+        clients = env.get_mender_clients()
+        assert len(clients) > 0, "Failed to setup clients"
+        env.devices = []
+        for client in clients:
+            dev = MenderDevice(client)
+            dev.ssh_is_opened()
+            env.devices.append(dev)
```
lluiscampos, line 76: Should this expect only one client? I would make this strict with something like:

```suggestion
        clients = env.get_mender_clients()
        assert len(clients) == 1, "Failed to setup clients"
        env.device = MenderDevice(clients[0])
        env.device.ssh_is_opened()
```

lluiscampos (approved):

tests/tests/test_compat.py
merlin-northern: fixed. we not only return one device and expect one device.

tests/tests/test_compat.py
merlin-northern: sure. fixed.

extra/integration-testing/docker-compose.compat-2.0.yml
merlin-northern: sure fixed. also: we now scan the directory and get the versions from there.

tests/tests/test_compat.py
merlin-northern: I do not think so :) "versions file" != "composition files" what I did now is: scan the directory for the versions.

mzedel: yeah I meant the json file, and then generate the composition file names or even files from there - they should surely follow a naming pattern and we would have to rely on a single place to update instead of adding more items to the (post) release checklist...

mzedel: but yeah scanning the directory is also a good option!

merlin-northern:
@mender-test-bot start pipeline please sweet sugar sugar

tests/tests/test_compat.py
lluiscampos: > yeah I meant the json file

One advantage of using the json file would be than after a new release, even the older servers (iow release branches from integration repo) would run the test with newer clients :rocket: However as we move out to rolling releases for the backend, then this specific advantage will lose value.

In any case the reason I thought you didn't mean this is because Piotr already tried to generate the compose files based on a template and it didn't work out. Due to the certificate changes and other minor things we have done along the way, different images need different tweaks to get them running.

Scanning the directory is a good improvement regardless. Well done @merlin-northern

tests/tests/test_compat.py
```diff
@@ -34,56 +34,69 @@
...
+        env.populate_clients()
+
+        clients = env.get_mender_clients()
+        assert len(clients) == 1, "Failed to setup clients: expecting exactly one."
+        env.devices = []
+        for client in clients:
+            dev = MenderDevice(client)
+            dev.ssh_is_opened()
+            env.devices.append(dev)
+            break
```
kacf, line 64: Minor: You already know that `len(clients) == 1`. Suggest removing the loop and the break and replacing with `dev = MenderDevice(clients[0])`.

tests/tests/test_compat.py
```diff
@@ -34,56 +34,69 @@
...
 
-    env.populate_clients(tenant_token=env.tenant.tenant_token)
+        clients = env.get_mender_clients()
+        assert len(clients) == 1, "Failed to setup clients: expecting exactly one."
+        env.devices = []
+        for client in clients:
+            dev = MenderDevice(client)
+            dev.ssh_is_opened()
+            env.devices.append(dev)
+            break
```
kacf, line 95: Same here, ref `dev = MenderDevice(clients[0])`.

extra/integration-testing/test-compat/docker-compose.compat-2.0.yml
```diff
@@ -0,0 +1,17 @@
+version: '2.1'
```
kacf, line 1: What about the files we already have? Shouldn't these be moved, rather than copied?

kacf (commented):
Good work!

tests/tests/test_compat.py
merlin-northern: fixed.

tests/tests/test_compat.py
merlin-northern: fixed.

merlin-northern:
@mender-test-bot start pipeline please sweet sugar sugar

extra/integration-testing/test-compat/docker-compose.compat-2.0.yml
kacf: Oh sorry, I was viewing with `-C` and git tricked me. Strike that.

New question instead: Do we actually need the template? Looks to me like we can just remove it.

merlin-northern: no worries. thought to preserve for reference. removed.

merlin-northern:
@mender-test-bot start pipeline pretty please sweet sugar sugar

extra/integration-testing/test-compat/docker-compose.compat-2.0.yml
kacf: The same content is in the new files you created, so I think removing it is fine! 👍

kacf (approved):

lluiscampos (approved):

merlin-northern:
@mender-test-bot start pipeline pretty please sweet sugar sugar

merlin-northern:
sorry last force push is restoring `extra/integration-testing/docker-compose.mender.2.5.yml` which I moved by mistake.

merlin-northern:
failures:
* enterprise: `test_monitorclient_alert_store` is unrelated and seems like an unlikely race
* opensource: known docker rm network error

merlin-northern:
@mender-test-bot  cherry-pick to:
 * 3.4.x
 * 3.3.x

PR #2259, opened 2022-12-06
test: Include demo cert in generate-delta-artifact in demo

merlin-northern:
Ticket: None


merlin-northern:
we will probably go with skip verify (https://github.com/mendersoftware/integration/pull/2268)

PR #2262, opened 2022-12-08
QA-507: Use openssl1.1-compat package instead of default one

lluiscampos:
* chore: Remove deprecated depencencies for integ tests
    
    Most of these were added because some pip packages (cryptography?) used
    to build their dependencies from source. In theory we should not need
    any kind of build related tools for running integration tests. Keep
    `make` which is used to install update module generators.
    
    Removed also filesystem images manipulation tools like e2fsprogs and
    device-mapper.

* fix: Use openssl1.1-compat package instead of default one
    
    Which for Alpine 3.17 on defaults on openssl 3, incompatible with
    mender-artifact pre-built binary.


tests/requirements/apk-requirements.txt
```diff
@@ -2,20 +2,9 @@ bash
...
 py-pip
 iptables
-gcc
-libffi-dev
-libc-dev
-openssl-dev
 make
 python3
-python3-dev
-e2fsprogs-extra
```
kacf, line 15: I would have thought that e2fsprogs was used by mender-artifact (debugfs). But maybe it never calls the sub commands that use it.

kacf (approved):

tests/requirements/apk-requirements.txt
lluiscampos: I also suspected they were used... but the evidence of all tests passing in here, in 3.4.x and in 3.3.x proved otherwise :)

PR #2267, opened 2022-12-09
autogenerate delta test

kjaskiewiczz:
Ticket: [MEN-5806](https://tracker.mender.io/browse/MEN-5806)
The test looks very basic, but it's actually testing quite a lot of things - deployments creation with new flag, preparation of delta request (including very problematic selection of currently installed artifact), lunching delta request, delta worker job execution and application of delta updates.
I can imaging couple of more complicated scenarios (some corner cases), but I decided to not implement them here because:
- it's quite complex (poor excuse; this maybe only me struggling with the tests)
- it's costly - in the full integration we run qemu clients, and to check some corner cases I'd need to run more than one for sure;
- if the feature will be tested fully and only using integration tests, then we should run full integration every time we change something in the deployments-enterprise and generate-delta-worker
  - this will be true for generate-delta-worker anyway; I can't imaging verifying generate-delta-worker functionality without applying generated delta; this may not be a huge issue, since I wouldn't expect many changes in this repo;

I'll create follow-up task for testing corner cases. Ideally, this should be done in the deployments-enterprise acceptance tests, where we can mock e.g. generate delta worker.


kjaskiewiczz:

tests/tests/test_autogenerate_delta.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # check if client is applying delta
+        isDelta = False
+
+        def deployment_triggered_callback():
+            # wait a minute
+            i = 60
+            while i > 0:
+                output = mender_device.run(
+                    "systemctl status --no-pager -l -n 1000000 mender-client | grep 'xdelta3'",
```
tranchitella, line 123: A bit flaky as a detection mechanism, however as we know we cannot use the Server's APIs to get this information just yet. However, once https://tracker.mender.io/browse/MEN-6138 is done, can't we do it that way?

tranchitella (approved):

tests/tests/test_autogenerate_delta.py
kjaskiewiczz: yes, I don't like this part too, and this is why I was trying to use backend API; I can see the MEN-6138 is in progress so I can wait and update the test (probably today)

kjaskiewiczz: or I'll just follow up with that; I'll have to run pipeline once again anyway

docker-compose.client.rofs.commercial.yml
```diff
@@ -0,0 +1,8 @@
+version: '2.1'
+services:
+
+    #
+    # mender-client
+    #
+    mender-client:
+        image: registry.mender.io/mendersoftware/mender-qemu-rofs-commercial:mender-master
```
lluiscampos, line 8: It should now be removed from https://github.com/mendersoftware/integration/blob/39941fe9712c64011362aae2e86dc241d5b6c59d/other-components-docker.yml#L9-L10

tests/run.sh
```diff
@@ -175,6 +175,8 @@ docker run --rm --privileged --entrypoint /extract_fs -v $PWD/output:/output \
         mendersoftware/mender-client-qemu-rofs:$(../extra/release_tool.py --version-of mender-client-qemu-rofs --version-type docker)
 docker run --rm --privileged --entrypoint /extract_fs -v $PWD/output:/output \
         registry.mender.io/mendersoftware/mender-gateway-qemu-commercial:$(../extra/release_tool.py --version-of mender-gateway --version-type docker)
+docker run --rm --privileged --entrypoint /extract_fs -v $PWD/output:/output \
+         registry.mender.io/mendersoftware/mender-qemu-rofs-commercial:$(../extra/release_tool.py --version-of mender-qemu-rofs-commercial --version-type docker)
```
lluiscampos, line 179: :sob: https://tracker.mender.io/browse/QA-512

tests/tests/test_autogenerate_delta.py
```diff
@@ -0,0 +1,155 @@
...
+        mender_device = env.device
+        devauth = DeviceAuthV2(env.auth)
+        deploy = Deployments(env.auth, devauth)
+
+        mender_conf = mender_device.run("cat /etc/mender/mender.conf")
+        valid_image_rofs = valid_image_rofs_commercial_with_mender_conf(mender_conf)
+        if valid_image_rofs is None:
+            pytest.skip(
+                "Thud branch and older from Yocto does not have R/O rootfs support"
+            )
```
lluiscampos, line 54: `thud` has been out of support for years now, what is this check about?

testutils/infra/container_manager/factory.py
```diff
@@ -183,8 +188,11 @@ def get_enterprise_docker_client_setup(self, name=None, num_clients=0):
     def get_enterprise_rofs_client_setup(self, name=None, num_clients=0):
         return DockerComposeEnterpriseRofsClientSetup(name, num_clients)
 
-    def get_compatibility_setup(self, tag, name=None, **kwargs):
-        return DockerComposeCompatibilitySetup(name, tag, **kwargs)
+    def get_enterprise_rofs_commercial_client_setup(self, name=None, num_clients=0):
+        return DockerComposeEnterpriseRofsCommercialClientSetup(name, num_clients)
+
+    def get_compatibility_setup(self, name=None, **kwargs):
+        return DockerComposeCompatibilitySetup(name, **kwargs)
```
lluiscampos, line 195: Committed by mistake? What happened to `tag` here?

PR #2268, opened 2022-12-12
chore: adding generate delta worker to enterprise tests composition

merlin-northern:
Ticket: None

### DIFF
--- a/docker-compose.testing.enterprise.yml
+++ b/docker-compose.testing.enterprise.yml
@@ -10,3 +10,8 @@ services:
     mender-tenantadm:
         environment:
             TENANTADM_ORCHESTRATOR_ADDR: http://mender-workflows-server:8080/
+
+    mender-generate-delta-worker:
+        command: --automigrate
+        environment:
+            - GENERATE_DELTA_SKIPVERIFY=1


merlin-northern:
@lluiscampos this if just fyi

merlin-northern:

merlin-northern:
@kjaskiewiczz could you run your pipelines from MEN-5806 with this?

docker-compose.demo.yml
```diff
@@ -94,3 +94,8 @@ services:
         command: --automigrate
         environment:
             - CREATE_ARTIFACT_SKIPVERIFY=1
+
+    mender-generate-delta-worker:
```
kjaskiewiczz, line 98: this will prevent starting open source demo, because mender-generate-delta-worker is not an open source component

lluiscampos: Yes, this cannot go here. Generally speaking we don't have an "enterprise demo" mode. We only have this `--enterprise-testing` flag that we use internally (hence it is not documented).

As this is only needed for tests, it should go into [`docker-compose.testing.enterprise`](https://github.com/mendersoftware/integration/blob/master/docker-compose.testing.enterprise.yml) instead.

lluiscampos (changes requested):

docker-compose.demo.yml
merlin-northern: yes, and also breaks the os test, I also forgot that this is ent only. fixed. sorry for that.

kjaskiewiczz (approved):

merlin-northern:
some surious failures, unrelated I would say.

PR #2270, opened 2022-12-13
feat: Enable changelog-generator to get data from GitHub instead of Git.

kacf:
Changelog: None
Ticket: QA-342

### DIFF
--- a/extra/changelog-generator/changelog-generator
+++ b/extra/changelog-generator/changelog-generator
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
 # Used to generate changelogs from the repository.
 
--- a/extra/changelog-generator/changelog-generator
+++ b/extra/changelog-generator/changelog-generator
@@ -114,6 +114,13 @@ parser.add_argument(
     + "Ignored if using --repo",
 )
 parser.add_argument(
+    "--query-github",
+    dest="query_github",
+    action="store_true",
+    help="Query Github for commit messages instead of local repositories. "
+    + "Set GITHUB_TOKEN if you want authorized access.",
+)
+parser.add_argument(
     "--sort-changelog",
     dest="sort_changelog",
     action="store_true",
@@ -139,6 +146,67 @@ parser.add_argument(
     + "Note that for technical reasons these must come last.",
 )
 
+
+class GitQuerier:
+    def __init__(self, gitargs, base_dir):
+        self.gitargs = gitargs
+        self.base_dir = base_dir
+
+    def get_commits_for_range(self, repo, range):
+        output = subprocess.check_output(
+            ["git", "rev-list", "--reverse", range] + self.gitargs,
+            cwd=os.path.join(self.base_dir, repo),
+        )
+        return output.decode().split()
+
+    def get_raw_commit_message(self, repo, sha):
+        output = subprocess.check_output(
+            ["git", "cat-file", "-p", sha] + self.gitargs,
+            cwd=os.path.join(self.base_dir, repo),
+        )
+        # Return only the part after headers (double newline).
+        return output.decode().split("\n\n", 1)[1]
+
+    def show_commit_without_diff(self, repo, sha):
+        output = subprocess.check_output(
+            ["git", "show", "-s", sha] + self.gitargs,
+            cwd=os.path.join(self.base_dir, repo),
+        )
+        return output.decode()
+
+
+class GitHubQuerier:
+    def __init__(self):
+        # Avoid importing this unless we have to, which is only when using this
+        # class.
+        import github
+
+        if os.getenv("GITHUB_TOKEN"):
+            self.github = github.Github(os.getenv("GITHUB_TOKEN"))
+        else:
+            self.github = github.Github()
+
+    def get_commits_for_range(self, repo, range):
+        base, head = range.split("..", 1)
+        # Cache the list, which has a lot of information, otherwise we need to
+        # query every commit later.
+        self.cached_result = self.github.get_repo(f"mendersoftware/{repo}").compare(
+            base, head
+        )
+        return [commit.sha for commit in self.cached_result.commits]
+
+    def get_raw_commit_message(self, repo, sha):
+        for commit in self.cached_result.commits:
+            if commit.sha == sha:
+                return commit.commit.message
+        raise Exception(f"Could not find {sha} in the cache")
+
+    def show_commit_without_diff(self, repo, sha):
+        # No way to return it exactly as Git does without manual piecing
+        # together. This is only used for information, so return this instead.
+        return self.get_raw_commit_message(repo, sha)
+
+
 args = parser.parse_args()
 
 if args.all and args.repo:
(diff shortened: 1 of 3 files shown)


extra/changelog-generator/changelog-generator
```diff
@@ -139,6 +146,73 @@ parser.add_argument(
...
+        output = subprocess.check_output(
+            ["git", "cat-file", "-p", sha] + self.gitargs,
+            cwd=os.path.join(self.base_dir, repo),
+        )
+        # Return only the part after headers (double newline).
+        return output.decode().split("\n\n", 1)[1]
+
+    def show_commit_without_diff(self, repo, sha):
+        output = subprocess.check_output(
+            ["git", "show", "-s", sha] + self.gitargs,
```
oleorhagen, line 172: Nit: `-s` = `--no-patch`

extra/changelog-generator/changelog-generator
```diff
@@ -151,7 +225,7 @@ repos = []
 if args.base_dir:
     base_dir = args.base_dir
 else:
-    base_dir = os.path.realpath(os.path.join(os.path.dirname(sys.argv[0]), "../../.."))
+    base_dir = os.path.realpath("..")
```
oleorhagen, line 228: :ninja:

extra/changelog-generator/changelog-generator
```diff
@@ -167,9 +241,18 @@ if args.all:
         .decode()
         .strip()
     )
-    repos += [os.path.join(base_dir, repo) for repo in repo_list.split()]
+    repos += repo_list.split()
 else:
-    repos.append(".")
+    repos.append(os.path.basename(os.path.realpath(".")))
+
+if not args.base_dir and not args.repo and not args.query_github:
```
oleorhagen, line 248: Just fine, we could play with:
```
not any([args.base_dir, args.repo, args.query_github])
```
But feel free to ignore. It's `_Pythonic_`, but  whether it's _actually_ more readable... 

You be the judge :smile_cat:

extra/changelog-generator/test-changelog-generator
```diff
@@ -746,6 +746,67 @@ then
...
+cat > expected.txt <<'EOF'
+### Changelogs
+
+#### mender (3.4.0)
+
+New changes in mender since 3.3.0:
+
+##### Bug Fixes
+
+* Append log entries to syslog at the correct level
```
oleorhagen, line 774: Too many bugs :see_no_evil:

extra/changelog-generator/changelog-generator
```diff
@@ -167,9 +241,18 @@ if args.all:
...
+    print("Need at least one of --base-dir, --repo or --query-github")
+    sys.exit(1)
+
+if args.query_github:
```
oleorhagen, line 252: Idea: 

Could we drop the `query_github` check, and do:

` locally` -> `!found` -> `github`

? :thinking:

oleorhagen (approved):
Nice work!

Good way to start the day :sun_with_face:

extra/changelog-generator/changelog-generator
```diff
@@ -139,6 +146,73 @@ parser.add_argument(
...
+        if os.getenv("GITHUB_TOKEN"):
+            self.headers["Authorization"] = "Bearer %s" % os.getenv("GITHUB_TOKEN")
+
+    def get_commits_for_range(self, repo, range):
+        range = range.replace("..", "...")
+        response = self.requests.get(
+            f"https://api.github.com/repos/mendersoftware/{repo}/compare/{range}",
+            headers=self.headers,
+        )
+        assert response.status_code == 200, f"Status code: {response.status_code}"
```
mzedel, line 198: since we're installing & importing a module already, would it not be better to go for [PyGithub](https://github.com/PyGithub/PyGithub) and then use the [compare method](https://pygithub.readthedocs.io/en/latest/github_objects/Repository.html#github.Repository.Repository.compare) here and have access to the other utils? that would also help with keeping this up with Github API changes (which looking at their versioning scheme, _might_ be a bit more frequent...)

mzedel (commented):
Considering that we rely on Github for the git part anyways, I very much appreciate this direction!

extra/changelog-generator/changelog-generator
```diff
@@ -139,6 +146,73 @@ parser.add_argument(
...
+    def get_commits_for_range(self, repo, range):
+        output = subprocess.check_output(
+            ["git", "rev-list", "--reverse", range] + self.gitargs,
+            cwd=os.path.join(self.base_dir, repo),
+        )
+        return output.decode().split()
+
+    def get_raw_commit_message(self, repo, sha):
+        output = subprocess.check_output(
+            ["git", "cat-file", "-p", sha] + self.gitargs,
```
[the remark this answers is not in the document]
kacf: This doesn't have a long option, actually. It's "pretty print", not "paginate", but there is no such option.

extra/changelog-generator/changelog-generator
```diff
@@ -139,6 +146,73 @@ parser.add_argument(
...
+
+    def show_commit_without_diff(self, repo, sha):
+        output = subprocess.check_output(
+            ["git", "show", "-s", sha] + self.gitargs,
+            cwd=os.path.join(self.base_dir, repo),
+        )
+        return output.decode()
+
+
+class GitHubQuerier:
```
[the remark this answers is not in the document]
kacf: I was asking myself the exact same question, and I wanted to do it. But interfaces are kind of useless in Python, they give you nothing:

1. Because of duck typing it makes no difference when you have a variable referring to it.
2. Because there is no real interface/abstract type, but just superclass functions that do nothing, it cannot even warn you if you forget to implement something.

So in the end I decided it's not helpful and omitted it instead. I could include it for documentation purposes of course, but I thought it wasn't really worth it, and it could go quickly out of date too.

extra/changelog-generator/changelog-generator
kacf: It *is* more readable! I will change this.

extra/changelog-generator/changelog-generator
kacf: Good idea! It's a very clean API, so the code is almost the same.

extra/changelog-generator/changelog-generator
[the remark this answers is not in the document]
oleorhagen: It's fine :) 

There is the `abc` package  (abstract base class). But it's not important.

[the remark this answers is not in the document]
kacf: Alright then.

But TIL, I didn't know about `abc`. I'll keep that in mind for next time.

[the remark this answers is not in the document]
lluiscampos: FWIW, the Python way of specifying interfaces is using protocol classes. See https://docs.python.org/3/library/typing.html#typing.Protocol

lluiscampos (approved):
I only had a high-level review as most of the detailing I assume it has been done by Ole.

Looks good!

extra/changelog-generator/changelog-generator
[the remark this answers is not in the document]
kacf: I think `Protocol` is different, it is for specifying static type checks. Also useful, but independent of abstract base classes, and I think a bit overkill for this PR.

In the end I used `abc` to make an abstract base class.

PR #2276, opened 2022-12-16
ci: Use GitHub token in addition to SSH key and fix hub command

lluiscampos:
This commit partially reverts 7bb7fc0. It uses a different token because the previous one only had repo:status permissions.

It took me a while to realize that we actually need both the token and the SSH key. See this comment:
* https://github.com/github/hub/issues/1644#issuecomment-359002547

Additionally, fix the hub command: a pull request needs to be launched
from a different fork or branch.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -22,6 +22,7 @@ publish:versions:
   image: python:slim
   before_script:
     - apt-get update && apt-get install -qqy curl hub unzip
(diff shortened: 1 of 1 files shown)


lluiscampos:
Relates to:
* https://github.com/mendersoftware/integration/pull/2171

mzedel (approved):
Nice find there!

lluiscampos:
> Nice find there!

Thank you. Although it wasn't working just yet. I think _now_ it should be fine

PR #2301, opened 2023-01-04
build(docker-client): Add support for building the cpp-client

oleorhagen:
Ticket: MEN-5796

### DIFF
--- a/extra/mender-client-docker-addons/Dockerfile
+++ b/extra/mender-client-docker-addons/Dockerfile
@@ -4,13 +4,14 @@ FROM ubuntu:22.04 AS build
 
 ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y make git build-essential golang liblzma-dev \
-    jq libssl-dev libglib2.0-dev curl
+    jq libssl-dev libglib2.0-dev curl cmake liblmdb++-dev
 
 ARG MENDER_CLIENT_REV=master
 ARG MENDER_CONNECT_REV=master
 
 RUN git clone https://github.com/mendersoftware/mender /src/mender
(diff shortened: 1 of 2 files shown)


oleorhagen:
Follow up on: https://github.com/mendersoftware/integration/pull/2300

kacf:
I suggest @lluiscampos reviews this one since he already looked into this on master.

extra/mender-client-docker-addons/Dockerfile
```diff
@@ -38,12 +42,12 @@ RUN mender-artifact write bootstrap-artifact \
         --output-path /bootstrap.mender
 
 WORKDIR /src/mender-connect
-RUN make prefix=/mender-install install
+RUN DESTDIR=/mender-install make prefix=/usr install
```
oleorhagen, line 45: I'm guessing you want this reverted to the old setup here?

But in a sense, it bothers me that we don't do this properly.

Would it be off if I fixed this in the given repos?

lluiscampos (changes requested):

oleorhagen:
Obviously the `extra-tools` test failes. This is pre-existing. I can fix.

Question is rather. Do we want to?

lluiscampos:
> Obviously the `extra-tools` test failes. This is pre-existing. I can fix.
> 
> Question is rather. Do we want to?

@oleorhagen Maybe just disable this CI job for the feature branch.

With regards to the license update,

What you should do is actually merge the current master into the feature branch (or rather fast-forward if we just created it now after this issue) so that you get the license update and also the apply + revert of this `Dockerfile`, so that you new commit goes on top of that and we don't get the conflict later.

lluiscampos (approved):

lluiscampos (approved):

PR #2307, opened 2023-01-06
test: Do not reuse CompatibilySetup for Mender 2.5 setup

lluiscampos:
There are a set of tests that specifically verify `mender-connect` 1.0 with current server versions, and for historical reasons they were reusing the `DockerComposeCompatibilitySetup`. Leave the latter for the rather specific test setup used there, and create new (smipler?) setups for the tests requiring only this specific client.

This commit workarounds the problem for QA-525 by not using this setup in other tests. The root cause for QA-525 is still unknown.

### DIFF
--- a/tests/tests/test_filetransfer.py
+++ b/tests/tests/test_filetransfer.py
@@ -1,4 +1,4 @@
-# Copyright 2022 Northern.tech AS
+# Copyright 2023 Northern.tech AS
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


lluiscampos:
@mender-test-bot start pipeline :rock:

kacf (approved):
I *think* it's correct. But yeah, the original cause is unknown, so I don't understand it 100%. The clean separation should help though.

lluiscampos:
:unicorn: Green pipeline!

PR #2310, opened 2023-01-13
ignore: test: dbg

merlin-northern:
Ticket: QA-510


lluiscampos:
Fixed in #2319 :+1:

PR #2314, opened 2023-01-26
fix: bump docker compose file format version to 2.3

kjaskiewiczz:
This is because the start_period option which we're started to use was added in file format 2.3 (see the [docs](https://docs.docker.com/compose/compose-file/compose-file-v2/#healthcheck))

### DIFF
--- a/backend-tests/docker/docker-compose.backend-tests.yml
+++ b/backend-tests/docker/docker-compose.backend-tests.yml
@@ -1,4 +1,4 @@
-version: '2.1'
+version: '2.3'
 services:
     mender-backend-tests-runner:
         image: mendersoftware/mender-test-containers:backend-integration-testing
--- a/common.yml
+++ b/common.yml
@@ -1,4 +1,4 @@
-version: '2.1'
+version: '2.3'
 services:
     mender-base:
         stdin_open: false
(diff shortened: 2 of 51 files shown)


0lmi (approved):
Thanks for fixing it 👍

lluiscampos (changes requested):
I made some local experiments and this will fix the issue, actually. It is just not testable from CI due to [QA-348
](https://tracker.mender.io/browse/QA-348) :see_no_evil: 

The CI in `mender-docs` repo uses docker-compose version 1.25.0 an docker 20.10.12, which [supports compose format 2.3](https://docs.docker.com/compose/compose-file/compose-versioning/#version-23) :+1: 

We need to set it up in all docker-compose files, however. Otherwise you'll get errors like:
```
root@1296fe3a3431:/work# docker-compose -f docker-compose.yml -f docker-compose.storage.minio.yml up
ERROR: Version mismatch: file ./docker-compose.yml specifies version 2.3 but extension file /work/common.yml uses version 2.1
```

So please update all docker-compose file, we run the pipeline here to make sure integration tests still pass, and then merge with the broken tests on this specific job (again, due to the QA task linked above that we don't have a reliable way to test before merge).

lluiscampos (approved):

PR #2315, opened 2023-01-30
MEN-6086: add app-update-module repo to the release tool

0lmi:
Changelog: None
Ticket: MEN-6086
Signed-off-by: Alex Miliukov <oleksandr.miliukov@northern.tech>

### DIFF
--- a/component-maps.yml
+++ b/component-maps.yml
@@ -288,6 +288,12 @@ git:
     release_component: false
     independent_component: true
 
+  app-update-module:
+    docker_image: []
+    docker_container: []
+    release_component: false
+    independent_component: true
+
 docker_image:
 
   iot-manager:


component-maps.yml
```diff
@@ -288,6 +288,12 @@ git:
     release_component: false
     independent_component: true
 
+  app-update-module:
+    docker_image: []
+    docker_container: []
+    release_component: true
+    independent_component: true
+
```
merlin-northern, line 296: I feel like Lluis is the expert here, I would defer to him, unless he does not have the time, then I will review (it will take me much, much longer)

component-maps.yml
```diff
@@ -288,6 +288,12 @@ git:
     release_component: false
     independent_component: true
 
+  app-update-module:
+    docker_image: []
+    docker_container: []
+    release_component: true
```
lluiscampos, line 294: We will not release it together with the rest of Mender. So this one needs to be `false`.

As per today we are in early steps to redesign completely the release process, and `components-maps` as we know it will change substantially. But for now we need to deal with what we have :)

lluiscampos (changes requested):

0lmi:
@mender-test-bot start pipeline --fast

lluiscampos (approved):

PR #2316, opened 2023-01-31
QA-461: test: Improve robustness of test_keepalive_disable

lluiscampos:
The mender client will always do an inventory update on boot, which is what presumably that 1s sleep was trying to wait for.

Increase considerably the sleep, while making the poll intervals long so that no new inventory update (nor update check) happen for the rest of the test.

This does not modify the goal of the test, which is to verify that the client respects keep alive settings (specifically, that disabling it does not keep TCP connections lingering around).

### DIFF
--- a/tests/tests/test_keepalive.py
+++ b/tests/tests/test_keepalive.py
@@ -192,9 +192,14 @@ class TestMenderClientKeepAlive:
         )
         assert get_opened_tcp_connections(mender_device) > 0
 
-        configure_connectivity(mender_device, disable_keep_alive=disable_keep_alive)
+        configure_connectivity(
+            mender_device,
+            disable_keep_alive=disable_keep_alive,
+            inventory_poll=3600,
+            update_poll=3600,
+        )
         logger.info("test_keepalive_disable: waiting for client to restart")
-        time.sleep(1)
+        time.sleep(15)
         assert get_opened_tcp_connections(mender_device) == 0
         logger.info("test_keepalive_disable: ok, no connections to backend")
 


tests/tests/test_keepalive.py
```diff
@@ -192,9 +192,14 @@ def test_keepalive_disable(self, monitor_commercial_setup_no_client):
...
-        configure_connectivity(mender_device, disable_keep_alive=disable_keep_alive)
+        configure_connectivity(
+            mender_device,
+            disable_keep_alive=disable_keep_alive,
+            inventory_poll=3600,
+            update_poll=3600,
+        )
         logger.info("test_keepalive_disable: waiting for client to restart")
-        time.sleep(1)
+        time.sleep(15)
```
merlin-northern, line 202: no magic numbers please :) could we define variables for them?

lluiscampos: I could define a variable, but it will still be an arbitrary number which I think it just hides away the problem (which IMO is worst than being explicit and honest about it).

A different approach however is to retry the check until it passes or we reach a timeout. This we could do but I wanted to respect the original intention of the test in that regard.

lluiscampos:
@mender-test-bot start pipeline :fast_forward:

tests/tests/test_keepalive.py
0lmi: I think if we know that this is a hiding a problem and know how to rework it in a better way, then it's the way to go. This is agile approach and an original implementer should not be hurt by that.

0lmi (approved):
Any of the discussed approach is a step into improvements direction, so, I believe it worth to merge.

PR #2324, opened 2023-02-07
chore: disable addons by default when creating tenants

tranchitella:
tenantadm CLI commands to create tenants now support flags to enable the addons; by default, all addons are disabled. Adapt the demo script and the tests to enable all the addons.

Changelog: None
Ticket: MEN-6292


kjaskiewiczz (approved):
looks good

PR #2325, opened 2023-02-08
test: device decommissioning is asynchronous, add retries

tranchitella:
Changelog: None
Ticket: None

### DIFF
--- a/backend-tests/tests/test_devauth.py
+++ b/backend-tests/tests/test_devauth.py
@@ -1,4 +1,4 @@
-# Copyright 2022 Northern.tech AS
+# Copyright 2023 Northern.tech AS
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
 #    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


tranchitella:
@mender-test-bot start pipeline --fast

backend-tests/tests/test_devauth.py
```diff
@@ -722,9 +722,16 @@ def do_test_delete_device_ok(self, devs_authsets, user, tenant_token=""):
         assert r.status_code == 401
 
         # verify the device is gone
-        r = devapim.with_auth(utoken).call(
-            "GET", deviceauth.URL_DEVICE, path_params={"id": dev_acc.id}
-        )
+        count = 5
```
merlin-northern, line 725: nit, feel free to ignore. I would rename to indicate that this is the retry count and means the number of seconds (since we have haardcoded sleep(1).

merlin-northern (approved):
usually when I do something like that I hear: we use a ptyhon mechanism to retry. but I will just not say that:) looks good.

backend-tests/tests/test_devauth.py
tranchitella: It's a copy-paster from another part of the very same test. Let's keep it consistent.

PR #2328, opened 2023-02-08
MEN-6294: fix(demo): Include `testing` composition with `--enterprise-testing`

lluiscampos:
We need this in order to correctly set the environment for `generate-delta-worker` container.

This is an internal and undocumented flag that we use for enterprise testing, so there should be no issue with adding all the testing settings into the composition.


lluiscampos:

PR #2330, opened 2023-02-09
Auto-generate release notes also from build tags + Miscellaneous chores

lluiscampos:


lluiscampos:
@kacf @oleorhagen The intention is to get the rid of this:

image

extra/release_tool.py
```diff
@@ -2849,7 +2849,19 @@ def do_release(release_state_file):
         elif reply.lower() == "l":
             do_license_generation(state, tag_avail)
         elif reply.lower() == "n":
-            do_generate_release_notes(state["repo_dir"], state["version"])
+            # Generate release notes for final tag if exists. Otherwise for the current build tag
+            tag_release_notes = state["version"]
+            try:
+                execute_git(
+                    state,
+                    repo.git(),
```
kacf, line 2857: Is `repo` valid here..?

extra/release_tool.py
```diff
@@ -2849,7 +2849,19 @@ def do_release(release_state_file):
...
+                    ["rev-parse", state["version"]],
+                    capture=True,
+                    capture_stderr=True,
+                )
```
kacf, line 2861: I think you don't need this, the information should already be available in `tag_avail`. I'm a bit rusty nowadays, but I think it should be `tag_avail["integration"]["build_tag"]`, which should always be the latest build tag, or the final tag, if it exists (check description of `check_tag_availability`).

extra/release_tool.py
```diff
@@ -2849,7 +2849,19 @@ def do_release(release_state_file):
         elif reply.lower() == "l":
             do_license_generation(state, tag_avail)
         elif reply.lower() == "n":
```
kacf, line 2851: Please also change the description of the `n` option.

extra/release_tool.py
lluiscampos: Thanks. I actually tried to find this exact path (that `tag_avail["integration"]["build_tag"]` will be updated with the final tag) but I couldn't see it. I'll look more into it :eyes:

oleorhagen (approved):
Nice!

Not too familiar with this code though.

extra/release_tool.py
lluiscampos: This comment is now obsolete.

lluiscampos:
* 3.5.x

lluiscampos:
* 3.3.x

PR #2332, opened 2023-02-17
QA-525: test: Fix and enable `TestClientCompatibility`

lluiscampos:
It is unknown at which point in time started to fail - the assumption is that an update in upstream `docker-compose` made this error reveal itself.

Basically the test was combining `up ... scale=0` with `run ...` which in effect would create two `mender-client` services. Although one will still be not running, the later logic to check that we have only one client would fail.

Moved also the TENANT_TOKEN to a shell variable because `up` command, unlike `run`, does not support `-e` options in the command line.

### DIFF
--- a/extra/integration-testing/test-compat/docker-compose.compat-2.0.yml
+++ b/extra/integration-testing/test-compat/docker-compose.compat-2.0.yml
@@ -8,7 +8,7 @@ services:
     tty: true
     privileged: true
     environment:
-      TENANT_TOKEN: ""
+      - TENANT_TOKEN=$TENANT_TOKEN
     volumes:
       # Provision demo certificate
       - ./cert/cert.crt:/mnt/config/server.crt
--- a/extra/integration-testing/test-compat/docker-compose.compat-2.1.yml
+++ b/extra/integration-testing/test-compat/docker-compose.compat-2.1.yml
@@ -8,7 +8,7 @@ services:
     tty: true
     privileged: true
     environment:
-      TENANT_TOKEN: ""
+      - TENANT_TOKEN=$TENANT_TOKEN
     volumes:
       # Provision demo certificate
       - ./cert/cert.crt:/mnt/config/server.crt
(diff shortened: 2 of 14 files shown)


lluiscampos:
@mender-test-bot start pipeline for real this time :hourglass_flowing_sand:

oleorhagen (approved):
LGTM

Where's `2.7` ? :thinking:

tranchitella (approved):

lluiscampos:
> LGTM
> 
> Where's `2.7` ? thinking

There is no mender _client_ 2.7. At the time of Mender 2.7, client was 2.6. Also note that at that time we had tags based on the mender _client_ version and not the Mender version (iow no `mender-2.7`)

lluiscampos:
My last force-push restores (although with a minor optimization) the `up --no-start` so that the containers are created before calling `ps --services`. This seems to behave differently in my machine that in CI.

lluiscampos:
@mender-test-bot start pipeline again :1234:

lluiscampos:
@mender-test-bot start pipeline :sunflower:

lluiscampos:
@mender-test-bot start pipeline :sake:

lluiscampos:
> (...). This seems to behave differently in my machine that in CI.

I figured out what the problem is and addressed it in the second commit:

```
 test: Fix listing of services for TestClientCompatibility

Since the CI workers have moved away from pip's `docker-compose` to the
actual Docker's `docker compose`, the `ps` command does not list the
services that are not running unless explicitly asked for it.

The previous fix was working by accident, and this might very well
explain why the test started failing suddenly - presumably after an
upstream change in `docker compose`
```

It is looking good now, merging!

oleorhagen:
> > LGTM
> > Where's `2.7` ? thinking
> 
> There is no mender _client_ 2.7. At the time of Mender 2.7, client was 2.6. Also note that at that time we had tags based on the mender _client_ version and not the Mender version (iow no `mender-2.7`)

Our naming is simply exquisite :laughing:

oleorhagen:
> My last force-push restores (although with a minor optimization) the `up --no-start` so that the containers are created before calling `ps --services`. This seems to behave differently in my machine that in CI.

Always a little worrying, as whichever is lagging behind is likely going to change once the version gets bumped :/

PR #2339, opened 2023-02-21
QA-348: ci: setup mender-docs tests

0lmi:
Changelog: None
Ticket: QA-348

### DIFF
--- a/.gitlab-ci-default-pipeline.yml
+++ b/.gitlab-ci-default-pipeline.yml
@@ -62,6 +62,7 @@ test:docs:
   before_script:
     - apt-get update && apt-get install -y
       docker.io bash git openssl pwgen python3 jq docker-compose wget sudo
+
     - git config --global user.name "user"
     - git config --global user.email "user@example.com"
     - git clone https://github.com/mendersoftware/mender-docs.git mender-docs
@@ -77,9 +78,11 @@ test:docs:
           echo $(git log --oneline $(git merge-base $i HEAD)..HEAD | wc -l) $i;
         done | sort -n | head -n1 | awk '{print $2}'
       )
-    - if [[ $MENDER_DOCS_BRANCH == "origin/staging" ]]; then
+    - if [[ "$MENDER_DOCS_BRANCH" == "origin/staging" ]]; then
     -   MENDER_DOCS_BRANCH="origin/hosted"
     - fi
+
+    - export INTEGRATION_BRANCH=$CI_COMMIT_REF_NAME
 
     - cd mender-docs
     - git checkout $MENDER_DOCS_BRANCH


0lmi:
@mender-test-bot start pipeline --pr integration/2339 --pr mender-docs/2026

0lmi:
It depends from https://github.com/mendersoftware/mender-docs/pull/2026

lluiscampos (approved):
This part looks good, but see the review in the related PR.

.gitlab-ci-default-pipeline.yml
```diff
@@ -59,26 +59,39 @@ test:docs:
     DOCKER_CERT_PATH: "/certs/client"
     DOCKER_TLS_VERIFY: "1"
     DOCKER_TLS_CERTDIR: "/certs"
+    # Testing of different integration<->mender-docs branches matrix can be done
+    # by setting MENDER_DOCS_BRANCH (in format 'origin/branch_name')
+    # and INTEGRATION_BRANCH (in format 'branch_name') variables from GitLab UI
+    # when run the pipeline manually
```
lluiscampos, line 65: I thought this was code you added for debugging purposes only. What is the value in leaving it?

0lmi: It allows to run the pipeline from master and test any combination of `integration`-`mender-docs` branches. 
I'm not really sure about all the details of our release process, but assume it might be useful there. Another theoretical use case might be feature branches.

lluiscampos: Okey.

I think it is a bad practice to leave lingering code "just in case". It is better to remove that to keep code that we don't run in any automated way and hence might break without noticing.

In any case, if you want to keep it, please move it to the global variables section and set a more global name, like `TEST_DOCS_MENDER_DOCS_BRANCH`

0lmi: I would say it's up to you to decide here, because I have very little knowledge about releases and needs for testing it. If it doesn't make sense then definitely it worth to remove it.

lluiscampos: > I would say it's up to you to decide here, (...)

Remove it, then :smiley:

.gitlab-ci-default-pipeline.yml
```diff
@@ -62,9 +62,10 @@ test:docs:
   before_script:
     - apt-get update && apt-get install -y
       docker.io bash git openssl pwgen python3 jq docker-compose wget sudo
+
     - git config --global user.name "user"
     - git config --global user.email "user@example.com"
-    - git clone https://github.com/mendersoftware/mender-docs.git mender-docs
+    - git clone https://gitlab.com/Northern.tech/Mender/mender-docs.git mender-docs
```
lluiscampos, line 68: This change is not necessary and adds a new entry point for error - if for any reason the GitLab fork is not up to date with GitHub, we might end up testing the wrong thing.

lluiscampos (changes requested):

.gitlab-ci-default-pipeline.yml
0lmi: Ok, now, after those discussions, I finally understood your comment in the related PR about "finding a way of testing it in CI"... It makes sense now. Thanks.

lluiscampos (approved):

PR #2349, opened 2023-03-03
3.3.x: feat(release_tool): Auto-generate release notes also from build tags

lluiscampos:
So that we can use this option early in the release process to sanitize the notes.


lluiscampos:
Cherry-pick of the relevant part of #2330.

PR #2371, opened 2023-06-26
feat: add docker-compose v2 support

Roman3349:
Adds support for newer docker compose versions (2.x).


kacf (approved):
Thanks, looks good to me! Let's see what the CI says.

PR #2384, opened 2023-07-11
chore: gw name change

merlin-northern:
Ticket: None

### DIFF
--- a/tests/tests/test_basic_integration.py
+++ b/tests/tests/test_basic_integration.py
@@ -47,7 +47,7 @@ class DeviceAuthFailover(DeviceAuthV2):
 
     def get_devauth_base_path(self):
         return "https://%s/api/management/v2/devauth/" % (
-            get_container_manager().get_ip_of_service("mender-api-gateway-2")[0]
+            get_container_manager().get_ip_of_service("mender-api-gateway-1")[0]
         )
 
 


oldgiova (approved):

alfrunes (changes requested):
@merlin-northern This `-2` prefix is by design (although the naming is very confusing). The container name is actually `mender-api-gateway-2-1`. The `get_ip_by_service` filters by the docker compose label of the service. You can modify the suffix of the [failover docker composition](https://github.com/mendersoftware/integration/blob/master/extra/failover-testing/docker-compose.failover-server.yml) to make this test more readable, but this change actually breaks the test.

merlin-northern:
> @merlin-northern This `-2` suffix is by design (although the naming is very confusing). The container name is actually `mender-api-gateway-2-1`. The `get_ip_by_service` filters by the docker compose label of the service. You can modify the suffix of the [failover docker composition](https://github.com/mendersoftware/integration/blob/master/extra/failover-testing/docker-compose.failover-server.yml) to make this test more readable, but this change actually breaks the test.

yes, I realized that this is not the correct fix, I commented on slack. I have a new lead now: https://github.com/mendersoftware/integration/pull/2385

merlin-northern:
> > @merlin-northern This `-2` suffix is by design (although the naming is very confusing). The container name is actually `mender-api-gateway-2-1`. The `get_ip_by_service` filters by the docker compose label of the service. You can modify the suffix of the [failover docker composition](https://github.com/mendersoftware/integration/blob/master/extra/failover-testing/docker-compose.failover-server.yml) to make this test more readable, but this change actually breaks the test.
> 
> yes, I realized that this is not the correct fix, I commented on slack. I have a new lead now: #2385

I am clueless again: https://github.com/mendersoftware/integration/pull/2385#issuecomment-1631091438

PR #2406, opened 2023-07-28
chore: made it easier to work with the docker client in tests

mzedel:
- this is primarily to allow usage of the demo artifact - even if it might not be fully functional, being able to trigger a deployment with it increases the possibilities this image provides

needed for tests in https://github.com/mendersoftware/gui/pull/3889


mzedel:

oleorhagen:
I don't fully get it. This is in the `docker-client` repo? :thinking:

mzedel:
> I don't fully get it. This is in the `docker-client` repo? 🤔

@oleorhagen 
- the `mender-client-docker-addons` image lives in this repo (the Dockerfile changed here)
- it is used in the gui e2e tests
- `docker-client` as device type is incompatible with the device types supported by the `mender-demo-artifact` (but that is uploaded automatically in staging, so I can't replace it with a rebuild that supports the device type)
- we don't build this on arm so `generic-x86` should be appropriate

oleorhagen (approved):

lluiscampos (approved):
Thank you. Long overdue fix.

PR #2433, opened 2023-09-20
Misc fixes for running tests with the cpp client

oleorhagen:


testutils/infra/device.py
```diff
@@ -153,7 +153,7 @@ def get_reboot_detector(self, host_ip):
     def get_client_service_name(self):
         if self._service_name is None:
             self._service_name = self.run(
-                "if test -e /lib/systemd/system/mender.service; then echo mender; else echo mender-client; fi"
+                "if test -e /lib/systemd/system/mender.service; then echo mender; elif test -e /lib/systemd/system/mender.service; then echo mender; else echo mender-updated; fi"
```
lluiscampos, line 156:
```suggestion
                "if test -e /lib/systemd/system/mender.service; then echo mender; elif test -e /lib/systemd/system/mender-client.service; then echo mender-client; else echo mender-updated; fi"
```

But also, what happens to `mender-authd`? For sure some of the tests that use this to restart a service expect both to be restarted :thinking:

lluiscampos: Can we get the rid of this helper? As far as I recall this was introduced for the transition between yocto releases (when we migrated from `mender` to `mender-client` recipe) but I think this time we will not have the same version of mender client (4.x) with different `systemd` service names, so it should be safe to remove.

@kacf ?

tests/tests/test_fault_tolerance.py
```diff
@@ -32,10 +32,6 @@
 from testutils.infra.container_manager.kubernetes_manager import isK8S
 
 DOWNLOAD_RETRY_TIMEOUT_TEST_SETS = [
-    {
-        "blockAfterStart": False,
-        "logMessageToLookFor": "[uU]pdate (check|fetch) failed:",
-    },
```
lluiscampos, line 38: Removed unintentionally?

tests/conftest.py
```diff
@@ -237,7 +237,8 @@ def run_remote_command(instances, command):
                 logger.info("Not able to print client deployment log")
 
             for service in [
-                "mender-client",
+                "mender-authd",
+                "mender-updated",
                 "mender-connect",
                 "mender-monitor",
                 "mender-gateway",
```
lluiscampos, line 244: This relates to my previous comment. If `integration` should be compatible with both versions of the recipe, then we should keep here `mender-client` -> So let's just clean-up `get_client_service_name` :broom:

tests/helpers.py
```diff
@@ -53,6 +53,8 @@ def ip_to_device_id_map(device_group, devauth=devauth):
         # Match them.
         identity_to_id = {}
         for dev in devauth_devices:
+            # Strip the array from the reported identity
+            dev["identity_data"]["mac"] = dev["identity_data"]['mac'][0]
```
lluiscampos, line 57: I am concerned about this (from the client point of view, not the test). Was it intentional? Then we should have a changelog entry. Was it not intentional? Then we should revert.

FYI @kacf

lluiscampos (changes requested):
Most of the things look good :soap:

tests/tests/test_fault_tolerance.py
oleorhagen: No :)

tests/helpers.py
oleorhagen: Yes, we should probably discuss this.

I guess it's not wrong in the sense that an array of one member is the same as one element, I guess we could ask @kjaskiewiczz for his opinion here.

Crux of the matter KJ is that we now send the identity data as an array of one member (by default).

lluiscampos: Yes, the backend will be happy with it but some custom scripts (like this exact test) will require adjustments by the user.

oleorhagen: What custom scripts? This is simply a difference in how we marshal and send it to the server. The identity script has not changed at all

testutils/infra/device.py
kacf: I think we cannot. Our service is called mender-updated now, and we decided not to have a mender-client service, so we need to adapt it to that.

kacf: No, wait, you're right. This only applies to v4.x, so then we don't need it. However, if there are similar constructs in meta-mender, then we do need those, because they will be used with multiple versions of the client.

tests/tests/test_update_control.py
kacf, line 1: :fire: :fire: :fire:

tests/module-state-scripts-test
```diff
@@ -18,7 +18,7 @@ case "$STATE" in
         ;;
 
     ArtifactReboot|ArtifactRollbackReboot)
-        systemctl restart mender-client || systemctl restart mender
+        systemctl restart mender-client || systemctl restart mender || systemctl restart mender-updated
```
kacf, line 21: As mentioned in other comments, you can remove all but the last one.

extra/mender-client-docker-addons/Dockerfile
```diff
@@ -57,7 +58,8 @@ RUN mkdir -p /mender-install/var/lib/mender && echo device_type=generic-x86_64 >
 
 FROM ubuntu:22.04
 
-RUN mkdir -p /run/dbus && apt-get update && apt-get install -y liblzma5 dbus openssh-server
+RUN mkdir -p /run/dbus && apt-get update && apt-get install -y \
+    liblzma5 dbus openssh-server liblmdb0 libarchive13 libboost-log1.74.0
```
kacf, line 62: Can we use just `libboost-log`? It will probably break on OS upgrade.

tests/tests/test_fault_tolerance.py
kacf: I think I know why, it's because the old fetch and store steps in the Golang client are done in the same state now, right? I think this deserves a "why" in the commit message though. I also found this confusing.

tests/helpers.py
kacf: If I understand correctly, we are simply using inventory data (where array is fine) to reconstruct identity data (where array is not fine). This will never be done by any user, so should be fine, right?

testutils/infra/device.py
lluiscampos: @kacf  Let me insist:

> I think this time we will not have the same version of mender client (4.x) with different systemd service names, so it should be safe to remove

The only scenario where to support multiple systemd service names from integration tests is when one specific mender client release branch (in this case 4.x) can have different systemdservice names based on the yocto branch we are building with.

But as I see the integration going, we either port mender 4.x recipes _and_ the systemd service names to any yocto branch we want to have support for it or none (iow we ~cannot~ will not port 4.x client recipes but keep the old systemd service names for it in an old yocto LTS branch)

Does it make sense?

tests/helpers.py
lluiscampos: I just don't get why this changed with the new client, sorry.

But you both seem pretty convinced that it is okey, so I won't insist further.

testutils/infra/device.py
kacf: Yes, exactly. Changing the service name is a requirement for any Yocto branch we port it to.

lluiscampos: Oh, I only now read that you agreed with me in the previous comment. Sorry.

Good :smiley:

lluiscampos: @oleorhagen You can ignore this for now and either me or you will do on a follow-up PR.

lluiscampos:
@oleorhagen Can we brush this up and review/merge today?

oleorhagen:
> @oleorhagen Can we brush this up and review/merge today?

Yaz :raised_hands:

extra/mender-client-docker-addons/Dockerfile
oleorhagen: `libboost-log` is not a package unfortunately :/

tests/tests/test_fault_tolerance.py
kacf: Should fix this, but otherwise looks good to me.

kacf (approved):

tests/tests/test_fault_tolerance.py
oleorhagen: Yes, should have been a separate commit, splitting it out.

lluiscampos (approved):

PR #2436, opened 2023-09-23
MEN-6544: Integrate `mender-setup` and `mender-snapshot` repositories

lluiscampos:


component-maps.yml
```diff
@@ -251,6 +251,30 @@ git:
     release_component: true
     independent_component: true
 
+  mender-setup:
+    docker_image:
+    - mender-client-qemu
+    - mender-client-qemu-rofs
```
kacf, line 257: I know that this isn't something we are going to bother fixing now, but I suppose this shouldn't actually be in the rofs images.

extra/mender-client-docker-addons/Dockerfile
```diff
@@ -8,13 +8,21 @@ RUN apt-get update && apt-get install -y make git build-essential golang liblzma
...
 RUN (cd /src/mender && git fetch origin $MENDER_CLIENT_REV && git checkout FETCH_HEAD || git checkout -f $MENDER_CLIENT_REV)
 
 RUN git clone https://github.com/mendersoftware/mender-connect /src/mender-connect
 RUN (cd /src/mender-connect && git fetch origin $MENDER_CONNECT_REV && git checkout FETCH_HEAD || git checkout -f $MENDER_CONNECT_REV)
 
+RUN git clone https://github.com/mendersoftware/mender-setup /src/mender-setup
+RUN (cd /src/mender-setup && git fetch origin $MENDER_SETUP_REV && git checkout FETCH_HEAD || git checkout -f $MENDER_SETUP_REV)
+
+RUN git clone https://github.com/mendersoftware/mender-snapshot /src/mender-snapshot
+RUN (cd /src/mender-snapshot && git fetch origin $MENDER_SNAPSHOT_REV && git checkout FETCH_HEAD || git checkout -f $MENDER_SNAPSHOT_REV)
```
kacf, line 24: You need to install it too, no?

Also I think you can omit mender-snapshot, it's unusable from this container.

kacf (approved):

oleorhagen (approved):

extra/mender-client-docker-addons/Dockerfile
lluiscampos: You are right. Removing

component-maps.yml
lluiscampos: This will depend on how it is integrated in `meta-mender`. I suspect that we will add it on demo and not exclude it depending on being read-only or not.

Additionally we could have a configuration where `/etc/mender/mender.conf` is a symlink to somewhere (we had a task for this) and then in theory you could use `mender-setup`.

Being an internal testing image, I will leave it as-is :)

PR #2437, opened 2023-09-27
test(full integration): fix the check if the devices reported inventory

kjaskiewiczz:
Changelog: Title
Ticket: None

### DIFF
--- a/tests/tests/test_compat.py
+++ b/tests/tests/test_compat.py
@@ -126,9 +126,8 @@ def accept_devices(api_deviceauth, devices=None):
 
 def assert_inventory_updated(api_inventory, num_devices, timeout=TIMEOUT):
     """
-    Polls the inventory every second, checking the returned "updated_ts"
-    properties for each device, waiting for the value to update to a value
-    later than when this function was called.
+    Polls the inventory every second, checking all devices reported some
+    attributes with inventory scope.
     :param api_inventory: testutils.api.client.ApiClient setup and authorized
                           to use the inventory management api,
                           i.e. api_client.with_auth(api_token).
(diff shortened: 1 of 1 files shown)


kjaskiewiczz:

kjaskiewiczz:

tests/tests/test_compat.py
```diff
@@ -153,18 +152,16 @@ def assert_inventory_updated(api_inventory, num_devices, timeout=TIMEOUT):
         if len(dev_invs) < num_devices:
             time.sleep(1)
             continue
-        # Check if inventories has been updated since starting this loop
+        # Check if inventory attributes has been reported
```
merlin-northern, line 155: nit: 
```suggestion
        # Check if inventory attributes have been reported
```

tests/tests/test_compat.py
```diff
@@ -153,18 +152,16 @@ def assert_inventory_updated(api_inventory, num_devices, timeout=TIMEOUT):
...
         for device in dev_invs:
-            # datetime does not have an RFC3339 parser, but we can convert
-            # the timestamp to a compatible ISO format by removing
-            # fractions and replacing the Zulu timezone with GMT.
-            updated_ts = datetime.fromisoformat(device["updated_ts"].split(".")[0])
-            updated_ts = updated_ts.replace(tzinfo=timezone.utc)
-            update_after = update_after.replace(tzinfo=timezone.utc)
-            if updated_ts > update_after:
-                num_updated += 1
-            else:
```
merlin-northern, line 167: I am a bit hesitant about removing the updated time check, as I consider it a vital part of the test. do you think you could run a command on a device and replace the inventory script, and trigger a real attributes update, and keep the check? if you wish not to do it, I would kindly request a followup QA ticket with just this scenario: 
* device keeps updating inventory, it does not change
* we edit the inventory script on a device
* inventory gets updated
* keeps reporting the same: no update.

kjaskiewiczz: this is compatibility test and if I understand it correctly, this part of the test is suppose to check if the client is able to report inventory attributes, and that's it;
I see no reason to modify inventory script on the device and look at updated_ts

merlin-northern: I will swallow this, as long as you create a follow up task :>

merlin-northern (approved):
all discussed

tests/tests/test_compat.py
kjaskiewiczz: follow up task - https://northerntech.atlassian.net/browse/QA-627

PR #2439, opened 2023-09-28
feat: enabled http3 in traefik & the demo setup

mzedel:
- this allows decreasing the time taken to init the gui on load
+ also made traefik rely on the static config from a mounted file

Ticket: None
Changelog: None


mzedel:

PR #2448, opened 2023-10-04
smtpd workaround

kjaskiewiczz:
In the python 3.12, smtpd package has been removed.
We need to migrate to aiosmtpd (QA-603), but for now,
as a workaround, we'll use python 3.11.6.


mzedel (approved):
the smtp fix is ok, I guess the changes to `test_compat` could help, so I'll be happy to give it a try 😄

kjaskiewiczz:

PR #2458, opened 2023-10-23
SEC-958: Separate test system and test Python dependencies

lluiscampos:
To workaround the behavior of `dependabot` that will assume all `.txt` files of the scanned folder to be Python dependencies.

This problem is preventing the bot from parsing the dependencies and submitting updates.


kacf (approved):

tests/README.md
```diff
@@ -33,12 +33,12 @@
 ## Getting Started
 
 The dependencies for the integration tests are collected and organized in
-dependency files in the `./requirements` folder, and separated into
+dependency files in the `./requirements-*` folders, and separated into
```
merlin-northern, line 36: is it only me or this sentence does not end well? separated into what?

merlin-northern (commented):
on nitpick, since Kristian reviewed and approved I am not doing a full review.

tests/README.md
lluiscampos: Separated into the table that is below.

Yeah, it is a bit strange but I'd rather let it be as it was :)

PR #2466, opened 2023-10-25
MEN-6671: Fixes for integration tests

lluiscampos:


tests/helpers.py
```diff
@@ -62,13 +62,13 @@ def ip_to_device_id_map(device_group, devauth=devauth):
         return ip_to_device_id
 
     @staticmethod
-    def _check_log_for_message(device, service, message, since=None):
+    def _check_log_for_message(device, message, since=None):
```
oleorhagen, line 65: S: Use a default argument for the service instead?

oleorhagen (approved):
Nice

tests/helpers.py
```diff
@@ -80,21 +80,21 @@ def _check_log_for_message(device, service, message, since=None):
...
         assert (
             sleepsec <= timeout
-        ), f"timeout for waiting for message '{message}' in {service} log"
+        ), f"timeout for waiting for message '{message}' in mender-authd log"
 
     @staticmethod
     def check_log_is_authenticated(device, since=None):
         Helpers._check_log_for_message(
-            device, "mender-updated", "No update available", since
+            device, "Successfully received new authorization data", since
```
kacf, line 93: Lol, several messages were posted, then deleted again, as I followed this evolution in the commits! 😂

extra/mender-client-docker-addons/Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN (cd /src/mender-configure-module && \
 RUN mkdir --parents /mender-install/etc/mender
 
 WORKDIR /src/mender
-RUN test -f CMakeLists.txt && cmake -D MENDER_NO_BUILD=1 -S .
+RUN if [ -f CMakeLists.txt ]; then cmake -D MENDER_NO_BUILD=1 -S .; fi
```
kacf, line 33: I thought failing was the intention here. Why do we want it to succeed if `CMakeLists.txt` is missing?

lluiscampos: > I thought failing was the intention here. 

Why the `test`, then?

> Why do we want it to succeed if `CMakeLists.txt` is missing?

Right now it fails because it builds from `master` branch from mender. You have a point in that `integration` and `mender` repos are both versioned together and we could _assume_ there is a `CMakeLists.txt`.

For now I'll keep the `if` to have green pipeline and we can clean it up later on after we get the rid of the feature branches. It is harmless as is.

tests/helpers.py
lluiscampos: Yeah, sorry for that. There was a bit of back and forth as I found issues in the tests.

tests/helpers.py
lluiscampos: I could have, indeed.

But after we moved the log message to `mender-auth`, and having both methods really about authentication, I thought it would make sense to hardcode `mender-authd`. Only if we extend it to grep for other general purpose messages in `mender-update` would make sense to parametrize again.

extra/mender-client-docker-addons/Dockerfile
kacf: Ref `test`: Of course, I was thinking of this totally backwards. You're right.

kacf (approved):

PR #2473, opened 2023-11-01
feat: Upgrade to Mender version `3.6.3`.

tranchitella:
Changelog: None
Ticket: None

### DIFF
--- a/git-versions.yml
+++ b/git-versions.yml
@@ -50,7 +50,7 @@ services:
         image: mendersoftware/mender:3.5.1
 
     mender-connect:
-        image: mendersoftware/mender-connect:2.1.0
+        image: mendersoftware/mender-connect:2.1.1
 
     mender-setup:
         image: mendersoftware/mender-setup:master
--- a/other-components.yml
+++ b/other-components.yml
@@ -5,16 +5,16 @@
 # 'services' tag.
 services:
     mender-artifact:
-        image: mendersoftware/mender-artifact:3.10.1
+        image: mendersoftware/mender-artifact:3.10.2
 
     mender-cli:
-        image: mendersoftware/mender-cli:1.11.0
+        image: mendersoftware/mender-cli:1.11.1
 
     mender-binary-delta:
         image: mendersoftware/mender-binary-delta:1.4.1
 
     mender-convert:
-        image: mendersoftware/mender-convert:4.0.2
+        image: mendersoftware/mender-convert:4.0.3
 
     meta-mender:
         image: mendersoftware/meta-mender:kirkstone


other-components.yml
```diff
@@ -5,16 +5,16 @@
 # 'services' tag.
 services:
     mender-artifact:
-        image: mendersoftware/mender-artifact:3.10.1
+        image: mendersoftware/mender-artifact:3.10.2
 
     mender-cli:
-        image: mendersoftware/mender-cli:1.11.0
+        image: mendersoftware/mender-cli:1.11.1
```
merlin-northern, line 11: I cant pull this image:
```bash
# docker pull mendersoftware/mender-cli:1.11.1
Error response from daemon: pull access denied for mendersoftware/mender-cli, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
should it be like that?

other-components.yml
```diff
@@ -5,16 +5,16 @@
 # 'services' tag.
 services:
     mender-artifact:
-        image: mendersoftware/mender-artifact:3.10.1
+        image: mendersoftware/mender-artifact:3.10.2
```
merlin-northern, line 8: I cannot pull this image:
```bash
# docker pull mendersoftware/mender-artifact:3.10.2
Error response from daemon: pull access denied for mendersoftware/mender-artifact, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
should it be like that?

git-versions.yml
```diff
@@ -50,7 +50,7 @@ services:
         image: mendersoftware/mender:3.5.1
 
     mender-connect:
-        image: mendersoftware/mender-connect:2.1.0
+        image: mendersoftware/mender-connect:2.1.1
```
merlin-northern, line 53: I cannot pull this image:
```bash
# docker pull mendersoftware/mender-connect:2.1.1
Error response from daemon: pull access denied for mendersoftware/mender-connect, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
should it be like that?

merlin-northern (commented):
as a review I checked if I can pull the images. I cant pull all of them, I wonder should it be like that? since I cant pull the old tags, I believe that we are just not pushing those components. just reporting.

other-components.yml
tranchitella: These are not Docker images (I know, I know).
It is just a way to express verisons.

other-components.yml
tranchitella: These are not Docker images (I know, I know).
It is just a way to express versions.

git-versions.yml
tranchitella: These are not Docker images (I know, I know).
It is just a way to express versions.

merlin-northern (approved):
thanks for answers

tranchitella:
I won't run the QA pipeline as the change doesn't affect it.

PR #2474, opened 2023-11-02
test: Top timeouts to 10 minutes

lluiscampos:
The one hours timeouts are ancient, they come from the Jenkins time. With such long timeouts, it is impossible to run the full suite when there are bugs in the software (the CI jobs will timeout after 8h and no results will be gathered).

My reasoning is that if a devices takes more than 10 minutes to respond to something, we have bugs in the software or we have such an overloaded CI that we might as well just abort.

### DIFF
--- a/testutils/infra/device.py
+++ b/testutils/infra/device.py
@@ -106,7 +106,7 @@ class MenderDevice:
 
         _put(self, file, local_path, remote_path)
 
-    def ssh_is_opened(self, wait=60 * 60):
+    def ssh_is_opened(self, wait=10 * 60):
         """Block until SSH connection is established on the device
 
         Keyword arguments:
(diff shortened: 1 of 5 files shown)


oleorhagen (approved):
Should we take the opportunity to unify this somewhere, like in an option for instance?

testutils/infra/device.py
```diff
@@ -361,7 +361,7 @@ def _put(device, file, local_path=".", remote_path="."):
 
 # Roughly the execution time of the slowest test (*) times 3
 # (*) As per 2020-03-24 test_image_download_retry_hosts_broken takes 515.13 seconds
```
lluiscampos, line 363: This comment I need to remove or update

kacf (approved):
More so than Jenkins, I think it was from the DigitalOcean time (where the performance was abysmal), as well as when we were using HDD storage instead of SDD storage.

Unification sounds good, but it's possible some tests have a need for longer timeouts than others, so it should at least be overridable. And nothing that should block this PR.

lluiscampos:
> Should we take the opportunity to unify this somewhere, like in an option for instance?

We could. We are also using already the `pytest-timeout` plugin and we could use it to set a default timeout per test and then manually set higher timeouts for tests that are known to be slow.

That would be a different approach. This PR is limiting the time for individual commands or other operations, but a test can still take longer, while with the plugin we would limit the whole test.

PR #2476, opened 2023-11-02
MEN-6671: Couple of integration tests fixes

lluiscampos:


tests/tests/test_mtls.py
```diff
@@ -133,15 +133,13 @@ def hsm_get_key_uri(self, pin, ssl_engine_id, device):
             + '\nMODULE_PATH = /usr/lib/softhsm/libsofthsm2.so\ninit = 0\n" >> /etc/ssl/openssl.cnf'
         )
         device.run(
-            'sed -i.backup -e "/\\[Service\\]/ a Environment=SOFTHSM2_CONF=/softhsm/softhsm2.conf" /lib/systemd/system/%s.service'
-            % device.get_client_service_name()
+            'sed -i.backup -e "/\\[Service\\]/ a Environment=SOFTHSM2_CONF=/softhsm/softhsm2.conf" /lib/systemd/system/mender-authd.service'
```
oleorhagen, line 136: This now, as far as I can tell, needs to be done for both services :thinking:

tests/tests/test_preauth.py
```diff
@@ -239,7 +237,7 @@ def substitute_id_data(device, id_data_dict):
     def restart(device):
         """Restart the mender service."""
 
-        device.run("systemctl restart %s.service" % device.get_client_service_name())
+        device.run("systemctl restart mender-updated")
```
oleorhagen, line 240: We shuuure this should not be `auth`?

oleorhagen (approved):
Real nice - thx!

:medal_military:

tests/tests/test_preauth.py
lluiscampos: This test actually passes, so it needs to be right.

However I share with you the hunch that it _should_ be `mender-authd`. Which brings me to the realization that we actually don't need to restart the service at all: `mender-auth` will pick up the updated id script once it is modified on disk. I'll remove the restart for clarity.

tests/tests/test_mtls.py
lluiscampos: I'll do it for good measure. This test is not passing yet.

tests/tests/test_mtls.py
```diff
@@ -183,8 +181,7 @@ def common_test_mtls_enterprise(self, env, algorithm=None, use_hsm=False):
                 remote_path="/var/lib/mender",
             )
 
-        client_service_name = env.device.get_client_service_name()
-        env.device.run("systemctl stop %s" % client_service_name)
+        env.device.run("systemctl stop mender-updated")
```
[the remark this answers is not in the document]
lluiscampos: Related to the comment above, I'll do both.

lluiscampos:
@oleorhagen I would appreciate your eyes again as I had to fix few things in the rework :+1:

lluiscampos:
This PR resulted in more failures than our nightly reference. So refrain further review from now.

I submitted the fix for test_configuration separately (this test actually passed here :tada: ). See:
* https://github.com/mendersoftware/integration/pull/2477

lluiscampos:
Superseded by #2503

PR #2481, opened 2023-11-03
chore: Update one missing `mender-update` CLI call

lluiscampos:


lluiscampos:
The test does not pass (yet), but I want to get this one merged so that I can delegate the debugging of the test :+1:

oleorhagen (approved):

PR #2486, opened 2023-11-07
Disable reporting

tranchitella:


tranchitella:

lluiscampos (approved):
Thank you for doing this.

I think it would have been better to skip the test than to remove it. But as long as we restore the test once we get reporting into production we should be fine.

kjaskiewiczz (approved):

PR #2488, opened 2023-11-09
chore: re-enable the DB migration tests

oleorhagen:
Ticket: MEN-6837


tests/tests/test_db_migration.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Northern.tech AS
+# Copyright 2023 Northern.tech AS
```
lluiscampos, line 1: By the way, these updates are not needed anymore :+1:

PR #2489, opened 2023-11-09
MEN-6843: Modify test to accommodate behavior change

lluiscampos:
Since the removal of the update control maps, the new client won't check for the deployment status _after_ the download and _before_ reboot, so this test needs to expect a reboot and rollback before failing the deployment.


lluiscampos:

oleorhagen (approved):

kacf (approved):

PR #2493, opened 2023-11-14
chore: Enable the mtls integration test

oleorhagen:

### DIFF
--- a/tests/tests/test_mtls.py
+++ b/tests/tests/test_mtls.py
@@ -252,7 +252,7 @@ pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin {pin} --write
         out = setup_ent_mtls.device.run("mender-update show-artifact").strip()
         assert out == "mtls-artifact"
 
-
+    @pytest.mark.skip(reason="QA-587")
     @MenderTesting.fast
     @pytest.mark.parametrize("algorithm", ["rsa"])
     def test_mtls_enterprise_hsm(self, setup_ent_mtls, algorithm):
--- a/tests/tests/test_mtls.py
+++ b/tests/tests/test_mtls.py
@@ -88,9 +88,6 @@ exit 0
 @pytest.mark.skipif(
     isK8S(), reason="not relevant in a staging or production environment"
 )
-@pytest.mark.skipif(
-    not (os.environ.get("NIGHTLY_BUILD", "false") == "true"), reason="MEN-6671",
-)
 class TestClientMTLSEnterprise:
     wait_for_device_timeout_seconds = 64
 
@@ -100,11 +97,13 @@ class TestClientMTLSEnterprise:
         script = f"""\
 #!/bin/bash
 echo "module: /usr/lib/softhsm/libsofthsm2.so" > /usr/share/p11-kit/modules/softhsm2.module
-mkdir -p /softhsm/tokens
-echo "directories.tokendir = /softhsm/tokens" > /softhsm/softhsm2.conf
-export SOFTHSM2_CONF=/softhsm/softhsm2.conf;
-softhsm2-util --init-token --free --label unittoken1 --pin {pin} --so-pin 0002
-pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin {pin} --write-object "{key}" --type privkey --id 0909 --label privatekey
+softhsm2-util --init-token --free --label testtoken --pin {pin} --so-pin 0002
+pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin {pin} --write-object "{key}" --type privkey --id 0909 --label certificatekey
+
+# Add the public key to the slot also
+# (Needed for the pkcs11-provider - since it does not extract it when needed)
+openssl pkey -in {key} -pubout -out {key}.pub
+pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin {pin} --write-object "{key}.pub" --type pubkey --id 0909 --label certificatekey
 """
         tmpdir = tempfile.mkdtemp()
         initialize_hsm_script = os.path.join(tmpdir, "init-hsm.sh")
@@ -123,32 +122,12 @@ pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin {pin} --write
 
     def hsm_get_key_uri(self, pin, ssl_engine_id, device):
         pt11tool_output = device.run(
-            "export SOFTHSM2_CONF=/softhsm/softhsm2.conf; p11tool --login --provider=/usr/lib/softhsm/libsofthsm2.so --set-pin="
-            + pin
-            + " --list-all-privkeys"
+            f"p11tool --login --provider=/usr/lib/softhsm/libsofthsm2.so --set-pin={pin} --list-all-privkeys"
         ).rstrip("\n")
         key_uri = re.search(r"URL:\s(.*)", pt11tool_output).group(1)
         key_uri = key_uri + ";pin-value=" + pin
 
-        device.run("cp /etc/ssl/openssl.cnf /etc/ssl/openssl.cnf.backup")
-        device.run(
-            'echo -ne "[openssl_init]\nengines=engine_section\n\n[engine_section]\npkcs11 = pkcs11_section\n\n[pkcs11_section]\nengine_id = '
-            + ssl_engine_id
-            + '\nMODULE_PATH = /usr/lib/softhsm/libsofthsm2.so\ninit = 0\n" >> /etc/ssl/openssl.cnf'
-        )
-        device.run(
-            'sed -i.backup -e "/\\[Service\\]/ a Environment=SOFTHSM2_CONF=/softhsm/softhsm2.conf" /lib/systemd/system/%s.service'
-            % device.get_client_service_name()
-        )
         return key_uri
-
-    def hsm_cleanup(self, device):
-        device.run(
-            "mv /lib/systemd/system/%s.service.backup /lib/systemd/system/%s.service || true"
-            % (device.get_client_service_name(), device.get_client_service_name())
-        )
-        device.run("rm -Rf /softhsm")
-        device.run("mv /etc/ssl/openssl.cnf.backup /etc/ssl/openssl.cnf || true")
 
     def common_test_mtls_enterprise(self, env, algorithm=None, use_hsm=False):
         # upload the certificates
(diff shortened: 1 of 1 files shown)


oleorhagen:
Goes along with: 
* https://github.com/mendersoftware/mtls-ambassador/pull/158
* https://github.com/mendersoftware/deviceauth-enterprise/pull/217

tests/tests/test_mtls.py
```diff
@@ -123,32 +122,17 @@ def hsm_setup(self, pin, ssl_engine_id, device):
 
     def hsm_get_key_uri(self, pin, ssl_engine_id, device):
         pt11tool_output = device.run(
-            "export SOFTHSM2_CONF=/softhsm/softhsm2.conf; p11tool --login --provider=/usr/lib/softhsm/libsofthsm2.so --set-pin="
+            "p11tool --login --provider=/usr/lib/softhsm/libsofthsm2.so --set-pin="
             + pin
             + " --list-all-privkeys"
```
lluiscampos, line 127: Once we are here, what about modernizing the formatting?
```suggestion
            f"p11tool --login --provider=/usr/lib/softhsm/libsofthsm2.so --set-pin={pin} --list-all-privkeys"
```

tests/tests/test_mtls.py
```diff
@@ -123,32 +122,17 @@ def hsm_setup(self, pin, ssl_engine_id, device):
...
         return key_uri
 
     def hsm_cleanup(self, device):
-        device.run(
-            "mv /lib/systemd/system/%s.service.backup /lib/systemd/system/%s.service || true"
-            % (device.get_client_service_name(), device.get_client_service_name())
-        )
-        device.run("rm -Rf /softhsm")
-        device.run("mv /etc/ssl/openssl.cnf.backup /etc/ssl/openssl.cnf || true")
+        device.run("rm -rf /var/lib/softhsm/tokens/*")
```
lluiscampos, line 135: I don't think we need to cleanup anything, really. We don't reuse the device for further tests.

tests/tests/test_mtls.py
```diff
@@ -212,12 +181,19 @@ def common_test_mtls_enterprise(self, env, algorithm=None, use_hsm=False):
                         "Certificate": f"/var/lib/mender/client.1.{algorithm}.crt",
                         "Key": key_uri,
                     }
-                    logger.info('client key set to "%s"' % key_uri)
+                    config["Security"] = {
+                        "SSLEngine": ssl_engine_id,
+                        "AuthPrivateKey": key_uri,
+                    }
+                    logger.info('set private key to "%s"' % key_uri)
```
lluiscampos, line 188:
```suggestion
                    logger.info(f"set private key to {key_uri}")
```

lluiscampos (approved):
Amazing that you got this one under control :rocket:

tests/tests/test_mtls.py
oleorhagen: Hmm, yeh. I did this first time around. Agreed. Let's just keep it

tests/tests/test_mtls.py
```diff
@@ -364,9 +326,7 @@ def test_mtls_enterprise_without_client_cert(self, setup_ent_mtls):
             "pending", max_wait=device_not_present_timeout_seconds * 0.5, no_assert=True
         ):
             devauth.decommission(device["id"])
-        setup_ent_mtls.device.run(
-            "systemctl start %s" % setup_ent_mtls.device.get_client_service_name()
-        )
+        setup_ent_mtls.device.run("systemctl start mender-authd")
```
merlin-northern, line 329: if we have `get_client_service_name` should we have also something for `mender-authd`?

tests/tests/test_mtls.py
```diff
@@ -162,19 +141,6 @@ def common_test_mtls_enterprise(self, env, algorithm=None, use_hsm=False):
...
-            remote_path="/etc/mender",
-        )
-        env.device.put(
-            "cert.crt",
-            local_path=os.path.join(basedir, "cert"),
-            remote_path="/etc/mender",
-        )
-        env.device.run(
-            "cat /etc/mender/server.crt /etc/mender/cert.crt > /tmp/server.crt && mv /tmp/server.crt /etc/mender/server.crt && rm /etc/mender/cert.crt"
-        )
```
merlin-northern, line 177: why is that not needed anymore?

tests/tests/test_mtls.py
```diff
@@ -123,33 +122,13 @@ def hsm_setup(self, pin, ssl_engine_id, device):
...
-        device.run("cp /etc/ssl/openssl.cnf /etc/ssl/openssl.cnf.backup")
-        device.run(
-            'echo -ne "[openssl_init]\nengines=engine_section\n\n[engine_section]\npkcs11 = pkcs11_section\n\n[pkcs11_section]\nengine_id = '
-            + ssl_engine_id
-            + '\nMODULE_PATH = /usr/lib/softhsm/libsofthsm2.so\ninit = 0\n" >> /etc/ssl/openssl.cnf'
-        )
-        device.run(
-            'sed -i.backup -e "/\\[Service\\]/ a Environment=SOFTHSM2_CONF=/softhsm/softhsm2.conf" /lib/systemd/system/%s.service'
-            % device.get_client_service_name()
-        )
```
merlin-northern, line 142: ok, I do not understand this: if you dont configure the engine, then from what do you load the key?

tests/tests/test_mtls.py
```diff
@@ -203,7 +168,6 @@ def common_test_mtls_enterprise(self, env, algorithm=None, use_hsm=False):
             config = json.loads(output)
             # replace mender.conf with an mTLS enabled one
             config["ServerURL"] = "https://mtls-ambassador:8080"
-            config["ServerCertificate"] = "/etc/mender/server.crt"
```
merlin-northern, line 206: I am having doubts here, isnt it the idea of this test to have the server certificate here? (it has been a long time, and I have not wrote this test in whole, so sorry for stupid quesitons)

tests/tests/test_mtls.py
```diff
@@ -237,7 +208,7 @@ def common_test_mtls_enterprise(self, env, algorithm=None, use_hsm=False):
         # start the Mender client
         logger.info("starting the client.")
         env.device.run("systemctl daemon-reload")
-        env.device.run("systemctl start %s" % client_service_name)
+        env.device.run("systemctl start mender-authd")
```
merlin-northern, line 211: same as comment below, nit, but maybe it would be nice to have something like `env.device.get_client_service_name` but for mender-authd?

tests/tests/test_mtls.py
```diff
@@ -100,11 +97,13 @@ def hsm_setup(self, pin, ssl_engine_id, device):
...
-export SOFTHSM2_CONF=/softhsm/softhsm2.conf;
-softhsm2-util --init-token --free --label unittoken1 --pin {pin} --so-pin 0002
-pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin {pin} --write-object "{key}" --type privkey --id 0909 --label privatekey
+softhsm2-util --init-token --free --label testtoken --pin {pin} --so-pin 0002
+pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin {pin} --write-object "{key}" --type privkey --id 0909 --label certificatekey
+
+# Add the public key to the slot also
+# (Needed for the pkcs11-provider - since it does not extract it when needed)
+openssl pkey -in {key} -pubout -out {key}.pub
+pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin {pin} --write-object "{key}.pub" --type pubkey --id 0909 --label certificatekey
```
merlin-northern, line 106: we discussed that over slack. all I can add is: the fact that we need to change the test in such fundamental manner means that the code behaves differently. I find it disturbing. what if your secure element does not let you store the public key?

tests/tests/test_mtls.py
oleorhagen: We had `get_client_service_name` because at the time, it could be two different ones. Not anymore :)

tests/tests/test_mtls.py
oleorhagen: This is exactly why I wanted your eyes on it :)

Why was it needed in the first place?

tests/tests/test_mtls.py
oleorhagen: We add the certificate to `ca-certs` now :)

tests/tests/test_mtls.py
oleorhagen: Then it won't work with the `v3 provider` we are using atm

tests/tests/test_mtls.py
merlin-northern: that cannot be. the engine id is arbitrary and there are many ways to give it in the config, how can it be in the code?

tests/tests/test_mtls.py
oleorhagen: This is not a limitation in _our_ code

tests/tests/test_mtls.py
oleorhagen: Yes, but there is a default directory, in which it loads modules from

oleorhagen: You can override this with the config if you so desire of course

oleorhagen: Also remember, that we are not testing `Engines` now, but `Providers`

oleorhagen: I used this as my guide:

https://github.com/openssl/openssl/blob/master/README-PROVIDERS.md#loading-providers

And went for the explicit approach. Also please remember that this is not an `Engine` test. They don't exist for us in OpenSSLv3 which is what we are testing here.

oleorhagen: I asked a question on Mender-hub also, to see what our users think about this :)

https://hub.mender.io/t/pkcs11-hsm-support-for-authentication-authorization-and-signature-verification/886/18

merlin-northern: I was not aware that we are deprecating ENgines in the client in this way.

tests/tests/test_mtls.py
merlin-northern: this is highly misleading. please change the description of the commit. it is a different test, testing different code, using different API.

merlin-northern:
@oleorhagen I will not block this one, we discussed privately, and also you gave me answers here. you have the approval of Lluis, and with the certificate I would ask Fabio, just to be on the safe side. (he was the  author of the original core test)

tests/tests/test_mtls.py
oleorhagen: Well, it's OpenSSL deprecating it. Not us :smile: 

We're just not enabling deprecated API's atm.

merlin-northern (approved):
please change the commit message, explaining exactly why this test is significantly modified. I am approving as you explained to me that the source of all my doubts lies in the decisions you had made.

tests/tests/test_mtls.py
oleorhagen: Alright, did a little bit of digging. The storage-proxy cert is already a part of the demo-cert now.

PR #2524, opened 2023-12-11
fix(release-tool): do not break if there are no commits from the previous version

tranchitella:
Changelog: None
Ticket: None


kacf:
There is something weird here. The `determine_version_bump` function should not be called if the component is at the previous version. If there are no commits at all, are you sure you input the previous version, and didn't try to make a new version out of nothing?

tranchitella:
Doesn't matter, let's skip this.

PR #2528, opened 2023-12-15
test: Do not override `mender-iot-manager` image for backend-tests

lluiscampos:
I believe that the only intention of having the service in the testing docker-compose file is to set the environment variable.

Specifying here the `image` is wrong, as it will override what we have in the root compose files and hence not run the tests with the freshly build image of `iot-manager` from the PR.

### DIFF
--- a/backend-tests/docker/docker-compose.backend-tests.yml
+++ b/backend-tests/docker/docker-compose.backend-tests.yml
@@ -49,7 +49,6 @@ services:
             REPORTING_REV: ${REPORTING_REV}
 
     mender-iot-manager:
-        image: mendersoftware/iot-manager:mender-master
         environment:
           IOT_MANAGER_DOMAIN_WHITELIST: "*.azure-devices.net *.iot.*.amazonaws.com mender-backend-tests-runner"
 


lluiscampos:
Build @ [Pipeline 1108406348](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1108406348)

kjaskiewiczz (approved):
I agree with this change.
There can be some consequences for the backend test on staging, since originally image was placed here to solve some issue with the tests on staging - https://github.com/mendersoftware/integration/commit/9e88c8bd442c59833b38ebf5a01ac9fb584d26e3, but the tests on staging are broken anyway, and if this was solving some issue, it should be solved differently.

merlin-northern (approved):
I was trying to find some reason for it, but I found none. maybe it was added here, because it was not present (yet) for some reason in the main integration composition files.

lluiscampos:
> I agree with this change. There can be some consequences for the backend test on staging, since originally image was placed here to solve some issue with the tests on staging - [9e88c8b](https://github.com/mendersoftware/integration/commit/9e88c8bd442c59833b38ebf5a01ac9fb584d26e3), but the tests on staging are broken anyway, and if this was solving some issue, it should be solved differently.

Discussed over Slack. We merge this fix and whenever fixing the staging test we can think about a different workaround or restore this one with some CI scripts changes to modify the image in place for PRs.

PR #2531, opened 2023-12-19
fix: Update `mender-client-docker-addons` image for C++ client

lluiscampos:
There are three fixes:
* Actually build the binaries (removing `MENDER_NO_BUILD`). We need the binaries in the final image, we don't know what the author of 394d99ab was thinking
* Set install prefix to `/usr` when configuring with CMake. Otherwise it would default to `/usr/local`. See: https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html
* Call the new binaries from the `entrypoint.sh` script.

And some optimizations:
* Remove conditional check for `CMakeLists.txt` and assume always new client (as we have done for the tests, for example). Note that `integration` repository is versioned so it will always follow _one_ version of the client.
* Run as many jobs in parallel with `make`.

Changelog: None
Ticket: None

### DIFF
--- a/extra/mender-client-docker-addons/Dockerfile
+++ b/extra/mender-client-docker-addons/Dockerfile
@@ -30,8 +30,8 @@ RUN (cd /src/mender-configure-module && \
 RUN mkdir --parents /mender-install/etc/mender
 
 WORKDIR /src/mender
-RUN if [ -f CMakeLists.txt ]; then cmake -D MENDER_NO_BUILD=1 -S .; fi
-RUN DESTDIR=/mender-install make prefix=/usr install
+RUN cmake -D CMAKE_INSTALL_PREFIX:PATH=/usr -S .
+RUN DESTDIR=/mender-install make --jobs=$(nproc --all) install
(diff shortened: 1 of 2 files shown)


extra/mender-client-docker-addons/Dockerfile
```diff
@@ -30,8 +30,8 @@ RUN (cd /src/mender-configure-module && \
 RUN mkdir --parents /mender-install/etc/mender
 
 WORKDIR /src/mender
-RUN if [ -f CMakeLists.txt ]; then cmake -D MENDER_NO_BUILD=1 -S .; fi
-RUN DESTDIR=/mender-install make prefix=/usr install
+RUN cmake -D CMAKE_INSTALL_PREFIX:PATH=/usr -S .
```
oleorhagen, line 33: Q: What's the extra `PATH` suffix?

oleorhagen (approved):

extra/mender-client-docker-addons/Dockerfile
lluiscampos: It is the "type" of the variable, see [here](https://cmake.org/cmake/help/latest/prop_cache/TYPE.html
).. I guess it would work anyway as an string, but 1) this is what I found in stackoverflow and 2) better to be explicit than sorry

PR #2558, opened 2024-01-26
MEN-6888: Re-enable (and modify) `test_in_poor_network_environment`

lluiscampos:
The original test design assumed that `mender-connect` could reuse the
    same TCP connection after the network is restored. However, the
    `http-forwarder` in the Mender client (`mender` or `mender-auth`) do
    close the TCP connection when we mess it up with `iptables`.
    
    Modify the test to expect `mender-connect` to be able to open a new
    shell after the connection is restored, which is evidence that it can
    "auto-heal" after a network disruption.
    
    `mender-connect` 2.2.0 and older do not pass this test due to MEN-6888.
    It is fixed now at:
    * https://github.com/mendersoftware/mender-connect/pull/130

### DIFF
--- a/tests/tests/test_mender_connect.py
+++ b/tests/tests/test_mender_connect.py
@@ -191,7 +191,6 @@ class _TestRemoteTerminalBase:
             assert prot.protoType == proto_shell.PROTO_TYPE_SHELL
             assert prot.typ == "bogusmessage"
 
-    @pytest.mark.xfail(reason="CE-277")
     def test_in_poor_network_environment(self, docker_env):
         self.assert_env(docker_env)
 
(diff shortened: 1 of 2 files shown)


tests/tests/test_mender_connect.py
```diff
@@ -197,7 +197,7 @@ def test_in_poor_network_environment(self, docker_env):
 
         receive_timeout_s = 16
 
-        def is_shell_is_working(shell):
+        def is_shell_working(shell):
```
oleorhagen, line 199: haha, thx!

oleorhagen (approved):
Nice!

lluiscampos:
* 3.7.x
* 3.6.x

PR #2571, opened 2024-02-23
test: Update debug query

lluiscampos:
This test is only interested in `mender-update`. Doing a grep on `mender` returns three PID (update, auth and connect) and the commands won't really work as expected.


oleorhagen (approved):

kacf (approved):

lluiscampos:

PR #2577, opened 2024-03-18
test: Stop running with capture=no so the test runner captures it.

kacf:
Then we can see the output localized to each test instead, both in the log and in HTML output.

This is only possible because Fabric has been removed.


lluiscampos (approved):
Beautiful :star_struck:

danielskinstad (approved):

kacf:
Being tested with failures in #2578.

kacf:
Works! With some caveats which I'll post in Slack.

PR #2611, opened 2024-05-31
Delta test fixes

kacf:


merlin-northern (approved):
looks good

tests/tests/test_autogenerate_delta.py
```diff
@@ -114,7 +114,7 @@ def deployment_triggered_callback():
             i = 60
             while i > 0:
                 output = mender_device.run(
-                    "systemctl status --no-pager --full --lines 1000000 mender-updated | grep 'xdelta3'",
+                    "grep 'xdelta3' /data/mender/deployments.*.log",
```
merlin-northern, line 117:
```suggestion
                    "grep 'xdelta3' /data/mender/deployments.001*.log",
```
I just wonder should we check the last one only?

merlin-northern: do not treat this as a blocker, I am ok with the original fix.

kacf: If we only check the last one, then we run into the same issue as before: After the reboot, it will not contain the string we're looking for, only the previous one will. So which log contains the string depends on whether it happens to run before or after the reboot.

lluiscampos (approved):

PR #2614, opened 2024-06-03
3.7.x test fixes

kacf:


lluiscampos (approved):

kacf:
One spurious error [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1316331205) which is definitely not related to this PR. It also passed in an earlier run, so ignoring it.

PR #2622, opened 2024-06-29
MEN-7302: Make large image larger

lluiscampos:
For scarthgap we have increased the rootfs size to around 400M. This test image then needs to be even larger to guarantee that it does not fit.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -181,7 +181,7 @@ def large_image(request):
     return _special_image(
         request,
         "large_image.dat",
-        ["dd", "if=/dev/zero", "of=large_image.dat", "bs=300M", "count=1"],
+        ["dd", "if=/dev/zero", "of=large_image.dat", "bs=500M", "count=1"],
     )
 
 


lluiscampos:
@mender-test-bot start pipeline for master too :+1:

tests/conftest.py
```diff
@@ -181,7 +181,7 @@ def large_image(request):
     return _special_image(
         request,
         "large_image.dat",
-        ["dd", "if=/dev/zero", "of=large_image.dat", "bs=300M", "count=1"],
+        ["dd", "if=/dev/zero", "of=large_image.dat", "bs=500M", "count=1"],
```
merlin-northern, line 184: I would create a variable out of it, maybe later we can even take it from the env. up to you.

merlin-northern (approved):
left one comment up to you :)

lluiscampos:
Build for `scarthgap` [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1353446407) (the target test passes) and for `master` [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1353448172) (everything :green_circle: )

PR #2624, opened 2024-07-01
MEN-7299: Update legacy client images with enlarged versions

lluiscampos:
The new images are based on the same tags and have been modified to have larger partitions so that they can fit latest yocto distributions.

The files have been moved/copied for consistency into a dedicated directory.


lluiscampos:

mzedel (approved):

PR #2628, opened 2024-07-12
fix: exclude the demo artifact in test_rbac_v2.py

tranchitella:
Changelog: None
Ticket: QA-733


backend-tests/tests/test_rbac_v2.py
```diff
@@ -1912,7 +1912,7 @@ def test_manage_releases(self, clean_mongo, test_case):
         api_client_v2.with_auth(tenant.users[0].token)
         r = api_client_v2.call("GET", deployments_v2.URL_RELEASES)
         assert r.status_code == 200
-        all_releases = r.json()
+        all_releases = [x for x in r.json() if not x['name'].startswith('mender-demo-artifact')]
```
merlin-northern, line 1915:
```suggestion
        excluded_artifact_prefix = 'mender-demo-artifact'
        all_releases = [x for x in r.json() if not x['name'].startswith(excluded_artifact_prefix)]
```
feel free to ignore. bonus points: `excluded_artifact_prefixes` (array)

merlin-northern (approved):
I have two comments: I do not know why we are excluding it, and are we sure that there is more than one in all_releases after exclusion? but I have no problem with this change.

merlin-northern (approved):
thank you

tranchitella:
> I have two comments: I do not know why we are excluding it, and are we sure that there is more than one in all_releases after exclusion? but I have no problem with this change.

@merlin-northern we have to exclude it because with https://github.com/mendersoftware/workflows-enterprise/pull/673 we restored the upload of the demo artifact in the default configuration (looks like it was broken before), and this breaks this specific test. I also updated the test to verify that we have two releases (so, indirectly, testing the demo artifact as well) before removing the demo one and checking the other.

kjaskiewiczz (approved):

PR #2603, opened 2024-05-03
MEN-5307: Deprecate `mtls-ambassador` in favor of `mender-gateway` and move the mTLS tests

lluiscampos:

PR #2636, opened 2024-08-04
chore: Remove redundant version definition for `mender-gateway`

lluiscampos:
It was already (wrong) defined in `git-versions.yml` and then re-introduced in `git-versions-enterprise.yml` without removing the former one. See 408078aa9c1ffbfdef9c76f6210f482a9ba89fc9

### DIFF
--- a/git-versions.yml
+++ b/git-versions.yml
@@ -64,8 +64,5 @@ services:
     monitor-client:
         image: mendersoftware/monitor-client:master
 
-    mender-gateway:
-        image: mendersoftware/mender-gateway:master
-
     mender-ci-workflows:
         image: mendersoftware/mender-ci-workflows:master


lluiscampos:
Relates to:
* https://github.com/mendersoftware/integration/pull/2603

git-versions.yml
```diff
@@ -64,8 +64,5 @@ services:
     monitor-client:
         image: mendersoftware/monitor-client:master
 
-    mender-gateway:
-        image: mendersoftware/mender-gateway:master
-
```
merlin-northern, line 69: as a review I checked that the definition exists in `./git-versions-enterprise.yml`:
```yaml
    mender-gateway:
        image: registry.mender.io/mendersoftware/mender-gateway:master
```

was it ever present in the `mendersoftware/mender-gateway:master`? (it is not at the moment)

lluiscampos: I am not sure I understand your question. Are you asking about the Docker container registry?

The name of the image was not really used (because mender-gateway was not a Docker image at the time) and it was originally wrongly added to the OS git versions file instead of the enterprise. It only matters now that we are publishing the Docker image

lluiscampos:

PR #2641, opened 2024-08-07
fix: aligned gui routing with prior behaviour & helm chart

mzedel:
passing gui e2e test run [here](https://gitlab.com/Northern.tech/Mender/gui/-/jobs/7526071258)


mzedel:

merlin-northern (approved):
I do not have a problem with that. I may have some problems with understanding it, but well, good that I trust you. right?

PR #2643, opened 2024-08-09
test: Make test more robust against restart latency

lluiscampos:
In my opinion, the previous test design was broken and worked by accident.

With an expire interval of 16s, it was inserting 4 alerts every 2 seconds and then waiting 4 seconds (total: 4 * 2 + 4 = 12) expecting then that the first two were already expired.

Now, because the restart of the service takes presumably 4 seconds approximately, this would work then. But if the restart is "fast", for example 3 seconds, then when checking the existing alerts we would find 3 instead of 2 (and, actually, if the service could restart in less than 2 seconds it would find the 4 of them still valid!).

Modify the test to separate them more, sleep more, and document more :)

Ticket: QA-745

### DIFF
--- a/tests/tests/test_monitor_client.py
+++ b/tests/tests/test_monitor_client.py
@@ -1564,17 +1564,17 @@ class TestMonitorClientEnterprise:
         """Tests the removal of older alerts from the persistent store"""
         alert_resend_interval_s = 4
         alert_max_age = 16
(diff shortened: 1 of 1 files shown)


lluiscampos:
I run the integration tests CI job 7 times, the test passed all of them :100:

merlin-northern (approved):

lluiscampos:
* 3.7.x

PR #2647, opened 2024-08-22
test: Update dependency for Docker addons image

lluiscampos:
Missing from 8a78b7b1471cba17e6a6855addd778320594bd93.


lluiscampos:
Thanks @mzedel for reporting - this image is not used by any test in `integration` repo so you/gui are the only real users :pray: 

By the way, it had nothing to with with Mac users, the image was straight broken since #2248 :see_no_evil:

mzedel (approved):
Thank you very much - tested and works flawlessly!

PR #2652, opened 2024-09-19
test: Fix unintentional inclusion of quotes in environment variable.

kacf:
When quotes are specified inline in Yaml variable values, they are interpreted as verbatim strings, not as quotation. So the filename ended up including quotes. This does not work anymore after 4d36b45511f4788d in meta-mender. Why it has worked until now is a good question; most likely shell commands accept the extra quotes, while they are not accepted in Python operations (where the new check is).


kacf:
For reference, this goes together with https://github.com/mendersoftware/meta-mender/pull/2170.

lluiscampos (approved):
TIL :+1:

PR #2653, opened 2024-09-23
test: Fix unintentional inclusion of quotes in environment variable.

kacf:
When quotes are specified inline in Yaml variable values, they are interpreted as verbatim strings, not as quotation. So the filename ended up including quotes. This does not work anymore after 4d36b45511f4788d in meta-mender. Why it has worked until now is a good question; most likely shell commands accept the extra quotes, while they are not accepted in Python operations (where the new check is).

(cherry picked from commit 7715452fdf108a12df882cc0317d827edc3f9080)


lluiscampos (approved):
3.6.x branch is EOL. Merge if you see fit, but don't worry if there unrelated build failures.

PR #2655, opened 2024-09-30
Run integration tests using mender-server container images

alfrunes:
New env variables to override compose image sources for mender images:
 * `MENDER_SERVER_REGISTRY` - overrides the registry component (docker.io|registry.mender.io)
 * `MENDER_SERVER_REPOSITORY` - overrides the repository namespace (mendersoftware|mender-server-enterprise)
 * `MENDER_SERVER_TAG` - overrides the image tag to use (main)
 * `MENDER_LEGACY` - adds the `-enterprise` suffix to services that used to have it.

TODO:
  - [x] Tests may need update for breaking changes

The new image references can be summarized as:
  * ${MENDER_SERVER_REGISTRY}/${MENDER_SERVER_REPOSITORY}/<service>:${MENDER_TAG}

The workflows-worker image has been replaced in the composition in favor of workflows (server) image.


tests/run.sh
```diff
@@ -79,20 +79,25 @@ function get_requirements() {
...
         echo "failed to download mender-artifact"
         exit 1
     fi
 
-    chmod +x downloaded-tools/mender-artifact
+    ar x --output "$EXTRACT_DIR" "$EXTRACT_DIR/mender-artifact.deb"
+    zstd -d "$EXTRACT_DIR/data.tar.zst"
+    tar -C $EXTRACT_DIR -xvf "$EXTRACT_DIR/data.tar" ./usr/bin/mender-artifact
+    mv $EXTRACT_DIR/usr/bin/mender-artifact downloaded-tools/mender-artifact
+    rm -rf $EXTRACT_DIR
```
lluiscampos, line 100: You could use `dpkg-deb --extract` instead of the custom extraction.

lluiscampos: See [here](https://github.com/mendersoftware/mender-server/blob/main/backend/services/create-artifact-worker/Dockerfile#L11) :+1:

alfrunes: I tried to avoid using `dpkg` since it's not shipped with Alpine and we're not installing the binary to the system. I updated the script and the requirements here https://github.com/mendersoftware/integration/pull/2655/commits/ebc9ee6b92bd189db89a38a40a2353fd88f559a7.

.gitlab-ci-full-integration.yml
```diff
@@ -0,0 +1,48 @@
...
+    PYTEST_XDIST_AUTO_NUM_WORKERS: "4"
+  before_script:
+    - |
+      docker login registry.mender.io \
+      --username $REGISTRY_MENDER_IO_USERNAME \
+      --password $REGISTRY_MENDER_IO_PASSWORD
+    - |
+      docker login docker.io \
+      --username $DOCKER_HUB_USERNAME \
+      --password $DOCKER_HUB_PASSWORD
```
lluiscampos, line 19: This won't work on on regular PRs (your branch is protected). How are we planning to solve it?

Edit: I see below that we plan to run these tests only on protected branches - that won't work when we develop new tests or debug existing ones.

.gitlab-ci.yml
```diff
@@ -1,21 +1,44 @@
...
+      - "false"
+      - "true"
+
+  DOCKER_VERSION:
+    description: "Docker version to use in jobs"
+    value: "27"
+
+  MENDER_TAG:
+    description: "Docker tag for running integration tests"
+    value: main
```
lluiscampos, line 41: What about the client? We should have a similar tag for it and update the docker-compose files accordingly

.gitlab-ci.yml
```diff
@@ -1,21 +1,44 @@
...
+      Run full integration tests.
+      NOTE: Can only run on protected branches
+    value: "false"
+    options:
+      - "false"
+      - "true"
+
+  DOCKER_VERSION:
+    description: "Docker version to use in jobs"
+    value: "27"
```
lluiscampos, line 37: Having it explicit in the `image:` tag will make it more visible and, hopefully, we will update it more often.

.gitlab-ci-full-integration.yml
```diff
@@ -0,0 +1,48 @@
+test:integration:
```
lluiscampos, line 1: The current pipeline takes more than 2 hours. The lowest hanging fruit to speed it up is to separate this into two jobs: open source tests and enterprise tests. This is the strategy on the current pipeline where it takes 1.5h + 1h (in parallel, so 1.5h human time) and I believe this one will run faster here.

lluiscampos (commented):
So I am the lucky one to review this one - yey!

Some preliminary comments below. I'll dig into the details a bit later today :memo:

.gitlab-ci.yml
alfrunes: True, but the task was about the `mender-server` so I tried to stick with it.

.gitlab-ci.yml
alfrunes: In my experience the fields defined at the job level tends to be forgotten and get scattered around in different files. By centralizing the versions to a single variable I was hoping to cope with this issue.

.gitlab-ci-full-integration.yml
alfrunes: I refactored the compatibility tests which amounts to a total of 1h execution time. Gonna give it another spin :wheel:

.gitlab-ci.yml
alfrunes: Added `MENDER_CLIENT_TAG`.

alfrunes:
I identified the issue with the integration tests: After introducing the MENDER_CLIENT_TAG env variable the image extraction fails. Working on a fix...

.gitlab-ci.yml
lluiscampos: We'll try it out :wink:

docker-compose.mender-gateway.commercial.yml
```diff
@@ -1,15 +1,16 @@
...
-        tty: true
-        privileged: true
-        environment:
-            - SERVER_URL=$SERVER_URL
-            - TENANT_TOKEN=$TENANT_TOKEN
+  #
+  # mender-gateway
+  #
+  mender-gateway:
+    image: registry.mender.io/mendersoftware/mender-gateway-qemu-commercial:mender-master
```
lluiscampos, line 6: Here also `${MENDER_CLIENT_TAG}`

The implementation detail is that `MENDER_CLIENT_TAG` is not really the client, but rather the "test device" or "virtual device" which contains client, addons, possibly gateway, possibly being rofs, etc

docker-compose.monitor-client.commercial.yml
```diff
@@ -1,15 +1,16 @@
...
-        tty: true
-        privileged: true
-        environment:
-            - SERVER_URL=$SERVER_URL
-            - TENANT_TOKEN=$TENANT_TOKEN
+  #
+  # monitor-client
+  #
+  mender-client:
+    image: registry.mender.io/mendersoftware/mender-monitor-qemu-commercial:mender-master
```
lluiscampos, line 6: Here also `${MENDER_CLIENT_TAG}`

docker-compose.yml
```diff
@@ -1,254 +1,253 @@
...
+
+  mender-nats:
+    image: nats:2.7.4-scratch
+    command: -js
+    networks:
+      - mender
 
 networks:
-    mender: null
+  mender: {}
```
lluiscampos, line 253: So what is this commit doing besides formatting? I don't get the "Complete the client compose files"

tests/conftest.py
```diff
@@ -119,7 +119,7 @@ def _extract_fs_from_image(request, client_compose_file, filename):
     )
 
     if not os.path.exists(os.path.join(THIS_DIR, filename)):
-        shutil.copy(os.path.join(d, filename), THIS_DIR)
+        shutil.move(os.path.join(d, filename), THIS_DIR)
```
lluiscampos, line 99: :clap:

testutils/infra/container_manager/docker_compose_manager.py
```diff
@@ -33,7 +33,6 @@ class DockerComposeNamespace(DockerComposeBaseNamespace):
     BASE_FILES = [
         COMPOSE_FILES_PATH + "/docker-compose.yml",
         COMPOSE_FILES_PATH + "/docker-compose.storage.minio.yml",
-        COMPOSE_FILES_PATH + "/docker-compose.testing.yml",
```
lluiscampos, line 37: This commit I don't understand, can you explain it?

The reason for removing the ports was to allow for parallel execution :shrug:

lluiscampos (changes requested):
Amazing work :rocket: Impressive tenacity getting this under control.

Thank you for the commit by commit experience. And thanks for all the small fixes for pre-existing things.

(Also, it was pretty cool that _while_ doing my review new commits popped up :laughing: )


Apart from the suggestions, comments, and questions below, there are couple of high level things I wanted to address:

- [ ] I think we should rename `${MENDER_TAG}` to `${MENDER_SERVER_TAG}`. The more explicit the better in light of the upcoming Mender Client and Mender Gateway releases
- [ ] Once we have migrated the docs away from the `demo` script, we have to remove it from here and remove the test here.
    * NOTE: I would wait until that other PR is approved by CEs just in case...
- [ ] Please squash most of your work before merging. Most of the commits are incomplete by themselves, so they should be logically grouped. If it is too much hastle, I would be okey with one monster commit (despite I don't like one single commit, it is better than ~30 non self contained commits)
- [ ] We need to set the GitLab trigger for nightly build on this
- [ ] And we need to update `mantra` to get nightly stats from here :scream: 
    * Please create a task for this to follow-up - obviously we won't do it ad-hoc.

.gitlab-ci-full-integration.yml
```diff
@@ -5,6 +5,9 @@ test:integration:
   stage: test
   tags:
     - mender-qa-worker-integration-tests
+  timeout: 24h
```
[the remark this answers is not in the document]
alfrunes: Okokok :upside_down_face: 
I was thinking about running these test on a tiny machine that can hum and sing integration tests all night :notes:
Still, I don't think it will take 24h - I'll adjust.

docker-compose.yml
alfrunes: The change is primarily to the other compose files. Making it possible to evaluate them without having to pull in the main `docker-compose.yml` file. I used it in the following commit when evaluating images when doing the `extract_fs` dance :dancers:

testutils/infra/container_manager/docker_compose_manager.py
alfrunes: See `extra/integration-testing/docker-compose.yml`. The ports are removed there instead. I had to do this as the services are extracted from `docker-compose config` which expected `docker-compose.testing.yml` to exist. However, this file is generated in a fixture later which in turn caused test_compat to fail.

tests/run.sh
```diff
@@ -79,20 +79,25 @@ function get_requirements() {
...
     # Download the tools
-    curl --fail "https://downloads.mender.io/mender-artifact/${MENDER_ARTIFACT_BRANCH}/linux/mender-artifact" \
-         -o downloaded-tools/mender-artifact \
-         -z downloaded-tools/mender-artifact
-
+    #FIXME: The distributed binaries are compiled for OpenSSL 1.1
+    #       As a workaround, we're installing the debian package (noble distribution).
+    EXTRACT_DIR=$(mktemp -d mender-artifact.XXXXXX)
+    curl --fail \
+        https://downloads.mender.io/repos/debian/pool/main/m/mender-artifact/mender-artifact_3.11.2-1%2bubuntu%2bnoble_amd64.deb \
```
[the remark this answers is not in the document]
alfrunes: I made this as a temporary workaround, but it's funny how it tends to stick around :sweat_smile:
I'll do it more proper :+1:

alfrunes:
> Once we have migrated the docs away from the demo script, we have to remove it from here and remove the test here.
NOTE: I would wait until that other PR is approved by CEs just in case...

I'm not sure about this one since it will continue to be referenced in the old documentation. Maybe it would be better to add a deprecation notice every time you invoke the script?

alfrunes:
> We need to set the GitLab trigger for nightly build on this

I will add a schedule as soon as this PR is merged to `master`.

> And we need to update mantra to get nightly stats from here 😱
> - Please create a task for this to follow-up - obviously we won't do it ad-hoc.

Done ([QA-789](https://northerntech.atlassian.net/browse/QA-789)).

testutils/infra/container_manager/docker_compose_manager.py
lluiscampos: Got it. Thank you.

lluiscampos:
> > Once we have migrated the docs away from the demo script, we have to remove it from here and remove the test here.
> > NOTE: I would wait until that other PR is approved by CEs just in case...
> 
> I'm not sure about this one since it will continue to be referenced in the old documentation.

The old documentation will refer to 3.7.x (or 3.6.x, etc) Git tags of the `integration` repo.

> Maybe it would be better to add a deprecation notice every time you invoke the script?

I think that we should not expose two sets of docker compose files to the users. It is okey to keep a second set here in `integration` for the tests (with the long term vision of removing them and rework the test framework), but I would remove the demo script completely to signal users that relied on it that something has changed (and then a user goes into the docs and figures out where is the new demo server composition).

tests/conftest.py
```diff
@@ -388,6 +388,7 @@ def run_remote_command(instances, command):
                 "mender-monitor",
                 "mender-gateway",
                 "mender-client",
+                "mender",
```
lluiscampos, line 391: This is really old, client 2.2 and older :laughing: 

(but I understand you added it to print errors from compat test - :ok_hand: )

.env
```diff
@@ -0,0 +1,14 @@
+MENDER_SERVER_REGISTRY=docker.io
+MENDER_SERVER_ENTERPRISE_REGISTRY=registry.mender.io
+MENDER_SERVER_REPOSITORY=mendersoftware
+MENDER_SERVER_ENTERPRISE_REPOSITORY=mender-server-enterprise
+MENDER_SERVER_TAG=main
+MENDER_CLIENT_TAG=mender-master
```
lluiscampos, line 6: Can we have `MENDER_ARTIFACT_VERSION` here also?

lluiscampos (approved):
It looks good to me now!

A part from the two comments below:
* Update the PR title :monocle_face: 
* Remember to fix the default pipeline. The bot is complaining because the default pipeline cannot be run:

image

Pre-aproving :rocket:

tests/conftest.py
alfrunes: :older_adult:

.env
alfrunes: It won't have any effect as this is a variable for the `tests/run.sh` script...
Unless we source this file from the script. Yes! Let's do it :+1:

alfrunes:
> It looks good to me now!
> 
> A part from the two comments below:
> 
> * Update the PR title 🧐
> * Remember to fix the default pipeline. The bot is complaining because the default pipeline cannot be run:

What have I done! :scream_cat:
Actually, it seems like my compose fixes to remove ports introduced another issue in a couple of tests. I'll try to fix those as well.

alfrunes:
> > > Once we have migrated the docs away from the demo script, we have to remove it from here and remove the test here.
> > > NOTE: I would wait until that other PR is approved by CEs just in case...
> > 
> > 
> > I'm not sure about this one since it will continue to be referenced in the old documentation.
> 
> The old documentation will refer to 3.7.x (or 3.6.x, etc) Git tags of the `integration` repo.

In that case, consider the demo script history.

alfrunes:
Squashed 38 commits into 21.
Last full integration test run: https://gitlab.com/Northern.tech/Mender/integration/-/pipelines/1516708776
I'm merging this PR so I can set up the daily schedule and tick the last box :ballot_box_with_check:

PR #2659, opened 2024-10-29
QA-794: Release tool work to prepare for the "non server" release

lluiscampos:


lluiscampos:
@oldgiova Thanks for volunteering :grin:

git-versions-enterprise.yml
```diff
@@ -31,9 +37,6 @@ services:
     mtls-ambassador:
```
oldgiova, line 37: We can get rid of the deprecated `mtls-ambassador` as well maybe?

oldgiova (approved):
LGTM! Only an optional comment about the `mtls-gateway` service

git-versions-enterprise.yml
lluiscampos: The deprecated repos are marked as `release_component: false`. We still want them there because this tool is used to check for older versions (for example: mtls-abassador is deprecated but it was part of Mender 3.7 which we sill support). Also note that there are even older repos like mender-conductor, etc.

I'll keep it around to not break even more the Mender 3.7 LTS release process.

PR #2663, opened 2024-11-05
QA-794: Deprecate `versions.json` updates

lluiscampos:
We don't know how this will look in the future, but for the time being we should stop updating a file that won't anymore represent what we are supporting or not.


mzedel:
Hm right, that will also mean there is going to be misleading info shown in hM from the next deployment on... from a user PoV I would rather have this adjusted than deactivated.

lluiscampos:
>  I would rather have this adjusted than deactivated.

What do you suggest? Is it only the client part the UI needs?

The thing is that this generation uses the release_tool, which I will iteratively breaking a part. We need to base it on something else but we have not yet decided what will that be.

mzedel:
> What do you suggest? Is it only the client part the UI needs?

I suggest we adjust this to work with the new release process and the client tools - the UI uses this as a single place to get the info whether to show new features (currently basing it on the integration version - so that will have to change), link to the mender-client/ -artifact/ -cli/ -convert... downloads and changelogs.

lluiscampos:
> I suggest we adjust this to work with the new release process and the client tools - 

There new release process is for the Mender Server. Everything else has been left TBD for later iterations.

But ok, I'll keep the file given the dependency from the UI that I did not remember.

PR #2667, opened 2024-11-06
Integration test environment fixes

alfrunes:
Some issues discovered working on QA-766.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -54,8 +54,6 @@ while [ -n "$1" ]; do
     shift
 done
 
-source ../.env
-
 function get_requirements() {
     # Download what we need.
     mkdir -p downloaded-tools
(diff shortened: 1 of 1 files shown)


docker-compose.mender-gateway.commercial.yml
```diff
@@ -3,7 +3,7 @@ services:
   # mender-gateway
   #
   mender-gateway:
-    image: registry.mender.io/mendersoftware/mender-gateway-qemu-commercial:${MENDER_CLIENT_TAG}
```
alfrunes, line 6: Seems like this image is following client versioning :monocle_face:

tests/run.sh
```diff
@@ -80,9 +78,12 @@ function get_requirements() {
...
     EXTRACT_DIR=$(mktemp -d mender-artifact.XXXXXX)
-    curl --fail \
-        "https://downloads.mender.io/repos/debian/pool/main/m/mender-artifact/mender-artifact_${MENDER_ARTIFACT_VERSION}-1%2bubuntu%2bnoble_amd64.deb" \
-        -o "$EXTRACT_DIR/mender-artifact.deb"
+    (
+        test -z "$MENDER_ARTIFACT_VERSION" && source ../.env
+        curl --fail \
+            "https://downloads.mender.io/repos/debian/pool/main/m/mender-artifact/mender-artifact_${MENDER_ARTIFACT_VERSION}-1%2bubuntu%2bnoble_amd64.deb" \
+            -o "$EXTRACT_DIR/mender-artifact.deb"
+    )
```
alfrunes, line 86: Have to execute this in a sub-shell since the `source` will overwrite the entire environment, but the existing environment should take precedence.

oldgiova (approved):

docker-compose.mender-gateway.commercial.yml
lluiscampos: Yes, [we talked about it](https://github.com/mendersoftware/integration/pull/2655#discussion_r1812117838).

It is not the client version but the "virtual device version", if you want a call it like that. Basically what used to be the Mender Bundle.

However I will be splitting them this month during the non-server release, so I'll revisit these versions then.

docker-compose.mender-gateway.commercial.yml
```diff
@@ -3,7 +3,7 @@ services:
...
+    image: registry.mender.io/mendersoftware/mender-gateway-qemu-commercial:mender-master
```
lluiscampos, line 6: But then, why do we set this back to master? It should be `MENDER_CLIENT_TAG`

lluiscampos (commented):
I don't understand the change for the Mender Gateway tag. See my comment below.

Can you wait with this? Because with the release that I am trying to get out I will split the versions correctly (as you expect).

lluiscampos (approved):

PR #2668, opened 2024-11-06
QA-767: re-enable TestMTLSClientEnterprise

alfrunes:


alfrunes:
The failing test failed because of generate delta worker not ignoring TLS verification when it should (https://github.com/mendersoftware/mender-server-enterprise/pull/158).
It seems like the MTLS tests are passing - it could potentially be a symptom of state leaking from another test.
Regardless, I'd like to add the addition to the exception handler pytest fixture for dumping the database (it helped me find the issue of the failing job).

lluiscampos (approved):

PR #2671, opened 2024-11-12
ci(compose): Expand compose .env with image references for mender-gateway

alfrunes:
Ticket: QA-804


lluiscampos (approved):
Splendid :clap:

PR #2672, opened 2024-11-12
ci(QA-793): Enable TestDBMigration

alfrunes:


alfrunes:
The culprit was the client running out of memory. I modified the `mender-qemu` script from the legacy client container entrypoint to increase the memory from 256M to 384M.

lluiscampos (approved):
This is unexpected, what was/is the culprit of the new memory usage? The change in the CI runners? In the base image? Something in the docker composition?

In any case I won't push for more investigations. Thanks for fixing :see_no_evil:

alfrunes:
> This is unexpected, what was/is the culprit of the new memory usage? The change in the CI runners? In the base image? Something in the docker composition?

It seem to be an issue with the old HTTP2 library. It seems buffer excessive amounts of data in memory (possibly the entire artifact payload?). The crash seem to happen as the device is reading the payload from stream and the `http2` package grows the buffer.

lluiscampos:
> > This is unexpected, what was/is the culprit of the new memory usage? The change in the CI runners? In the base image? Something in the docker composition?
> 
> It seem to be an issue with the old HTTP2 library. It seems buffer excessive amounts of data in memory (possibly the entire artifact payload?). The crash seem to happen as the device is reading the payload from stream and the `http2` package grows the buffer.

Thanks for the detailed research.

But we have not changed the legacy client image nor, that I know, the generated Artifact for the test update. So the Artifact used to fit but no more, which is still a mystery.

PR #2673, opened 2024-11-13
chore: Add pytest marker for tests that uses mender-gateway

alfrunes:


tests/pytest.ini
```diff
@@ -18,3 +18,6 @@ testpaths = tests
 #
 # Use v1 of xunit format (default will change in pytest 6.0)
 junit_family=xunit1
+
+markers =
+    mender_gateway: marks tests that uses mender-gateway
```
lluiscampos, line 23: As I see it, this marker is only for the "mender-gateway as a container" tests. If so, it should be more explicit. I understand that this is the purpose of this PR (not much justified in the commit message, if you ask me): to select only the tests that need a run from `mender-gateway` CI trigger.

If this is not the case, and we want to gather all tests that use mender-gateway code, then we need to mark the "mender-gateway on a device" tests in addition, see `TestMenderGatewayOpenSource` and `TestMenderGatewayEnterprise`. This will make it that from today's `mender-gateeway` CI trigger we run two unnecessary tests, but gets us better prepared to whenever we can rework the virtual device image to be built from `mender-gateway` repo too.

alfrunes: Correct, the marker is intended for tests using the `mender-gateway` application image, not the qemu image with mender-gateway installed.
I see the problem of using this marker name as it occupies the namespace of all markers containing `mender_gateway` as a substring.
How about using two different markers?
 * mender_gateway_docker
 * mender_gateway_qemu

lluiscampos: Exactly. The suggested names look good :100:

tests/tests/test_mtls.py
```diff
@@ -85,6 +85,7 @@ def make_script_artifact(artifact_name, device_type, output_path):
     return get_script_artifact(script, artifact_name, device_type, output_path)
 
 
+@pytest.mark.mender_gateway
```
lluiscampos, line 88:
```suggestion
@pytest.mark.mender_gateway_qemu
```

lluiscampos (changes requested):
Here we still need to correct the existing marker (see comment below) and add the markers to the rest of the tests

PR #2675, opened 2024-11-14
Fix nightlies, step 1

tranchitella:


lluiscampos:
FYI @tranchitella - the mender-test-bot pipeline does not run the integration tests anymore, so that one you can skip. If you want to run the tests, you need first to protect the branch and then run manually from GitLab setting `RUN_TESTS_FULL_INTEGRATION` to true.

tests/run.sh
```diff
@@ -34,6 +34,7 @@ usage() {
     echo "Recognized Environment Variables:"
     echo
     echo "TESTS_IN_PARALLEL_INTEGRATION        The number of parallel jobs for pytest-xdist"
+    echo "TESTS_SELECTION                      The pytest selector to use for selecting tests"
```
alfrunes, line 37: Can we instead expose built-in `PYTEST_ADDOPTS` to the CI?

PR #2679, opened 2024-11-21
chore: disable naive 401 API endpoints tests for auth_requests.

merlin-northern:
Ticket: MEN-7714


merlin-northern:
I went insane, this is the wrong repo.

PR #2682, opened 2024-11-25
Tentative fix: preventing too many open files issue

oldgiova:
This issue could be raised on integration tests

Ticket: None
Changelog: None

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,6 +56,11 @@ services:
     extends:
       file: common.yml
       service: mender-base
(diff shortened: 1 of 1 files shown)


kjaskiewiczz (approved):
approving, but when I was testing it locally, this setting was not enough to make it work

oldgiova:
> approving, but when I was testing it locally, this setting was not enough to make it work

So is it reproducible? Could you please help me reproduce it tomorrow?

PR #2686, opened 2024-11-30
ci: Control parallelization parameters from CI variables

lluiscampos:
Rework a bit the parallelization parameters and expose both as CI environment variables, allowing now for a user (or another pipeline trigger) to customize the parallelization.

For simplification, remove `PYTEST_XDIST_AUTO_NUM_WORKERS` and use instead the already existing CI variable.

### DIFF
--- a/.gitlab-ci-full-integration.yml
+++ b/.gitlab-ci-full-integration.yml
@@ -6,9 +6,7 @@ test:integration:
   tags:
     - mender-qa-worker-integration-tests
   timeout: 10h
-  parallel: 4
-  variables:
-    PYTEST_XDIST_AUTO_NUM_WORKERS: "4"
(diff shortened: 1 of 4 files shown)


.gitlab-ci-full-integration.yml
```diff
@@ -6,9 +6,7 @@ test:integration:
   tags:
     - mender-qa-worker-integration-tests
   timeout: 10h
-  parallel: 4
-  variables:
-    PYTEST_XDIST_AUTO_NUM_WORKERS: "4"
+  parallel: $CI_JOBS_IN_PARALLEL_INTEGRATION
```
lluiscampos, line 9: This won't work :cry: See https://gitlab.com/gitlab-org/gitlab/-/issues/11549

PR #2688, opened 2024-12-02
ci: Control parallelization parameters from CI variables

lluiscampos:
Rework a bit the parallelization parameters and expose both as CI environment variables, allowing now for a user (or another pipeline trigger) to fully customize the parallelization.

For simplification, remove `PYTEST_XDIST_AUTO_NUM_WORKERS` and use instead the already existing CI variable.

Ideally, we would have done this with just setting `parallel` to the variable (iow `parallel: $CI_JOBS_IN_PARALLEL_INTEGRATION`), but this won't correctly work in GitLab CI due to some parsing issues.

See:
* https://gitlab.com/gitlab-org/gitlab/-/issues/11549
* https://gitlab.com/gitlab-org/gitlab/-/issues/426904

And our previous naive attempt at:
* https://github.com/mendersoftware/integration/pull/2686

### DIFF
--- a/.gitlab-ci-full-integration-generator.yml
+++ b/.gitlab-ci-full-integration-generator.yml
@@ -1,0 +1,35 @@
+#
+# Ideally we would have used "parallel: $CI_JOBS_IN_PARALLEL_INTEGRATION" to run the tests, but it
+# is not supported to use an integer variable here. Do a blame on these lines for more context.
+#
+# To workaround this issue, we set up "Dynamic child pipelines" to control how many CI jobs to
+# run tests (iow how many child pipelines) to launch. For an overview of this features see:
+# https://docs.gitlab.com/ee/ci/pipelines/downstream_pipelines.html#dynamic-child-pipelines
+#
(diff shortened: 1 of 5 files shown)


lluiscampos:
This is how it would look :point_down: 

image

In the above example I have used for example 3 CI jobs to validate the behavior of the CI variables, but by default it would be 4.

Also, I know this seems like over-engineering the pipeline, but I couldn't think about a simpler workaround. I suggest we review and merge this one ASAP in order to unblock release builds, and if someone has ideas on how to make this cleaner we can follow-up later on.

danielskinstad (approved):
I found it a bit confusing at first, especially with how the `.gitlab-ci-full-integration-template.yml` and `.gitlab-ci-full-integration-generator.yml` were used. Perhaps a comment above the envsubst to more explicitly explain what's happening?

lluiscampos:
> I found it a bit confusing at first (...)

@danielskinstad  Good feedback, thank you. Can you read again the comment and see if it is bra nok?

PR #2696, opened 2024-12-06
CI: Increase system limits and amount of RAM available for legacy client.

kjaskiewiczz:


.gitlab-ci-full-integration-template.yml
```diff
@@ -50,6 +50,12 @@ test:integration:$CI_NODE_INDEX:
         echo "Timeout waiting for docker to start"
         exit 1;
       fi
+    # Increase system limits to make sure the tests are not limited while
+    # running with high parallelism on a single VM
+    - sysctl -w fs.inotify.max_user_instances=1024
```
danielskinstad, line 55: How did you land on these numbers? Perhaps worth a brief comment either here or in the commit for future reference

danielskinstad (approved):
Looks good to me, just a question regarding the system limits.

.gitlab-ci-full-integration-template.yml
kjaskiewiczz: > How did you land on these numbers?

By guessing, sorry.
In the past we were setting it to this number too, without good explanation. The default is 128 and people on the internet will told you it's not enough in many cases. Some are setting it to 8192, some to 1024. In our case 1024 was enough to get rid of "too many open files" type errors.
So long story short, I don't have a good explanation for any of those numbers. The only thing I can tell is that they are higher than defaults.

PR #2704, opened 2025-01-16
test: fix and enable test_mtls_enterprise_hsm for engines

danielskinstad:
After updating libp11, openssl can't seem to load the pkcs11 module. Explicitly set `MODULE_PATH` to the `libsofthsm2.so` path.

Ticket: QA-824

### DIFF
--- a/tests/tests/test_mtls.py
+++ b/tests/tests/test_mtls.py
@@ -127,6 +127,7 @@ pkcs11 = pkcs11_section
 
 [pkcs11_section]
 engine_id = pkcs11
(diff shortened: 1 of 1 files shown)


lluiscampos (approved):
I couldn't find any good documentation for the config file online. In the `libp11` repo they have an [example configuration](https://github.com/OpenSC/libp11?tab=readme-ov-file#using-the-engine-from-the-command-line) that uses the `MODULE_PATH`, so that seems to be a legit configuration option :+1: 

I love these one-liner PRs that are super simple to review, yet you can see that there was a lot of work to figure out. Well done!

PR #2711, opened 2025-01-23
MEN-7900: Improve test to catch bugs with the state machine hanging

lluiscampos:
The test was way to naive to test that we have "less than X loops". This statement cannot catch bugs where the machine does not loop (iow does 1 loop).

Set the min to 4 to ensure at least two loops for each of the events: polling for update and sending inventory.

This test fix would have caught bug MEN-7900.

### DIFF
--- a/tests/tests/test_state_scripts.py
+++ b/tests/tests/test_state_scripts.py
@@ -907,12 +907,14 @@ class BaseTestStateScripts(MenderTesting):
 
             # Test cases with an expectation of success/failure shall do only 1 iteration
             # Test cases with None expectation will loop through the error sequence in a loop, but still
-            # we want to make sure that it is reasonable (i.e. looping with the correct time intervals).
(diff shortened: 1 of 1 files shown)


lluiscampos:
Running integration tests with current client @ [1637924946](https://gitlab.com/Northern.tech/Mender/integration/-/pipelines/1637924946) - we expect the test to fail

lluiscampos:
New run @ [Pipeline 1638045578](https://gitlab.com/Northern.tech/Mender/integration/-/pipelines/1638045578) - still expected to fail

lluiscampos:
After the [client update](https://github.com/mendersoftware/mender/pull/1724), the test now passes @ [Pipeline 1638657251](https://gitlab.com/Northern.tech/Mender/integration/-/pipelines/1638657251) - the failing test is unrelated to this PR and due to have run the pipeline with parallelism.

danielskinstad (approved):

PR #2714, opened 2025-01-28
QA-828 - trigger integrationt tests from mender-qa

danielskinstad:
Allows us to trigger integration tests with images built in mender-qa

Ticket: QA-828

### DIFF
--- a/.env
+++ b/.env
@@ -4,14 +4,36 @@ MENDER_SERVER_REPOSITORY=mendersoftware
 MENDER_SERVER_ENTERPRISE_REPOSITORY=mender-server-enterprise
 MENDER_SERVER_TAG=main
 
+MENDER_CLIENT_REGISTRY=docker.io
+MENDER_CLIENT_ENTERPRISE_REGISTRY=registry.mender.io
+MENDER_CLIENT_REPOSITORY=mendersoftware
+
+MENDER_CLIENT_DOCKER_IMAGE=mender-client-docker
+MENDER_CLIENT_DOCKER_TAG=mender-master
+
+MENDER_CLIENT_DOCKER_ADDONS_IMAGE=mender-client-docker-addons
+MENDER_CLIENT_DOCKER_ADDONS_TAG=mender-master
+
+MENDER_CLIENT_QEMU_IMAGE=mender-client-qemu
+MENDER_CLIENT_QEMU_TAG=mender-master
+
+MENDER_CLIENT_QEMU_ROFS_IMAGE=mender-client-qemu-rofs
+MENDER_CLIENT_QEMU_ROFS_TAG=mender-master
+
+MENDER_GATEWAY_QEMU_COMMERCIAL_IMAGE=mender-gateway-qemu-commercial
+MENDER_GATEWAY_QEMU_COMMERCIAL_TAG=mender-master
+
+MENDER_MONITOR_QEMU_COMMERCIAL_IMAGE=mender-monitor-qemu-commercial
+MENDER_MONITOR_QEMU_COMMERCIAL_TAG=mender-master
+
+MENDER_QEMU_ROFS_COMMERCIAL_IMAGE=mender-qemu-rofs-commercial
+MENDER_QEMU_ROFS_COMMERCIAL_TAG=mender-master
+
 # Artifacts from mender-gateway
+MENDER_GATEWAY_IMAGE=mender-gateway
 MENDER_GATEWAY_REGISTRY=registry.mender.io
 MENDER_GATEWAY_REPOSITORY=mendersoftware
 MENDER_GATEWAY_TAG=master
-
-# Virtual devices, built from mender-qa
-MENDER_CLIENT_TAG=mender-master
-MENDER_GATEWAY_QEMU_TAG=mender-master
 
 MENDER_ARTIFACT_VERSION=3.11.2
 
--- a/docker-compose.client.rofs.commercial.yml
+++ b/docker-compose.client.rofs.commercial.yml
@@ -4,4 +4,4 @@ services:
     # mender-client
     #
     mender-client:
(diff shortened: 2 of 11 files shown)


extra/release_tool.py
```diff
@@ -433,6 +433,18 @@ def _is_known_yml_file(entry):
     ]
 
 
+def load_env_variables_from_file(file_path):
```
danielskinstad, line 436: Initially I used `dotenv` - but adding a new dependency causes issues and then we'll need to change how we get the requirements in mender-qa etc.

lluiscampos: I see.

I am sorry that you had to deal with this file. This is because of some calls in mender-qa pipeline? Can you point me to the errors that we were having?

We should be able to run the pipeline as we want without really using the release_tool.

.env
```diff
@@ -4,15 +4,37 @@ MENDER_SERVER_REPOSITORY=mendersoftware
...
+MENDER_CLIENT_QEMU_ROFS_TAG=mender-master
+
+MENDER_GATEWAY_QEMU_COMMERCIAL_IMAGE=mender-gateway-qemu-commercial
+MENDER_GATEWAY_QEMU_COMMERCIAL_TAG=mender-master
+
+MENDER_MONITOR_QEMU_COMMERCIAL_IMAGE=mender-monitor-qemu-commercial
+MENDER_MONITOR_QEMU_COMMERCIAL_TAG=mender-master
+
+MENDER_QEMU_ROFS_COMMERCIAL_IMAGE=mender-qemu-rofs-commercial
+MENDER_QEMU_ROFS_COMMERCIAL_TAG=mender-master
```
lluiscampos, line 30: Once we are renaming,

I would go as far as to say that all of these must have a "VIRTUAL_DEVICE" prefix somehow.

```suggestion
VIRTUAL_DEVICE_DOCKER_IMAGE=mender-client-docker
VIRTUAL_DEVICE_DOCKER_TAG=mender-master

VIRTUAL_DEVICE_DOCKER_ADDONS_IMAGE=mender-client-docker-addons
VIRTUAL_DEVICE_DOCKER_ADDONS_TAG=mender-master

VIRTUAL_DEVICE_QEMU_IMAGE=mender-client-qemu
VIRTUAL_DEVICE_QEMU_TAG=mender-master

VIRTUAL_DEVICE_QEMU_ROFS_IMAGE=mender-client-qemu-rofs
VIRTUAL_DEVICE_QEMU_ROFS_TAG=mender-master

VIRTUAL_DEVICE_QEMU_GATEWAY_IMAGE=mender-gateway-qemu-commercial
VIRTUAL_DEVICE_QEMU_GATEWAY_TAG=mender-master

VIRTUAL_DEVICE_QEMU_MONITOR_IMAGE=mender-monitor-qemu-commercial
VIRTUAL_DEVICE_QEMU_MONITOR_TAG=mender-master

VIRTUAL_DEVICE_QEMU_ROFS_COMMERCIAL_IMAGE=mender-qemu-rofs-commercial
VIRTUAL_DEVICE_QEMU_ROFS_COMMERCIAL_TAG=mender-master
```

Does it sound too crazy? Some reasons to switch:

* The `client` is not really a good keyword: all of the images have the client
* The `commercial` for gateway and monitor are not adding any info, these components are closed source

extra/release_tool.py
danielskinstad: I don't remember exactly in which job - but it was this:
```
daniel@daniel-thinkpad:~/.../repos/integration$ extra/release_tool.py --map-name docker mender-client-docker docker_url
Traceback (most recent call last):
  File "/home/daniel/Documents/repos/integration/extra/release_tool.py", line 3286, in <module>
    main()
  File "/home/daniel/Documents/repos/integration/extra/release_tool.py", line 3261, in main
    do_map_name(args)
  File "/home/daniel/Documents/repos/integration/extra/release_tool.py", line 2969, in do_map_name
    print("%s/%s" % (data[result.name]["image_prefix"], result.name))
                     ~~~~^^^^^^^^^^^^^
KeyError: 'mender-client-docker'
```

lluiscampos: I think we can completely remove this from mender-qa. I will explain and I can help out doing so.

lluiscampos: Discussed f2f,

We will amend mender-qa CI scripts to not use `--map-name` option anymore and just hardcode the "docker_url" in the few places that it is used (note that these "docker_url" are already hardcoded somewhere else in the repo)

The idea being that we want to stop parsing docker-compose files from the release tool - the tool should only be concerned about git versions

.env
danielskinstad: After talking F2F we decided to postpone the renaming to not create confusion between the variables and the docker image names.

extra/release_tool.py
danielskinstad: Talked f2f, hardcoded the docker-urls to avoid having to use the map.

.env
```diff
@@ -4,15 +4,37 @@ MENDER_SERVER_REPOSITORY=mendersoftware
 MENDER_SERVER_ENTERPRISE_REPOSITORY=mender-server-enterprise
 MENDER_SERVER_TAG=main
 
+MENDER_REGISTRY=registry.mender.io
+MENDER_DOCKER_REGISTRY=docker.io
```
lluiscampos, line 8: To follow the convention above (server) and below (gateway), these should be:
```suggestion
MENDER_CLIENT_REGISTRY=docker.io
MENDER_CLIENT_ENTERPRISE_REGISTRY=registry.mender.io
```

I apologize I didn't comment on this on the first pass of this PR.

lluiscampos (approved):

danielskinstad:
> @danielskinstad We should also update the variables in `.gitlab-ci.yml`. Otherwise when running the pipeline from GitLab UI the values are incomplete (or broken, what is worst, as for example we have removed `MENDER_CLIENT_TAG`).

Nice catch!

PR #2724, opened 2025-02-24
test: Mark the correct flaky test with xfail

alfrunes:


oldgiova (approved):

tests/tests/test_mender_connect.py
```diff
@@ -355,6 +355,7 @@ def assert_env(self, docker_env):
 
 
 class _TestRemoteTerminalBaseBogusProtoMessage:
+    @pytest.mark.xfail(reason="QA-916")
```
lluiscampos, line 358: I don't get it. Why do we expect to fail? Broken test? Missing implementation?

Can we add more reasoning to the JIRA ticket, commit message, comment in the code, something? Otherwise we make it really hard for the ourselves of the future to understand what is the state of the test and/or product.

alfrunes: I updated the ticket info.

PR #2727, opened 2025-02-24
QA-951: Capture junit reports in GitLab CI

lluiscampos:
The old pipeline used to rename the reports to distinguish between different CI jobs.

### DIFF
--- a/.gitlab-ci-full-integration-template.yml
+++ b/.gitlab-ci-full-integration-template.yml
@@ -64,7 +64,7 @@ test:integration:$CI_NODE_INDEX:
     when: always
     paths:
       - tests/mender_test_logs
-      - tests/results_full_integration.xml
-      - tests/report_full_integration.html
+      - tests/results.xml
+      - tests/report.html
     reports:
-      junit: results_full_integration.xml
+      junit: tests/results.xml


oldgiova (approved):

lluiscampos:

lluiscampos:

lluiscampos:
@mender-test-bot start integration pipeline :rocket:

lluiscampos:
It partially works: we get the tests parsed but only for the child pipeline:

image

@alfrunes @oldgiova @danielskinstad if we are confident that we can run now always test in parallel (at least two jobs in parallel) we could then revert https://github.com/mendersoftware/integration/pull/2688 ? That would make the test reports show correctly in the parent pipeline and, in addition, simplify the pipeline code :innocent:

lluiscampos:
> @alfrunes @oldgiova @danielskinstad if we are confident that we can run now always test in parallel (at least two jobs in parallel) we could then revert #2688 ? That would make the test reports show correctly in the parent pipeline and, in addition, simplify the pipeline code 😇

Well, if we revert that we will get the number of CI jobs hard-coded, which gives us less flexibility when tests are unstable :thinking:

danielskinstad:
> > @alfrunes @oldgiova @danielskinstad if we are confident that we can run now always test in parallel (at least two jobs in parallel) we could then revert #2688 ? That would make the test reports show correctly in the parent pipeline and, in addition, simplify the pipeline code 😇
> 
> Well, if we revert that we will get the number of CI jobs hard-coded, which gives us less flexibility when tests are unstable 🤔

I think it's fine to leave it as is. It's nice to have control of the parallelization

lluiscampos:

PR #2731, opened 2025-02-27
ci: Remove dynamically generating full integration pipelines

alfrunes:
The second level of triggering masks CI variables passed from external triggers. This prevents passing credentials to internal CI registries for artifact built in other projects (e.g. mender-server).


alfrunes:
My assumption was incorrect.

PR #2733, opened 2025-03-03
ci: Enable test_bogus_proto_message

alfrunes:
The test uncovered a race condition when connecting and provisioning a device concurrently.
The issue was addressed here:
https://github.com/mendersoftware/mender-server/pull/486


lluiscampos (approved):
Thank you for following up :+1:

PR #2741, opened 2025-04-10
MEN-8269: Expose `MENDER_ARTIFACT_VERSION` through CI

lluiscampos:


.gitlab-ci.yml
```diff
@@ -129,6 +129,10 @@ variables:
     description: "Mender QEMU commercial read-only filesystem image tag"
     value: mender-master
 
+  MENDER_ARTIFACT_VERSION:
+    description: "mender-artifact tool version"
+    value: 3.11.2
```
mzedel, line 134: just for my understanding: is there a reason not to use 4.0.0?

mzedel (approved):

.gitlab-ci.yml
lluiscampos: We will soon (today) release 4.1.0, so my idea was to update directly to that - having it in CI already allows me to test the right thing :ok_hand:

PR #2752, opened 2025-06-04
fix: Revert "ci: Add mender-flash to mender-client-docker-addons"

vpodzime:
This reverts commit
26a28f61c2f7ce3eef49201037e7b556e39e9e88. `mender-flash` is not needed in the client container as it never performs a rootfs-image update.

Ticket: QA-993
Changelog: none


danielskinstad (approved):

oldgiova:
Force merging, as requested by @vpodzime

PR #2773, opened 2025-07-21
Refactor fs extraction and fix potential race condition 

danielskinstad:


danielskinstad:

danielskinstad:
@mender-test-bot sync

danielskinstad:

tests/conftest.py
```diff
@@ -89,89 +86,69 @@ def _extract_fs_from_image(request, client_compose_file, filename):
...
+        )
 
-def _image(request, compose_file, filename):
-    return _extract_fs_from_image(
-        request, os.path.join(THIS_DIR, "..", compose_file), filename
-    )
+    if not os.path.exists(dst_path):
+        raise FileNotFoundError(f"Expected ext4 not found: {dst_path}")
 
+    yield filename
```
michalkopczan, line 106: So this is the main thing really. Do I understand correctly? We use file lock, and you added yield here?

danielskinstad: No, this is just to avoid having to do 
```
   def cleanup():
        shutil.rmtree(d, ignore_errors=True)

    request.addfinalizer(cleanup)
 ```
 The main thing was to lock the function; previously two workers could enter the function at the same time.

danielskinstad: Previously we had `run_if_non_existent` which took a lock, checked if the file exited and returned. If two workers enters run_if_non_existent and there were no files, both would enter since we had released the lock. So now we're making sure that we have the lock before calling extract_fs_from_image

PR #2777, opened 2025-07-28
test: decorate functions with redo.retriable

danielskinstad:
Add a `redo.retriable` decorator the dns_request and refactor out the scp command to its own function with the same deocrator.

This will make sure that both the dns request and the scp commands will be retried properly in case it runs before the portwarding is ready.

Ticket: QA-1107


danielskinstad:

PR #2778, opened 2025-07-31
feat: add `SERVER_IP` arg to mender-client-docker-addons

danielskinstad:
This will allow us to use this as virtual device for testing mender-gateway, e.g.:

```
docker run -it -p 85:85  \
    -e SERVER_IP="$SERVER_IP" \
    -e SERVER_URL="https://gateway.docker.mender.io" \
    -e TENANT_TOKEN="$TENANT_TOKEN" \
    mendersoftware/mender-client-docker-addons
```

It will add an entry to /etc/hosts so we can map SERVER_URL to SERVER_IP

Ticket: None
Changelog: Title


vpodzime (approved):
It's not really an *arg*, right? The change makes it respect an environment variable. Not a biggie, of course. ACK.

danielskinstad:
You pass it to the entrypoint by sending it as an arg when starting the container, perhaps it can take it from the environment as well? Anyways, I kinda agree so I'll formulate it some other way

PR #2782, opened 2025-08-19
test: update tests for updated abortion handling 

danielskinstad:
Allow specifying the number of reboots expected when aborting a
deployment. A reboot after install has been posted will only reboot once
as it goes straight to rollback reboot.

Don't expect a reboot of the client if the deployment is aborted when
sending "downloading"

Set the polling_frequency in check_expected_statistics to 0 to make sure
we catch the abort steps for `installing`.

Ticket: ME-527

### DIFF
--- a/tests/tests/test_deployment_aborting.py
+++ b/tests/tests/test_deployment_aborting.py
@@ -26,19 +26,16 @@ from .mendertesting import MenderTesting
 
 class BaseTestDeploymentAborting(MenderTesting):
     def abort_deployment(
-        self,
-        container_manager,
-        install_image,
-        abort_step=None,
-        mender_performs_reboot=False,
+        self, container_manager, install_image, abort_step=None, number_of_reboots=0,
     ):
         """
         Trigger a deployment, and cancel it within 15 seconds, make sure no deployment is performed.
 
         Args:
-            mender_performs_reboot: if set to False, a manual reboot is performed and
-                                        checks are performed.
-                                    if set to True, wait until device is rebooted.
+            number_of_reboots: if set to 0, a manual reboot is performed and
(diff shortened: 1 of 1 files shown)


danielskinstad:
@mender-test-bot sync

tests/tests/test_deployment_aborting.py
```diff
@@ -31,6 +31,7 @@ def abort_deployment(
         install_image,
         abort_step=None,
         mender_performs_reboot=False,
+        number_of_reboots=2,
```
lluiscampos, line 34: As discussed face to face, we can unify these two parameters and just interpret `number_of_reboots == 0` as `mender_performs_reboot == False`

tests/tests/test_deployment_aborting.py
```diff
@@ -94,12 +97,21 @@ def test_deployment_abortion_instantly(
     @MenderTesting.fast
     def test_deployment_abortion_downloading(
         self, standard_setup_one_client_bootstrapped, valid_image
+    ):
+        self.abort_deployment(
+            standard_setup_one_client_bootstrapped, valid_image, "downloading",
```
lluiscampos, line 102: Here and above we have an opportunity to make the test more readable by explicitly stating the parameters:
```suggestion
            standard_setup_one_client_bootstrapped, valid_image, "downloading",
            mender_performs_reboot=False
```

(or, if we do the suggestion above, `number_of_reboots=0`)

In general I think in test code the more explicit and less default values the more readable. Don't you agree?

lluiscampos (approved):
Thank you for the face to face clarifications. It looks good!

See two comments for possible improvements 🍿

PR #2787, opened 2025-09-02
QA-1172: Skip HSM providers test on kirkstone based QEMU devices

lluiscampos:
On `kirkstone` we didn't enable the PKCS#11 module for OpenSSL providers API, so this test will fail with:
```
error:16000069:STORE routines::unregistered scheme
```

Add simple logic to skip the test instead.


danielskinstad (approved):
LGTM

PR #2789, opened 2025-09-08
test: package integration as python package

dhaustein:
Ticket: QA-1170
Changelog: Packaged integration test suite as a python package, so it can be imported in other Mender client related tests.


backend-tests/docker/docker-compose.backend-tests.yml
```diff
@@ -10,7 +10,7 @@ services:
             - /var/run/docker.sock:/var/run/docker.sock
         command: ${PYTEST_ARGS}
         environment:
-            PYTHONPATH: "/testutils"
+            PYTHONPATH: "/"
```
lluiscampos, line 13: As far as I know, the whole `backend-tests` can be removed. I will ask to the Server team and follow-up independently (trying to keep the scope creep at bay with your work).

pyproject.toml
```diff
@@ -1,6 +1,53 @@
...
+    "stripe==12.5.0",
+    "aiosmtpd==1.4.6",
+]
+
+[tool.setuptools]
+packages = ["tests", "testutils"]
+
+# include compose files so they can be used downstream
+[tool.setuptools.package-data]
+"*" = ["docker-compose*.yml", "other-components*.yml"]
```
lluiscampos, line 49: `other-components*.yml` should not be used at all in any tests - this is a relic from our release tool that we also plan to deprecate anyway.

pyproject.toml
```diff
@@ -1,6 +1,53 @@
...
 
 [tool.black]
-target-version = ['py38']
+target-version = ['py312']
```
lluiscampos, line 52: Careful with this, our CI uses a [hardcoded version for black](https://github.com/mendersoftware/mendertesting/blob/18e1e8c1e86ebabf63afaf816007dedee831ebbd/.gitlab-ci-check-python3-format.yml#L57).

This is something that I also really want to fix (in CI) but in order to not increase the scope of your work I would keep it in the existing version.

lluiscampos (commented):
This looks promising! I drooped some comments :memo:

PR #2790, opened 2025-09-10
QA-1182: Remove deprecated `backend-tests` test suite

lluiscampos:
These were moved to `mender-server` last year and are no longer part of the "integration" pipeline.


kjaskiewiczz (approved):
thanks for cleaning that up

PR #2881, opened 2026-04-15
fix: poll deviceconnect websocket instead of fixed sleeps

pasinskim:
The docker-client fixture is function-scoped (since 59f3afc2), so test_in_poor_network_environment starts each run with a freshly bootstrapped client and races the initial deviceconnect session handshake. The fixed 30s heal window after the iptables drop has the same problem. Replace both with a bounded retry loop around get_websocket() so the test waits for actual readiness.

Related to QA-1527 and QA-1563.


pasinskim:

merlin-northern (commented):
I am ok with that change, two comments:
* there was a reason I mentioned the HEALTHCHECK on the Mender Docker Client: let's try to consider introducing it with the dbus cli to make the container healthy when the mender-auth responds over dbus (as was the conclusion of the discussion).
* the pipeline failed: https://gitlab.com/Northern.tech/Mender/integration/-/jobs/13944335034 why was it not reported?

tests/tests/test_mender_connect.py
```diff
@@ -212,7 +212,23 @@ def detect_shell_prompt(shell):
...
+            last_err = None
+            while time.monotonic() < deadline:
+                try:
+                    return docker_env.devconnect.get_websocket()
+                except Exception as e:
+                    last_err = e
+                    time.sleep(interval_s)
+            raise AssertionError(
+                f"deviceconnect websocket never became available: {last_err}"
+            )
```
merlin-northern: sweet. usually we used a lib for that. but it is ok.

elkoniu (approved):
LGTM

tests/tests/test_mender_connect.py
```diff
@@ -212,7 +212,23 @@ def detect_shell_prompt(shell):
...
+        def open_websocket_with_retry(timeout_s=180, interval_s=5):
+            # The docker-client fixture is function-scoped, so each test starts
+            # with a freshly-bootstrapped client whose deviceconnect session may
+            # not yet be established. Poll get_websocket() until it succeeds.
+            deadline = time.monotonic() + timeout_s
+            last_err = None
+            while time.monotonic() < deadline:
+                try:
+                    return docker_env.devconnect.get_websocket()
+                except Exception as e:
```
p-targowicz: LGTM but one thing -->  do we want to catch **all** exceptions ? perhaps if there's some specific error idk we want to kill it instantly ? if not that's fine 🥇

p-targowicz (approved):

pasinskim:

merlin-northern (approved):

tests/tests/test_filetransfer.py
```diff
@@ -107,7 +107,11 @@ def set_limits(docker_env, mender_device, limits, auth, devid):
         )
     finally:
         shutil.rmtree(tmpdir)
-    mender_device.run("kill -TERM `pidof %s`" % connect_service_name)
+    mender_device.run(
+        "kill -TERM `pidof %s` 2>/dev/null; "
+        "while pidof %s >/dev/null 2>&1; do sleep 0.1; done"
+        % (connect_service_name, connect_service_name)
+    )
```
merlin-northern: this potentially can hang forever, I would introduce some way to bail out.

pasinskim: @merlin-northern thanks. That makes sense. Fixed!

merlin-northern: oh, and there is one more thing: [touch /stop-respawn](https://github.com/mendersoftware/mender-client-subcomponents/pull/43/changes#diff-6c8a67bbf17b524340b85453f65102eefbfa74a9cdfeed78fb2ca0071fe26842R40), remember?

pasinskim:
@mender-test-bot start integration pipeline but please don't fail it this time!

pasinskim:
@mender-test-bot start integration pipeline but please don't fail it this time!

tests/tests/test_filetransfer.py
pasinskim: Yup. Actually how about 

```
# Kill mender-connect and let the entrypoint's supervise loop restart
 # it automatically with the updated config.
 mender_device.run("kill `pidof %s` 2>/dev/null" % connect_service_name)
```

pasinskim:

pasinskim:

pasinskim:

pasinskim:

pasinskim:

p-targowicz (approved):
lgtm but I see pipelines are failing

pasinskim:

tests/tests/test_mender_connect.py
```diff
@@ -212,7 +212,23 @@ def detect_shell_prompt(shell):
                 "$ ",
             ], "Could not detect shell prompt."
 
-        with docker_env.devconnect.get_websocket() as ws:
+        def open_websocket_with_retry(timeout_s=180, interval_s=5):
```
danielskinstad: A suggestion which applies  here and other places, but we could simplify retries by decorating functions with `redo.retriable` (https://pypi.org/project/redo/)
Example: https://github.com/mendersoftware/integration/blob/2a1d6c521de0aa98720dc85ca88e23627d74916c/testutils/common.py#L561
It retries the function on exceptions and removes the need for stuff like:
```
while not timeout:
    time.sleep(1)
````

File: tests/tests/test_filetransfer.py [hunk 5]
```diff
@@ -465,8 +465,8 @@ def test_upload_limits_preserve_owner_and_group(self, mender_device_setup):
                 "FileTransfer": {
                     "Chroot": "/var/lib/mender/filetransfer",
                     "FollowSymLinks": True,  # in the image /var/lib/mender is a symlink
+                    # PreserveOwner covers both owner and group via os.Chown(uid, gid)
```
danielskinstad, line 470: Unrelated, but just to note: looks like we need to remove this from our docs: https://docs.mender.io/add-ons/mender-connect#limits-configuration

tests/tests/test_mender_connect.py
```diff
@@ -237,8 +237,6 @@ def open_websocket_with_retry(timeout_s=180, interval_s=5):
             detect_shell_prompt(shell)
             is_shell_working(shell)
 
-        docker_env.device.run("apt-get update")
-        docker_env.device.run("apt-get install -y iptables")
```
danielskinstad, line 225: You're restoring this in a separate commit, please clean up

danielskinstad (commented):
I think the commits should be cleaned up a bit. I would remove the formatting commit (the code should be formatted in the commits touching the code) and I would clean up the two commits where you're first adding the iptables stuff and then removing it

pasinskim:

merlin-northern:
> start integration pipeline

@pasinskim do not forget to start a couple of pipelines in the 4.1.x -- thats where the problems are (I think 4.1.x needs some cherry picks).

pasinskim:

p-targowicz:
I think we need to coordinate our merging @pasinskim my PR also touches the same files as yours I don't mind merging yours and then mine what's left untouched but we definitely cant merge both as we have overlapping changes https://github.com/mendersoftware/integration/pull/2900

pasinskim:

pasinskim:
> > start integration pipeline
> 
> @pasinskim do not forget to start a couple of pipelines in the 4.1.x -- thats where the problems are (I think 4.1.x needs some cherry picks).

Passed with `4.1.x` here: https://gitlab.com/Northern.tech/Mender/integration/-/pipelines/2558962191

pasinskim:
> I think we need to coordinate our merging @pasinskim my PR also touches the same files as yours I don't mind merging yours and then mine what's left untouched but we definitely cant merge both as we have overlapping changes #2900

I did run it with `4.1.x`. Now will try again with `main` with the latest changes and we can discuss what to do.

p-targowicz (approved):
lgtm

tests/tests/test_mender_connect.py
```diff
@@ -236,17 +286,21 @@ def detect_shell_prompt(shell):
...
+        # 30 minutes (see connectionmanager/exponentialbackoff.go in
+        # mender-connect); after a 128s outage the backoff timer can leave the
+        # next reconnect attempt minutes away, which is not testable in a CI
+        # window. Restart mender-connect so the fresh process starts at
+        # attempts=0 and reconnects on its first cycle. The entrypoint's
+        # supervise loop respawns it; this mirrors the canonical pattern used
+        # by test_filetransfer.update_limits(). (QA-1527, QA-1591)
+        docker_env.device.run(
+            "kill -TERM `pidof mender-connect` 2>/dev/null || true"
+        )
```
merlin-northern, line 300: I am sorry but this breaks the idea of this test. we absolutely cannot restart mender-connect in here. by construction it has to heal the connection by itself. this is the main purpose of this test, and the main purpose of the Mender Troubleshooting Add-on -- we cannot leave the device without access to the server. and by the way this is what all the fuss here is all about. 4.1.x misses some features that fixed deviceconnect around error handling and I think that the correct path is to try to cherry-pick. could you give me more time to debug it over the weekend until Monday?

merlin-northern (changes requested):
I am sorry, this has lots of good work, but one thing is kind of breaking.

PR #2883, opened 2026-04-18
chore(integration): native docker healthchecks for all services

p-targowicz:
chore: Replaces custom Python container-readiness polling with docker-native HEALTHCHECK

Python cleanup:
- testutils/common.py: removes wait_until_healthy (obsolete HTTP-poll path_map) and its sole helper wait_for_api_gateway.
- docker_compose_manager.py: removes the _wait_for_containers method and its 8 call sites; --wait in _docker_compose_up is authoritative.

Ticket: QA-1581

### DIFF
--- a/testutils/common.py
+++ b/testutils/common.py
@@ -24,7 +24,6 @@ import subprocess
 from contextlib import contextmanager
 from typing import List
 
-import docker
 import redo
 import requests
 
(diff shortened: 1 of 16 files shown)


p-targowicz:

p-targowicz:

p-targowicz:

p-targowicz:

michalkopczan:
How does the internal docker healthcheck work?
In wait_for_containers we're not waiting for the container simply to be up, we wait until our services started and are able to respond to api calls, e.g. `/api/internal/v1/deployments/health` . If we replace it with just the container being up, we can end up in a situation where the service is not running yet (or at all; if it's broken), but try to test anyway - resulting in nonsensical error messages.

p-targowicz:
> How does the internal docker healthcheck work? In wait_for_containers we're not waiting for the container simply to be up, we wait until our services started and are able to respond to api calls, e.g. `/api/internal/v1/deployments/health` . If we replace it with just the container being up, we can end up in a situation where the service is not running yet (or at all; if it's broken), but try to test anyway - resulting in nonsensical error messages.

The new healthchecks are not "just the container being up" — they poll the same /api/internal/v1/<service>/health endpoints that wait_until_healthy used to poll, only now Docker itself runs the poll inside the container instead of our python code hitting it from outside. 

Every service that previously had a mender.healthcheck.path label now has an equivalent healthcheck: block in the compose file - so the logic frankly speaking **did not change** now its just docker's daemon repsponsibility to do do healthchecks rather than python from outside :)

michalkopczan:
@p-targowicz , ah, very nice! thank you

michalkopczan (approved):

p-targowicz:

p-targowicz:
@mender-test-bot  start integration pipeline

p-targowicz:

p-targowicz:

p-targowicz:

p-targowicz:

testutils/common.py
```diff
@@ -679,3 +564,36 @@ def requests_get(url):
     req = requests.get(url, timeout=30)
     req.raise_for_status()
     return req
+
+
+def wait_until_healthy(compose_project: str = "", timeout: int = 60):
```
danielskinstad: If a container has a healthcheck, `docker compose up -d --wait` will wait until the container is running/healthy. Why is this function needed? AFAICT it tries to do the same thing as I would expect docker compose to handle natively with `--wait`.

danielskinstad (commented):
I would squash the two commits, as one seems to undo the other one (at least partially).

testutils/common.py
p-targowicz: good point - I think its leftover from other tries until I got it right, will remove it good catch 💯

p-targowicz:
> I would squash the two commits, as one seems to undo the other one (at least partially).

squashed into one

p-targowicz:

p-targowicz:
@danielskinstad done all latest pipeline share same failures as nightly - its even slightly less

danielskinstad (approved):
Alright, let's give it a go. One thing: the commit linter is failing because you have `Ticket` below `Signed-off-by` -- `Signed-off-by` should be the last line (you can't see the failing job here because the full pipeline overwrite the status)

Approving, but please update the commit

p-targowicz:
@danielskinstad thank you - done

PR #2885, opened 2026-04-23
ci: switch from SSH to HTTPS git checkout

p-targowicz:
chore(ci): switch from SSH to HTTPS git checkout

Ticket: QA-1580

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -175,13 +175,9 @@ renovate:
     - pip3 install pyyaml
     - curl -fsSL https://deno.land/x/install/install.sh | sh
     - curl -sLO https://docs.mender.io/releases/versions.json
-    # Prepare SSH key
-    - eval $(ssh-agent -s)
-    - echo "$SSH_PRIVATE_KEY" | tr -d '\r' | ssh-add - > /dev/null
-    - mkdir -p ~/.ssh
-    - chmod 700 ~/.ssh
-    - ssh-keyscan github.com >> ~/.ssh/known_hosts
-    # Configure git
+    # Configure git to use HTTPS with token instead of SSH
+    - git config --global url."https://mender-test-bot:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/".insteadOf "git@github.com:"
+    - git config --global --add url."https://mender-test-bot:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/".insteadOf "https://github.com/"
     - git config --global user.email "mender@northern.tech"
     - git config --global user.name "Mender Test Bot"
   script:


p-targowicz:

.gitlab-ci.yml
```diff
@@ -175,13 +175,9 @@ renovate:
...
-    # Prepare SSH key
-    - eval $(ssh-agent -s)
-    - echo "$SSH_PRIVATE_KEY" | tr -d '\r' | ssh-add - > /dev/null
-    - mkdir -p ~/.ssh
-    - chmod 700 ~/.ssh
-    - ssh-keyscan github.com >> ~/.ssh/known_hosts
-    # Configure git
+    # Configure git to use HTTPS with token instead of SSH
+    - git config --global url."https://mender-test-bot:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/".insteadOf "git@github.com:"
+    - git config --global url."https://mender-test-bot:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/".insteadOf "https://github.com/"
```
oldgiova: Here probably the second git config entry overwrites the first one: it produces:
```
  [url "https://user:TOKEN@github.com/"]
      insteadOf = https://github.com/

```

```suggestion
    # Configure git to use HTTPS with token instead of SSH
    - git config --global url."https://mender-test-bot:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/".insteadOf "git@github.com:"
    - git config --global --add url."https://mender-test-bot:${GITHUB_BOT_TOKEN_REPO_FULL}@github.com/".insteadOf "https://github.com/"
```

which produces
```
  [url "https://user:TOKEN@github.com/"]
      insteadOf = git@github.com:
      insteadOf = https://github.com/
```
which is probably what you want to achieve

oldgiova (changes requested):
Great stuff!! However, I think the double config creates the wrong result, please check

oldgiova (commented):
Getting closer, thanks! Now could you please squash the two commits in a single one?

Additionally, maybe is it better to use `ci: switch from SSH to HTTPS git checkout` instead of `chore(ci)` as git commit header?

You don't have to run the integration pipeline for non-code changes ;)

.gitlab-ci.yml
p-targowicz: valid point - applied 💯

p-targowicz:
> Getting closer, thanks! Now could you please squash the two commits in a single one?
> 
> Additionally, maybe is it better to use `ci: switch from SSH to HTTPS git checkout` instead of `chore(ci)` as git commit header?
> 
> You don't have to run the integration pipeline for non-code changes ;)

ok changed header - what about commit ?  is it ok ?

oldgiova:
> ok changed header - what about commit ? is it ok ?

I see you squashed it, but I still see `chore(ci)` as header. It's still not a blocker, just a suggestion for improvement, as `chore` is fine for this kind of changes.
For the commit message, it's empty but it's fine: no need to be so verbose with this tiny change.

oldgiova (approved):

p-targowicz:
> > ok changed header - what about commit ? is it ok ?
> 
> I see you squashed it, but I still see `chore(ci)` as header. It's still not a blocker, just a suggestion for improvement, as `chore` is fine for this kind of changes. For the commit message, it's empty but it's fine: no need to be so verbose with this tiny change.

no no I want to make it the good way ! :D indeed i squashed it can change commit too - I can change just commit message w/o squashing to be `ci` only w/o `chore` so no problem 💯

danielskinstad (approved):

PR #2888, opened 2026-04-24
test: skip failing test_filetransfer test

danielskinstad:
I'll create a ticket and replace `QA-XXXX` with it if we want to go this way


danielskinstad:

PR #2928, opened 2026-06-15
test: Poll instead of fixed sleep in test_websocket_reconnect

danielskinstad:
Ported the necessary changes from 539f51d4. The rest of that commit is docker-client specific and does not apply to 3.8.x.

Ticket: QA-1638


p-targowicz (approved):
lgtm

PR #2935, opened 2026-06-22
test: Fetch and build gen_docker-compose for docker-compose tests

vpodzime:
With the changes done in
mendersoftware/mender-container-modules#22 the
`gen_docker-compose` generator script can no longer be fetched over HTTP, it has to be built (and thus the whole repo needs to be cloned).

Ticket: MEN-9442
Changelog: none


danielskinstad (approved):
The pipeline isn't green, but AFAICT the failing test is unrelated

PR #2939, opened 2026-06-26
Reboot crossing watchdog timeout

pasinskim:
Fix the race condition with client reboots inside the integration tests.

### DIFF
--- a/tests/tests/test_fault_tolerance.py
+++ b/tests/tests/test_fault_tolerance.py
@@ -25,6 +25,7 @@ from ..common_setup import (
 )
 from .common_update import common_update_procedure, update_image_failed
 from ..MenderAPI import DeviceAuthV2, Deployments, logger
+from ..MenderAPI.deployments import REBOOT_CROSSING_DEPLOYMENT_TIMEOUT
 from .mendertesting import MenderTesting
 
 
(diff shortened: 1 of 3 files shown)


pasinskim:

danielskinstad (approved):

tests/tests/test_fault_tolerance.py
```diff
@@ -156,8 +157,10 @@ def do_test_update_image_breaks_networking(
             deployment_id, _ = common_update_procedure(
                 broken_network_image, devauth=devauth, deploy=deploy
             )
-            reboot.verify_reboot_performed()  # since the network is broken, two reboots will be performed, and the last one will be detected
-            deploy.check_expected_statistics(deployment_id, "failure", 1)
+            reboot.verify_reboot_performed(max_wait=REBOOT_CROSSING_DEPLOYMENT_TIMEOUT)  # since the network is broken, two reboots will be performed, and the last one will be detected
+            deploy.check_expected_statistics(
+                deployment_id, "failure", 1, max_wait=REBOOT_CROSSING_DEPLOYMENT_TIMEOUT
```
michalkopczan, line 162: won't we as a result wait `2*REBOOT_CROSSING_DEPLOYMENT_TIMEOUT`? Once in verify_reboot_performed and a second time in check_expected_statistics?

I think there's no reason to wait for statistics this long.

tests/tests/test_fault_tolerance.py
```diff
@@ -194,8 +197,10 @@ def do_test_deployed_during_network_outage(
             self.manipulate_network_connectivity(mender_device, True)
 
             logger.info("Network stabilized")
-            reboot.verify_reboot_performed()
-            deploy.check_expected_statistics(deployment_id, "success", 1)
+            reboot.verify_reboot_performed(max_wait=REBOOT_CROSSING_DEPLOYMENT_TIMEOUT)
```
michalkopczan, line 200: I'm not sure I understand the reasoning here. If we wait here for the deployment to finish (successfully), then why would we need to wait for the client watchdog to kill mender-update? If it happens, then there would be no successfully deployment AFAIK.

PR #2943, opened 2026-07-02
test: capture the QEMU client serial console on failure

pasinskim:
The failure teardown only prints device output over SSH, which is empty when the device is unreachable -- which is exactly the "Device never rebooted" failure mode (a hung guest reset). Capture the client's serial console via "docker compose logs" (which works without SSH) into the per-test logs, so the guest's actual reboot behaviour (e.g. "reboot: machine restart" followed by silence) is visible directly in CI artifacts.

Changelog: None


pasinskim:

PR #2950, opened 2026-07-20
test(state-scripts): survive the "Device never rebooted" infra flake

p-targowicz:
test_reboot_recovery is the last reboot-and-verify test without a retry for the "Device never rebooted" QEMU flake: the class-level flaky retry was dropped in 708212cc because a rerun of the slow powerloss tests could blow the 2h job timeout, and with the class-persistent fixtures a rerun would have reused the same dead VM anyway.

This makes the retry affordable and effective:

- fail fast on a dead device: the post-failure log dump and cleanup now use wait=60 instead of burning 2x600s against an unreachable VM
- heal on death: a function-scoped fixture probes the device after each reboot recovery test (~1s when healthy) and only if it is unreachable docker-restarts the client container and re-bootstraps it (bounded waits, swallows its own errors)
- flaky(max_runs=2) on the two test_reboot_recovery methods, so the rerun starts against the healed device

Happy-path cost is ~10s per nightly. A flake now costs ~19 min and a green pipeline instead of ~25 min and a red one.

The tip commit is a TEMPORARY verification patch (marked do-not-merge): it forces the probe to treat every device as dead and fails the first attempt of each class, so this PR's pipeline exercises the heal + rerun path end-to-end. It will be dropped once the pipeline proves the mechanism.

### DIFF
--- a/tests/tests/test_state_scripts.py
+++ b/tests/tests/test_state_scripts.py
@@ -16,6 +16,7 @@ import shutil
 import time
 import os
 import subprocess
+import traceback
 import pathlib
 
 import pytest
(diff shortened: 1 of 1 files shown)


p-targowicz:

p-targowicz:

p-targowicz:

p-targowicz:

p-targowicz:

p-targowicz:

p-targowicz:

michalkopczan (approved):

PR #2951, opened 2026-07-22
fix: Expect a finished deployment after both enough and too few download retries

vpodzime:
Even when the number of download retries is not enough for the
artifact download to succeed, the deployment should be finished
when network is re-established and not left in the `inprogress`
state.

Ticket: MEN-9954
Changelog: None
Signed-off-by: Vratislav Podzimek <vratislav.podzimek+auto-signed@northern.tech>

### DIFF
--- a/tests/tests/test_fault_tolerance.py
+++ b/tests/tests/test_fault_tolerance.py
@@ -388,16 +388,17 @@ class BasicTestFaultTolerance(MenderTesting):
                 mender_device, True, hosts=[blocked_service]
             )
 
+            # In either case the deployment should be marked as finished --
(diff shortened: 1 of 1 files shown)


vpodzime:
> Hello 😺 I created a pipeline for you here: [Pipeline-2699405058](https://gitlab.com/Northern.tech/Mender/integration/-/pipelines/2699405058)

100% :green_heart:  :partying_face:

michalkopczan (approved):

vpodzime:
> ci/gitlab/integ-tests Expected — Waiting for status to be reported

No functional changes since the last green pipeline.

PR #2956, opened 2026-08-06
test: migrate tests to mender-server compose files

rewanrashid-boop:
The backend composition now comes from the mender-server repo, added as a submodule at tests/mender_server. Two overlays in tests/compose/ replace the twelve compose files this repo maintained by hand; 34 files are deleted in total, including cert/, keys/, config/traefik/ and storage-proxy/, all of which mender-server now owns.

Full suite verified locally: 195 collected, 0 failures.

Four things are not obvious from the diff.

Traefik's Docker provider is load-bearing for parallelism. It discovers backends by container label and sees every container on the daemon, not just its own project's. Without the --providers.docker.constraints override in docker-compose.testing.yml, two concurrent namespaces each register the other's services under identical router names and requests round-robin into the wrong environment. Measured, not theorised: every service ended up with two backends.

Compose is invoked with --project-directory pinned to the repo root. Compose otherwise derives it from the first -f file, now under tests/compose/, and silently resolves the root .env and every relative bind mount against the wrong directory. include: paths are resolved the same way, hence written relative to the repo root.

requests_retry() sets a Host header on the session, because Traefik routes on it and tests address the gateway by container IP. The trap: it also retries 5xx, so any test asserting on an error response must pass status_forcelist=[]. This turned test_filetransfer.py's xfail(NotImplementedError) cases into RetryError. Websockets need a different fix again -- the library derives Host from the URL, so passing one via additional_headers emits it twice; a pre-connected socket keeps the URL, and therefore both Host and the TLS SNI name, canonical.

Failover is two compose projects instead of duplicated services. The old file hand-copied every service with a "-2" suffix into one project, which cannot work for the Traefik reason above. Server B is now the same composition in its own project with MENDER_HOSTNAME set, so there is nothing to keep in step with mender-server. B needs its own admin user: useradm's Verify looks up both the user id and the token id in its own database, so a token issued by A can never authorise a request to B, even though the two share a signing key.

Two smaller notes. Mongo runs on tmpfs for speed and per-test isolation; the OS -> Enterprise migration test opts out with persistent_mongo=True because it replaces the backend over a surviving database. XDIST_JOBS_IN_PARALLEL_INTEGRATION drops from 4 to 3, since an environment is now 18 containers (23 for Enterprise) rather than ~14 -- that number wants re-tuning against the real runner.

Known characteristic, not a regression: test_monitor_client.py is the most timing-sensitive file in the suite. Several tests restart services on a QEMU device and poll on short fixed budgets, one as low as 18 seconds. They pass at -n 1 and flake above it.

Ticket: QA-1702


rewanrashid-boop:

rewanrashid-boop:

p-targowicz (approved):
code wise all looking good

rewanrashid-boop:

rewanrashid-boop:
Review waits until package which itself needs: https://github.com/mendersoftware/mender-server/pull/2146

rewanrashid-boop:

rewanrashid-boop:

rewanrashid-boop:

tests/container_manager/__init__.py
```diff
@@ -11,7 +11,10 @@
 #    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
+"""Compose setups specific to this repository.
 
-URL_MGMT = "/api/management/v1/deviceconfig"
-URL_MGMT_DEVICE_CONFIGURATION = "/configurations/device/{id}"
-URL_MGMT_DEVICE_CONFIGURATION_DEPLOY = "/configurations/device/{id}/deploy"
+These used to live under a fork of mender-server's testutils. They stayed behind
```
danielskinstad, line 16: testutils was not a fork of mender-server's. https://github.com/mendersoftware/mender-server/commit/70196ba1 -- rather the other way around: mender-server copied the tests/utilities from this repo when they moved to the monorepo.

.env
```diff
@@ -1,8 +1,14 @@
...
-MENDER_SERVER_ENTERPRISE_REPOSITORY=mender-server-enterprise
-MENDER_SERVER_TAG=main
+# Tag for all mender-server images (OS and Enterprise).
+#
+# Deliberately do NOT set MENDER_IMAGE_REGISTRY or MENDER_IMAGE_REPOSITORY here:
+# mender-server's base and enterprise compose files reuse those same two
+# variables with different defaults (docker.io/mendersoftware vs
+# registry.mender.io/mender-server-enterprise), so setting either one globally
+# breaks the other flavour.
+MENDER_IMAGE_TAG=main
```
danielskinstad, line 8: We have scheduled pipelines running different versions of mender-server: https://gitlab.com/Northern.tech/Mender/integration/-/pipeline_schedules
If we want to move away from `MENDER_SERVER_TAG` because the compose files from mender-server uses `MENDER_IMAGE_TAG`, then we must update the variable in the schedules or find a way to set `MENDER_IMAGE_TAG` based on `MENDER_SERVER_TAG`

tests/run.sh
```diff
@@ -118,6 +121,40 @@ if [[ -z "$BUILDDIR" ]] && [[ -n "$DOWNLOAD_REQUIREMENTS" ]]; then
...
+# the failure surfaced as an opaque exit 127 mid-run.
+export PATH=$PWD/downloaded-tools:$PATH
+
+# Pull the backend images up front. _docker_compose_up passes --pull missing, so
+# a cold cache would otherwise be fetched by whichever test got there first while
+# holding the global compose lock, serialising every other worker behind it. Doing
+# it here also turns a registry auth problem into an immediate, obvious failure.
+if [[ -n "$DOWNLOAD_REQUIREMENTS" ]]; then
+    docker compose \
+        --project-directory .. \
```
danielskinstad, line 136: This doesn't actually work in CI because `compose_dir(dest=os.path.join(THIS_DIR, "mender_server"))` isn't called yet. There are no compose files in tests/mender_server/ yet

danielskinstad (changes requested):
Nice work! I added some comments

.env
rewanrashid-boop: Actually, I was going to remove this but just missed it, it should not be a problem, but I'll see when i update the PR. I remember removing it in other client repo `.env` also.

rewanrashid-boop:

.env
danielskinstad: Without a variable, wow do we select which version of the server to use?

rewanrashid-boop: We don't need the MENDER_IMAGE_TAG=main in the .env because that is already being supplied in the package. About the SERVER_TAG you are right and we can map it. I think we should do it in the run.sh since it is the entry point and then it would work both locally and in CI. wdyt :)

rewanrashid-boop:
@mender-test-bot start client pipeline

.env
danielskinstad: > I think we should do it in the run.sh since it is the entry point and then it would work both locally and in CI. wdyt :)

Sounds reasonable to me :+1:

danielskinstad (approved):

danielskinstad:

PR #2957, opened 2026-08-10
ci: migrate integration fully to renovate

p-targowicz:
Part of QA-1485

Dependabot still owned pip docker and ci here so renovate had those categories disabled behind a TODO. Dropped dependabot and removed the guard

Config now follows renovate.json5.sample from mender-qa. Kept both MENDER_ARTIFACT_VERSION custom managers since it is pinned in .gitlab-ci.yml and .env

The 17 compose files at the root are now covered which dependabot never did - only literal image tags like traefik, the MENDER_SERVER_* ones stay with release tooling

Targeting master and 3.8.x. 3.8.x last moved 2026-06-15 so it is still live, while 3.7.x has been quiet since 2025-03 and 3.6.x since 2024-08 so those are left out. Added the standard rule blocking major bumps on maintenance branches

Dependabot only ever opened pip PRs here - its docker entry pointed at extra/mender-client-docker-addons which contains just a README, and there are no merged dependabot docker PRs in this repo


oldgiova (approved):

p-targowicz:
no need for integration tests its just renovate, merging

PR #2959, opened 2026-08-18
test: retry curl downloads in get_requirements

p-targowicz:
Test jobs die instantly when raw.githubusercontent.com or downloads.mender.io hit a transient error like GitHub's 429 rate limit. Adding curl retries so one flaky request doesn't fail the whole shard.


p-targowicz:

PR #2967, opened 2026-08-28
ci: drop the per repo renovate job

p-targowicz:
Renovate now runs centrally from NorthernTechHQ/renovate-ring, which reads this repo's renovate.json5 straight from GitHub. Nothing about grouping or managers changes.

The job only ever triggered on a pipeline schedule and that schedule is already deleted, so this removes config that can no longer run.


oldgiova (approved):

p-targowicz:
No need for int tests just renovate change merging

PR #2968, opened 2026-08-31
test: verify docker-compose relative bind mounts survive a daemon res…

danielskinstad:
…tart

The docker-compose module used to start compositions from a directory that was renamed on commit, so the containers' bind mounts pointed at a dead path after a reboot (MEN-10102). Deploy a composition with a relative bind mount and restart the docker daemon, which re-executes the mounts from the stored paths -- the same re-bind as a power cycle.

Ticket: MEN-10102
Changelog: none


danielskinstad:

I expect this pipeline to fail due to https://northerntech.atlassian.net/browse/MEN-10102

PR #2970, opened 2026-09-02
test(delta-docker): Add delta-docker-compose integration tests

michalkopczan:
Ticket: MEN-9947


michalkopczan:

michalkopczan:

michalkopczan:
@mender-test-bot start client pipeline

michalkopczan:

michalkopczan:
The failing test: `test_delta_update_pruned_layers`, passes locally. Investigating what can be the difference in the CI.
The fail is because:
` 
           target_size = os.path.getsize(target_artifact)
           assert (
                delta_size < 0.25 * target_size
            ), f"delta={delta_size} full_target={target_size}"
           AssertionError: delta=10627072 full_target=10642944
           assert 10627072 < (0.25 * 10642944)
 `

michalkopczan:

michalkopczan:

michalkopczan:

michalkopczan:
@danielskinstad , @vpodzime , two tests are marked as skipped, and bugtickets for them are created:
* https://northerntech.atlassian.net/browse/MEN-10126
* https://northerntech.atlassian.net/browse/MEN-10122

The tests need to check out the private `mender-delta-container-modules` repo, as they need a generator from there. AFAICT it's OK, but please verify, thank you.

tests/tests/test_delta_docker_compose.py
```diff
@@ -0,0 +1,961 @@
...
+
+Setup notes / assumptions:
+
+  * The *source* full composition is deployed through the open base
+    ``docker-compose`` Update Module (ships in the extended image, same
+    assumption as test_docker_compose.py). The *delta* is deployed through the
+    proprietary ``delta-docker-compose`` module, which is baked into the
+    commercial extended image (mender-client-qemu-extended-commercial) the tests
+    boot.
+  * Because the base and delta modules use *separate* persistent stores
```
vpodzime, line 40: This is no longer the case, see https://github.com/mendersoftware/mender-delta-container-modules/pull/15

tests/tests/test_delta_docker_compose.py
```diff
@@ -0,0 +1,961 @@
...
+@pytest.mark.min_mender_client_version("6.0.0")
+class TestDeltaDockerCompose(MenderTesting):
+    @pytest.mark.skip(
+        reason="MEN-10126: gen_delta-docker-compose's prune_image_layers() relies "
+        "on GNU `find -printf` and `tar --delete`; on the Alpine/musl CI image the "
+        "former is absent (busybox find, nothing pruned) and the latter corrupts "
+        "the image tarball (musl GNU tar), so the generator either emits a "
+        "non-pruned delta or aborts (xargs exit 123). Re-enable once MEN-10126 is "
+        "fixed."
+    )
```
vpodzime, line 447: I don't like this, TBH. We skip a good test of something that is "working fine" because of the selection of a container to run this on in CI. Or am I missing something?

tests/tests/test_delta_docker_compose.py
```diff
@@ -0,0 +1,961 @@
...
+                    n.endswith(".tar") for n in names
+                ), f"delta_with images/ contents: {names}"
+
+    def test_delta_rollback_on_failure(
+        self, delta_device, base_gen_script, delta_gen_script
+    ):
+        """A failing delta (target composition never becomes healthy) rolls back to
+        the previously-committed delta composition.
+
+        A meaningful rollback requires a prior committed delta in the delta store,
```
vpodzime, line 654: This comment is confusing. There is a test for the "non-meaningful" right below.

vpodzime (commented):
Looks good to me otherwise. But it skips tests for the two most important parts.

tests/tests/test_delta_docker_compose.py
```diff
@@ -0,0 +1,961 @@
...
+                image_tars = [
+                    os.path.join(images_dir, f)
+                    for f in os.listdir(images_dir)
+                    if f.endswith(".tar")
+                ]
+                assert image_tars, "no image tarball in delta"
+                assert any(_image_tar_has_empty_layer(t) for t in image_tars)
+
+    @pytest.mark.skip(
+        reason="MEN-10122: the device's xdelta3 is built --without-liblzma and "
```
vpodzime: https://github.com/mendersoftware/mender-delta-container-modules/pull/20 should have made this unnecessary.

michalkopczan: Reenabled the test, thanks

tests/tests/test_delta_docker_compose.py
michalkopczan: Rewritten the comment and changed the test name.

tests/tests/test_delta_docker_compose.py
michalkopczan: Removed this part. Seems that I started working on these tests using an older version of the repo.

The tests do not need any modifications AFAICT. Should delta be able to work with a new/different compose project/service name too? Then we could add a test for this situation.

tests/tests/test_delta_docker_compose.py
michalkopczan: So what you're saying is that because it's a generator and not the device-side tool, we should be OK with using gnu-specific extensions to find/tar?
While we could simply add these gnu version of find, it's not possible for tar AFAICT.

We could change the alpine image to debian/ubuntu but then it would influence all integration tests, not only the delta docker ones.

Alternative approach is to replace the tar --delete with the portable tar -x/tar -c in the generator. What do you think?

michalkopczan:

tests/tests/test_delta_docker_compose.py
vpodzime: > Should delta be able to work with a new/different compose project/service name too? Then we could add a test for this situation.

I believe it should. But at the same time, all of that logic is in the _docker-compose_ Update Module so if we want to test that, we should do it there.

tests/tests/test_delta_docker_compose.py
vpodzime: > We could change the alpine image to debian/ubuntu but then it would influence all integration tests, not only the delta docker ones.

I see. We could split these tests into a new job, though, right?

> Alternative approach is to replace the tar --delete with the portable tar -x/tar -c in the generator. What do you think?

That adds an overhead which is not necessary on most platforms. I don't want to make things worse for all platforms just because one doesn't provide what we need. So it would have to be conditional, I think.

michalkopczan: We could. I updated the description of https://northerntech.atlassian.net/browse/MEN-10126. As this needs creating new jobs in integration I think it goes beyond this task scope - so we leave the test skipped and then we enable it with MEN-10126.

tests/tests/test_delta_docker_compose.py
michalkopczan: Are you sure there's no way for the delta-docker created on top of the logic from docker-compose won't introduce any potentially breaking things? It seems better to test here.

vpodzime: > Are you sure there's no way for the delta-docker created on top of the logic from docker-compose won't introduce any potentially breaking things?

Yes. It only (re)constructs the images, the rest is handled by the same code as _docker-compose UM_ and I don't see a reason for why this should ever change.

michalkopczan: Ok, created https://northerntech.atlassian.net/browse/MEN-10146

vpodzime (approved):
Looks good to me.

